PR #9, opened 2023-05-26
CFP-25694: Move cilium/cilium-cli code into cilium/cilium repository

michi-covalent:
https://github.com/cilium/cilium/issues/25694

### DIFF
--- a/cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
+++ b/cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
@@ -135,6 +135,27 @@ to [cilium/cilium], and run this workflow for pull requests that modify files
 under `cli/` directory. This ensures that any changes under `cli/` directory get
 tested against a released version of Cilium.
 
+## Alternative approach: Extend connectivity test from cilium/cilium repo
+
+Sebastian Wicki suggests another approach to extend `cilium connectivity test`
+command from [cilium/cilium] repo using cilium-cli's [Hook interface]. This
+approach involves two steps:
+
+### Step 1: Implement the Hook interface in cilium/cilium repo.
+
+See [this commit](https://github.com/cilium/cilium/commit/26526a1890de6e05ca58d37b5a2da5822a3f22f0)
+as an example for how to implement the [Hook interface]. This enables you to
+extend `cilium connectivity test` without having to migrate the entire [cilium/cilium-cli]
+code base into [cilium/cilium] repo. Then, build a local version of cilium-cli
+with the additional hook during the CI.
+
+### Step 2: Import the hook back in cilium/cilium-cli repo.
+
+Import the hook from the previous step in [cilium/cilium-cli] repo so that
+additional tests in [cilium/cilium] repo get included in upstream cilium-cli
+releases. See [this commit](https://github.com/cilium/cilium-cli/commit/7b474e61be358fb0b9f2cd6fd075ba843b5d78f5)
+as an example.
+
 [sig-cilium-cli]: https://github.com/orgs/cilium/teams/sig-cilium-cli
 [cilium/cilium]: https://github.com/cilium/cilium
 [cilium/cilium-cli]: https://github.com/cilium/cilium-cli
@@ -143,3 +164,4 @@ tested against a released version of Cilium.
 [backporting process]: https://docs.cilium.io/en/stable/contributing/release/backports/
 [kubernetes/kubernetes staging directory]: https://github.com/kubernetes/kubernetes/tree/master/staging/
 [apimachinery]: https://github.com/kubernetes/apimachinery
+[Hook interface]: https://github.com/cilium/cilium-cli/blob/4a6cb76243704f96dfc02ff312b57e4d0ced0d84/internal/cli/cmd/hooks.go#L13-L17
--- a/cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
+++ b/cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
@@ -10,6 +10,8 @@
 
 **Authors:** Michi Mutsuzaki `<michi@isovalent.com>`
 
+**Status:** Implementable
+
 ## Summary
 
 Move [cilium/cilium-cli] code into [cilium/cilium] repository.
--- a/cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
+++ b/cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
@@ -17,16 +17,33 @@ Move [cilium/cilium-cli] code into [cilium/cilium] repository.
 ## Motivation
 
 The main motivation for moving [cilium/cilium-cli] code into [cilium/cilium]
-repository is to minimize the feature development overhead for:
+repository is to minimize the feature development overhead for adding new tests
+to `cilium connectivity test`.
 
-- adding new tests to `cilium connectivity test`.
-- adding new tasks to `cilium sysdump`.
+Currently, adding a new test to `cilium connectivity test` requires the following
+steps:
+
+- Open a pull request against [cilium/cilium-cli] repo to add a new test, and
+  wait for the CI build to finish.
+- Modify GitHub workflows files to pick up the CI version of `cilium-cli` in
+  your pull request against [cilium/cilium] repo.
+- Once the new tests passes, undo the change in the workflow files.
+- Get these pull requests reveiewed and merged.
+
+Even after all these steps, the new test does not become a part of the regular
+CI until the [sig-cilium-cli] team releases a new version of `cilium-cli` that
+includes the test, and Renovate updates GitHub wokflow files to pick up the new
+`cilium-cli` version. This has been a major source of frustration for some Cilium
+contributors.
 
 ## Goals
 
-* Move [cilium/cilium-cli] code into [cilium/cilium] main branch under `cli/` directory.
+* Move [cilium/cilium-cli] code into [cilium/cilium] main branch under `cli/`
+  directory. This enables Cilium contributors to implement new features with
+  tests in a single pull request against [cilium/cilium] repository.
 * Establish a backporting process for changes inside `cli/` directory.
-* Don't disrupt the cilium-cli development and release processes.
+* Avoid disrupting the cilium-cli development and release processes as much as
+  possible.
 
 ## Non-Goals
 
@@ -41,7 +58,8 @@ This proposal consists of three steps to move [cilium/cilium-cli] code into
 ### 1. Move [cilium/cilium-cli] code to [cilium/cilium] under `cli/` directory
 
 I'm assuming that you want to preserve the Git history of [cilium/cilium-cli]. See
-[michi-covalent/cilium] on how the code migration might look like.
+[michi-covalent/cilium] on how the code migration might look like. Squashing
+these commits into a single commit is another option.
 
 ### 2. Make [cilium/cilium-cli] code a part of `github.com/cilium/cilium` Go module
 
@@ -51,23 +69,19 @@ removing `cli/go.mod`. This allows [cilium/cilium-cli] code to depend on
 
 This is **not** to suggest `cilium-cli` to be tied to a specific `cilium`
 version. The goal of this proposal is to make it easy for contributors to add
-tests against the latest Cilium code. We'll continue to maintain the backward
+tests against the latest Cilium code. Continue to maintain the backward
 compatibility of `cilium-cli` with respect to `cilium` versions.
 
 ### 3. Release `cilium-cli` from [cilium/cilium-cli] repository
 
-We do not need to change the release process / cadence to accomplish the goals
-defined in this CFP. Continue releasing `cilium-cli` from [cilium/cilium-cli] at
-its own cadence using its own versioning that is independent of [cilium/cilium]
-versioning as you have been doing in [cilium/cilium-cli] repository.
+Set up an automated process to periodically sync `cilium-cli` code under `cli/`
+directory back to [cilium/cilium-cli] repo, similar to how Kubernetes syncs some
+modules from [kubernetes/kubernetes staging directory] to module repos
+([apimachinery] for example). Continue releasing `cilium-cli` from
+[cilium/cilium-cli] at its own cadence using its own versioning.
 
-See [michi-covalent/cilium-cli] as an example of how [cilium/cilium-cli] repo
-might look like after the code migration. Basically it contains:
-
-- [`go.mod`](https://github.com/michi-covalent/cilium-cli/blob/865cac4f148ce88cd04d99f8ecfe61a0ae4f645f/go.mod)
-- A simple [`main.go`](https://github.com/michi-covalent/cilium-cli/blob/865cac4f148ce88cd04d99f8ecfe61a0ae4f645f/main.go)
-  that calls [`NewCiliumCommand`](https://github.com/cilium/cilium-cli/blob/44ae1874fae4544c0db34dac89c11e37365b76ef/cli/cmd.go#L27)
-- Release [tags](https://github.com/cilium/cilium-cli/tags) and [artifacts](https://github.com/cilium/cilium-cli/releases)
+See [Tobias' comment](https://github.com/cilium/design-cfps/pull/9#discussion_r1211996796)
+for more details.
 
 ## Impacts / Key Questions
 
@@ -79,26 +93,25 @@ branches since these stable branches use their own copy of `cli/` to run
 `cilium connectivity test`. Apply the same [backporting process] as the rest of
 the Cilium codebase.
 
-### Impact: Additional overhead for cilium-cli development
+### Key Question: How do you run CI for CLI? More specifically, how do you ensure that changes to Cilium CLI remain compatible with older released versions of Cilium?
 
-### Impact: Module name change
+Here is the list of GitHub workflows that currently run for each pull requests in [cilium/cilium-cli]
+using the latest version of Cilium:
 
-This proposal impacts users who depend on [cilium/cilium-cli] as a library.
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/aks-byocni.yaml
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/eks-tunnel.yaml
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/eks.yaml
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/externalworkloads.yaml
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/gke.yaml
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/kind.yaml
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/multicluster.yaml
 
-- The module path needs to be updated from `github.com/cilium/cilium-cli` to
-  `github.com/cilium/cilium/cli`.
-- Instead of depending on a version of `github.com/cilium/cilium-cli` (`v0.14.5`
-  for example), you need to inspect `go.mod` of a specific `github.com/cilium/cilium-cli`
-  version, and transitively find the `github.com/cilium/cilium/cli` version to
-  depend on.
-
-### Key Question: Do we lose any test coverage?
-
-### Key Question: How do we run CI for CLI? More specifically, how do we ensure that changes to Cilium CLI remain compatible with older released versions of Cilium?
-
-### Key Question: Where do we store the release artifacts? Will the cilium/cilium repository release page contain both CNI and CLI releases? Or do we keep cilium-cli for CLI releases?
-
-## Alternatives to Consider
+[cilium/cilium] has a similar coverage using CI images instead of released images,
+so it's probably an overkill to port all of these workflows to [cilium/cilium]. My
+proposal is to move https://github.com/cilium/cilium-cli/blob/main/.github/workflows/kind.yaml
+to [cilium/cilium], and run this workflow for pull requests that modify files
+under `cli/` directory. This ensures that any changes under `cli/` directory get
+tested against a released version of Cilium.
 
 [sig-cilium-cli]: https://github.com/orgs/cilium/teams/sig-cilium-cli
 [cilium/cilium]: https://github.com/cilium/cilium
@@ -106,3 +119,5 @@ This proposal impacts users who depend on [cilium/cilium-cli] as a library.
 [michi-covalent/cilium]: https://github.com/michi-covalent/cilium/pull/169
 [michi-covalent/cilium-cli]: https://github.com/michi-covalent/cilium-cli/tree/cli-test
 [backporting process]: https://docs.cilium.io/en/stable/contributing/release/backports/
+[kubernetes/kubernetes staging directory]: https://github.com/kubernetes/kubernetes/tree/master/staging/
+[apimachinery]: https://github.com/kubernetes/apimachinery
--- a/cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
+++ b/cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
@@ -74,14 +74,18 @@ compatibility of `cilium-cli` with respect to `cilium` versions.
 
 ### 3. Release `cilium-cli` from [cilium/cilium-cli] repository
 
-Set up an automated process to periodically sync `cilium-cli` code under `cli/`
-directory back to [cilium/cilium-cli] repo, similar to how Kubernetes syncs some
-modules from [kubernetes/kubernetes staging directory] to module repos
-([apimachinery] for example). Continue releasing `cilium-cli` from
-[cilium/cilium-cli] at its own cadence using its own versioning.
+We do not need to change the release process / cadence to accomplish the goals
+defined in this CFP. Continue releasing `cilium-cli` from [cilium/cilium-cli] at
+its own cadence using its own versioning that is independent of [cilium/cilium]
+versioning as you have been doing in [cilium/cilium-cli] repository.
 
-See [Tobias' comment](https://github.com/cilium/design-cfps/pull/9#discussion_r1211996796)
-for more details.
+See [michi-covalent/cilium-cli] as an example of how [cilium/cilium-cli] repo
+might look like after the code migration. Basically it contains:
+
+- [`go.mod`](https://github.com/michi-covalent/cilium-cli/blob/865cac4f148ce88cd04d99f8ecfe61a0ae4f645f/go.mod)
+- A simple [`main.go`](https://github.com/michi-covalent/cilium-cli/blob/865cac4f148ce88cd04d99f8ecfe61a0ae4f645f/main.go)
+  that calls [`NewCiliumCommand`](https://github.com/cilium/cilium-cli/blob/44ae1874fae4544c0db34dac89c11e37365b76ef/cli/cmd.go#L27)
+- Release [tags](https://github.com/cilium/cilium-cli/tags) and [artifacts](https://github.com/cilium/cilium-cli/releases)
 
 ## Impacts / Key Questions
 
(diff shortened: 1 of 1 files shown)


cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
```diff
@@ -0,0 +1,70 @@
...
+### Releasing cilium-cli
+
+We do not need to change the release process / cadence to accomplish the goals
+defined in this CFP. Continue releasing `cilium-cli` from the main branch at its
+own cadence using its own versioning that is independent of [cilium/cilium]
+versioning as we have been doing in [cilium/cilium-cli] repository.
+
+We need to come up with a git tag convention for cilium-cli releases to distinguish
+tags for Cilium releases from tags for cilium-cli releases. My proposal to use
+`cilium-cli/vX.Y.Z` for cilium-cli release tags.
```
gandro, line 49: Whatever scheme we come up with, this needs to be compatible with Go modules, otherwise it will break users who pull in cilium/cilium or cilium/cilium-cli as a go module.

If we want to merge merge cilium-cli into the `github.com/cilium/cilium` Go module, then pulling in cilium-cli as a dependency will be tricky, because then the version of the Go module would inherently be tied to Cilium (the CNI) releases, because Go mod requires semantic versioning, and `cli-v0.13.1` is not a valid semantic version.

If, on the other hand, we want to keep cilium-cli as it's own Go module, and it's import path would be e.g. `github.com/cilium/cilium/cli` then we'll need something like `cli/vX.Y.Z` according to https://go.dev/blog/publishing-go-modules - but that means we'd still a dependency from the CLI module to the CNI module. That might require us to use Go workspaces.

I think this point (separate Go module vs. shared Go module) should be a key question that needs to be answered as part of the CFP.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
```diff
@@ -0,0 +1,70 @@
...
+branches since these stable branches use their own copy of `cilium-cli/` to run
+`cilium connectivity test`. Apply the same [backporting process] as the rest of
+the Cilium codebase.
+
+## Impacts / Key Questions
+
+### Impact: Additional overhead for cilium-cli development
+
+### Key Question: Do we lose cilium-cli git history when we move cilium-cli code into cilium/cilium repository?
+
```
gandro, line 64: Missing key question: How do we run CI for CLI? More specifically, how do we ensure that changes to Cilium CLI remain compatible with older released versions of Cilium?

At the moment, `cilium/cilium` CI only tests against `main`, we would have to have a CI workflow that tests the Cilium CLI main at least against the last Cilium CNI stable release, otherwise we'd loose CI coverage during the merge.

gandro, line 64: Missing key question: Where do we store the release artifacts? Will the `cilium/cilium` repository release page contain both CNI and CLI releases? Or do we keep `cilium-cli` for CLI releases?

gandro (commented):
I know this is still in draft, but I've left a bit of early feedback

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
michi-covalent: ... i have another radical idea. let me try and see if it works 🧠

tklauser: Another option might be to use an approach similar to what upstream k8s is doing for some of its modules, e.g. `k8s.io/apimachinery` where code changes are made in the main k8s repo (i.e. https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/apimachinery in case of `k8s.io/apimachinery`) and these are then synced to their own repo (i.e. https://github.com/kubernetes/apimachinery in case of `k8s.io/apimachinery`).

I think using a similar approach for cilium-cli would allow to use the in-tree version for cilium/cilium CI while still allowing independent releases from cilium/cilium-cli without having to keep e.g. multiple `go.mod` files in cilium/cilium or using Go workspaces.

michi-covalent: yeah i think this is the cleanest way to avoid impacting users who import `github.com/cilium/cilium-cli` as a module.

i was thinking about a similar, but much lazier approach:

- https://github.com/cilium/design-cfps/blob/bf48480707bb3b31ed182b21754b1d50748cbc35/cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md?plain=1#L57-L70
- https://github.com/cilium/design-cfps/blob/bf48480707bb3b31ed182b21754b1d50748cbc35/cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md?plain=1#L84-L93

i'm fine with either of these approaches. i'll document both 📝

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
michi-covalent: added an answer 👍

michi-covalent: answered in the proposal 👍 but yeah the proposal is to keep `cilium-cli` repo to host releases.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
michi-covalent: ... on the second thought, i like your approach better. let me just document your approach.

michi-covalent: ... i'm having third thought 😹 if we take this approach, where does `go.mod` file for `cilium-cli` come from? i see that these staging directories contain `go.mod` files with some `replace` directives. for example: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/go.mod

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
```diff
@@ -0,0 +1,123 @@
...
+**Authors:** Michi Mutsuzaki `<michi@isovalent.com>`
+
+## Summary
+
+Move [cilium/cilium-cli] code into [cilium/cilium] repository.
+
+## Motivation
+
+The main motivation for moving [cilium/cilium-cli] code into [cilium/cilium]
+repository is to minimize the feature development overhead for adding new tests
```
bimmlerd, line 20: I read this as trying to improve development velocity. Won't there be the unintended side-effect of worsening the velocity of cilium-cli? If we require full cilium CI runs for changes that touch just cilium-cli, I think that'd be a source of frustration again.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
```diff
@@ -0,0 +1,123 @@
...
+
+Currently, adding a new test to `cilium connectivity test` requires the following
+steps:
+
+- Open a pull request against [cilium/cilium-cli] repo to add a new test, and
+  wait for the CI build to finish.
+- Modify GitHub workflows files to pick up the CI version of `cilium-cli` in
+  your pull request against [cilium/cilium] repo.
+- Once the new tests passes, undo the change in the workflow files.
+- Get these pull requests reveiewed and merged.
```
bimmlerd, line 31:
```suggestion
- Get these pull requests reviewed and merged.
```

bimmlerd (commented):
It seems to me that the main disconnect is that we want both
1. codevelopment of feature (in cilium/cilium) and test (currently in cilium-cli),
2. but also the benefits of isolating cilium-cli from the cilium codebase (for release cadence, dev velocity, ensuring compatability, ...)

Or put differently, is cilium-cli a tool primarily for users, or primarily for CI?

Should we be instead working towards a future where we can develop tests in cilium/cilium, and in CI the CLI picks them up and runs them against the PR version?

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
michi-covalent: a perfect timing, i was just adding a point in the impact section.

i think this will be the main practical drawback with this CFP. this is a major slowdown in terms of cilium-cli development speed unless we are super careful about defining required tests for various files. i don't really have any good answer for it though, other than to hand wave and say that this will motivate more people to make the cilium CI run faster.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
michi-covalent: > not keeping a separate go.mod for the in-tree cilium-cli at all and just use the main go.mod. 

yeah i don't think keeping cilium-cli as a separate module is worth all the effort. 

however, with this approach, you also need to modify all the import paths in .go files back from `github.com/cilium/cilium/cli` to `github.com/cilium/cilium-cli`.

... i'm starting to like my original proposal again 😹

michi-covalent:
> but also the benefits of isolating cilium-cli from the cilium codebase (for release cadence, dev velocity, ensuring compatability, ...)

i'm not too worried about release cadence and ensuring compatibility. i believe this CFP addresses these points reasonably well.

dev velocity is a real concern. however, once we finish migrating cilium-cli to helm mode, most of the cilium-cli development might become:

- adding tests
- adding sysdump tasks
- occasionally adding new sub-commands / flags to accommodate new features

for these things, it might actually be faster to develop everything in cilium/cilium 🤔 

merging two repos will most likely reduce the overall number of dependency update pull requests too.

i guess what i'm saying is: dev velocity is a bit difficult to quantify 💭 

> Should we be instead working towards a future where we can develop tests in cilium/cilium, and in CI the CLI picks them up and runs them against the PR version?

potentially, i need a bit more detail on how this might look like. happy to chat about it.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
```diff
@@ -0,0 +1,145 @@
...
+
+Even after all these steps, the new test does not become a part of the regular
+CI until the [sig-cilium-cli] team releases a new version of `cilium-cli` that
+includes the test, and Renovate updates GitHub wokflow files to pick up the new
+`cilium-cli` version. This has been a major source of frustration for some Cilium
+contributors.
```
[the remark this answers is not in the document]
joestringer: My recollection of the origins of Cilium CLI in general was that it's intended to focus on the user experience in order to simplify installing, observing, maintaining live Cilium clusters. Tools like `cilium connectivity test` were conceived as a sanity check for user environments so we can quickly see whether a prod environment is healthy or not.

Over time, given that this framework provided mechanisms to develop new tests without the baggage of ginkgo and jenkins, I think it became an attractive place for Cilium developers to expand testing. Now we're at several minutes to assess the cluster sanity, and it's doing a lot of advanced feature testing. We're talking about reconfiguring the target environment to perform some of this testing, and starting to hide such "extra test" options behind special flags to avoid impacting prod environments.

I have a bit of concern that over time we're losing sight of the original UX-focused goals of Cilium-CLI and turning it into the Cilium developer CI tool instead. On the way on that path, we'll simplify some Cilium developer workflows (ideal) but also potentially raise the barrier to entry for cilium-cli contribution (not ideal).

Should the tests for advanced Cilium functionality exist in cilium/cilium or cilium/cilium-cli? Well, probably cilium/cilium. Does that mean that all of cilium-cli should then go into cilium/cilium? I'm not so sure. I'd hope that most testing can be done in unit or module / cell level tests directly in the tree, rather than being e2e. Even if we do need e2e tests for some of the new functionality, is the user-facing tool the right place to implement a lot of that advanced testing?

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
```diff
@@ -0,0 +1,145 @@
...
+
+### 3. Release `cilium-cli` from [cilium/cilium-cli] repository
+
+We do not need to change the release process / cadence to accomplish the goals
+defined in this CFP. Continue releasing `cilium-cli` from [cilium/cilium-cli] at
+its own cadence using its own versioning that is independent of [cilium/cilium]
+versioning as you have been doing in [cilium/cilium-cli] repository.
+
+See [michi-covalent/cilium-cli] as an example of how [cilium/cilium-cli] repo
+might look like after the code migration. Basically it contains:
```
joestringer, line 83: Another option could just be a git mirror for the cli/ directory, something like how https://github.com/libbpf/libbpf mirrors in from the Linux kernel tree.

michi-covalent: yeah @tklauser also suggested something similar.

it's a bit more complicated for Go modules because we need to modify all the import paths from `github.com/cilium/cilium/cli` back to `github.com/cilium/cilium-cli`.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
```diff
@@ -0,0 +1,145 @@
...
+includes the test, and Renovate updates GitHub wokflow files to pick up the new
+`cilium-cli` version. This has been a major source of frustration for some Cilium
+contributors.
+
+## Goals
+
+* Move [cilium/cilium-cli] code into [cilium/cilium] main branch under `cli/`
+  directory. This enables Cilium contributors to implement new features with
+  tests in a single pull request against [cilium/cilium] repository.
+* Establish a backporting process for changes inside `cli/` directory.
```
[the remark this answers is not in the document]
michi-covalent: there are more details in "Impact: Backporting to stable branches" section. let me know if it makes sense 👀

joestringer:
One more thought, I like how sysdump isn't too heavily tied to individual cilium versions, because cilium-cli delegates some of the details to Cilium. In the back of my mind, I'm wondering if the solution to this isn't to more tightly couple cilium & cilium-cli, but rather the inverse - further decouple the details so that cilium-cli needs fewer cilium-specific changes over time.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
```diff
@@ -0,0 +1,145 @@
...
+Here is the list of GitHub workflows that currently run for each pull requests in [cilium/cilium-cli]
+using the latest version of Cilium:
+
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/aks-byocni.yaml
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/eks-tunnel.yaml
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/eks.yaml
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/externalworkloads.yaml
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/gke.yaml
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/kind.yaml
+- https://github.com/cilium/cilium-cli/blob/main/.github/workflows/multicluster.yaml
```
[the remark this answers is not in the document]
michi-covalent: these tests build cilium-cli from the tree.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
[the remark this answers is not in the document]
joestringer: Ah, I missed the `since` in the middle there. I think the _goal_ in that case is "Continue to test older Cilium versions against new versions of the CLI". The extra backporting overhead is a logical consequence of the proposal.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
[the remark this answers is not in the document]
michi-covalent: > Could we model the cilium/cilium CI like this instead, and autopublish CLI images for every PR or every main commit... then provide a way for cilium/cilium PRs to point to those CLI dev images?

we already do that, sort of. you can specify a commit SHA of cilium-cli here: https://github.com/cilium/cilium/blob/6d4b2f7092d634f67c6192c85143051c49846245/.github/workflows/conformance-aks.yaml#L74. you just need to edit these workflow files. i guess some people find this step more annoying than others.

>  I'd hope that most testing can be done in unit or module / cell level tests directly in the cilium/cilium tree, rather than being e2e. Even if we do need e2e tests for some of the new functionality, is the user-facing tool the right place to implement a lot of that advanced testing?

i think this is the key question. `cilium connectivity test` unintentionally became "the" place to add e2e tests. maybe that's where things started going wrong 🔥 cc @brb.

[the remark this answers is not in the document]
michi-covalent: @gandro had another idea: use cilium-cli's `Hook` interface https://github.com/cilium/cilium-cli/blob/main/internal/cli/cmd/hooks.go. this might make more sense if we want to keep cilium-cli relatively lean, and at the same time let cilium developers use the same `cilium connectivity test` framework to add advanced e2e tests.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
brb: I think today there is a hidden cost of cilium-cli development. CLI changes gets eventually tested by cilium/cilium. Discovering breakages in changes after merging leads leads to higher costs.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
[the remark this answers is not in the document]
brb: > then provide a way for cilium/cilium PRs to point to those CLI dev images?

I added this - https://github.com/cilium/cilium/pull/25228. However, bunch of workflows `# pull_request: {}` need to be uncommented though to make changes effective.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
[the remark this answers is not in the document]
brb: > then provide a way for cilium/cilium PRs to point to those CLI dev images?

I added this - https://github.com/cilium/cilium/pull/25228. However, bunch of workflows `# pull_request: {}` need to be uncommented though to make changes effective.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
brb: Sure. We run CLI's CI with only a few configurations compared to cilium/cilium's CI (by "CI" I mean GHA workflows), which has quite a few. For example, after releasing the https://github.com/cilium/cilium-cli/releases/tag/v0.14.4 release, some workflows got broken, which required releasing v0.14.5 with https://github.com/cilium/cilium-cli/pull/1660 regardless that we ran some of the CLI CI builds on cilium/cilium (manually editing workflows).

I think @jibi can tell more war stories from his recent effort to add the Egress GW tests.

jibi: > I think @jibi can tell more war stories from his recent effort to add the Egress GW tests.

what happened in my case is that I added some more tests to the connectivity suite, everything was green in cilium-cli CI but then the moment we switched to the new version in cilium/cilium (we had to bump from v0.13.x to v0.14.y) the e2e workflow broke.

Worth noting it was because of unrelated/previous changes, so we might end up again in a situation where we get changes merged in the CLI and then it's up to the person bumping the version used in e2e to figure out what caused a breakage rather than the original author of the CLI changes

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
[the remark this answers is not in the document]
jibi: > @gandro had another idea: use cilium-cli's `Hook` interface

I like also this idea :bulb: keep the connectivity test in cilium-cli relatively contained and have:
* a cilium-cli in OSS cilium that extends cilium/cilium-cli, like we are currently doing with CEE, with the full connectivity test suite
* a cilium-cli in CEE cilium that extends the OSS one with the enterprise specific testing bits

in this way each repo would own its tests (and CEE can still run the full OSS suite).

One problem I see with that approach is that it still doesn't solve the current cilium-cli CI issue, as it's possible to miss breakages until the CLI changes are tested in the cilium/cilium workflows.

We could go either with the renovate approach, which while not optimal, would still improve the current state of things, as at least once a week cilium/cilium would be synced with the most recent CLI revision, or we could add an additional workflow in the CLI that pulls latest cilium/cilium main, bumps the CLI dependency to the CLI PR revision and run the full e2e suite (even though that doesn't work with CEE I guess)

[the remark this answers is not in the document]
brb: > at least once a week cilium/cilium would be synced with the most recent CLI revision

Unfortunately, everything which gets tested after a merge usually involves a few folks spending their cycles instead of a single PR owner.

michi-covalent:
thank you all for your feedback 🙏. let me try to summarize all the discussions here.

at a very high level, there are 2 options that emerged from this CFP:
- `option A`: merge `cilium-cli` code into `cilium/cilium` repository. this is what i originally proposed.
- `option B`: use cilium-cli's [`Hook`](https://github.com/cilium/cilium-cli/blob/main/internal/cli/cmd/hooks.go) interface to extend `cilium connectivity test` without merging `cilium-cli` code into `cilium/cilium` repo.

based on the discussion in [this thread](https://github.com/cilium/design-cfps/pull/9#discussion_r1218627471), looks like people are ok with `option B`. since we have already discussed `option A` extensively, i'll focus on `option B` in this comment. 

### pros

- no need to touch `cilium/cilium-cli` repo at all.

### cons / open questions

- [from jibi](https://github.com/cilium/design-cfps/pull/9#discussion_r1222704140): One problem I see with that approach is that it still doesn't solve the current cilium-cli CI issue, as it's possible to miss breakages until the CLI changes are tested in the cilium/cilium workflows.

  maybe we could add one workflow in `cilium/cilium-cli` repo that runs against the latest cilium from main branch 👀 then, we just bump `cilium/cilium-cli` version in `cilium/cilium` repo whenever there is a new release using renovate.

- [from martynas](https://github.com/cilium/design-cfps/pull/9#discussion_r1219921661): However, I find it useful that users have means to do an extensive testing in their clusters.

  maybe there is a way to import these tests in `cilium/cilium` repo from `cilium/cilium-cli` repo 🤔 

- (from michi) if we go with `option B`, we need general guidelines to clarify when a test should be added directly to `cilium/cilium-cli` repo, vs when it should go under `cilium/cilium` repo.

i'll write up a proposal for `option B` in a separate file 📝

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
```diff
@@ -0,0 +1,167 @@
...
+
+[cilium/cilium] has a similar coverage using CI images instead of released images,
+so it's probably an overkill to port all of these workflows to [cilium/cilium]. My
+proposal is to move https://github.com/cilium/cilium-cli/blob/main/.github/workflows/kind.yaml
+to [cilium/cilium], and run this workflow for pull requests that modify files
+under `cli/` directory. This ensures that any changes under `cli/` directory get
+tested against a released version of Cilium.
+
+## Alternative approach: Extend connectivity test from cilium/cilium repo
```
michi-covalent, line 138: looking at https://github.com/cilium/cilium-cli/pull/1683 as an example of extending `cilium connectivity test`, i'm not sure how well this approach works in practice. maybe the hook interface needs to be extended to support stuff like:

- add new flags to `cilium connectivity test` command
- extend the setup logic

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
michi-covalent: > If we require full cilium CI runs for changes that touch just cilium-cli

i think this is much less of a concern now that jenkins is gone. the full CI run takes ~45 minutes maybe, which is comparable to how long cilium-cli CI takes 👀

michi-covalent:
ok i've been sleeping on this for a while now, talking to various people here and there. let me summarize what i'm currently thinking, and what next steps we should take to achieve the original goal of minimizing the feature development overhead for adding new integration tests for cilium features.

## tldr

pursue option B: use cilium-cli's [Hook](https://github.com/cilium/cilium-cli/blob/main/internal/cli/cmd/hooks.go) interface to add tests.

## what's wrong with merging cilium-cli repo to cilium repo?

nothing is wrong per se, and merging cilium-cli repo is certainly one way to make it easier to use [the connectivity package](https://github.com/cilium/cilium-cli/tree/main/connectivity). however now i see flaws in my original reasoning for merging cilium-cli repo to cilium repo. my original reasoning went like this:

1. i need a way to add integration tests in cilium/cilium.
2. i like using cilium-cli's [connectivity package](https://github.com/cilium/cilium-cli/tree/main/connectivity) for writing tests.
3. however i don't like the connectivity package being in a different repository because i need to open a separate pull request if i need to enhance the connectivity package to accommodate what i need to accomplish.
4. therefore let's merge cilium-cli repo into cilium repo.
5. face any unpredictable consequences in terms of disrupting cilium-cli's development and release processes.
6. figure out if we need to maintain cilium-cli's go module name to avoid breaking projects that import cilium-cli.

the main flaw of this reasoning is in 3 where it jumps from "i need to improve the package i'm using for testing" to "let's merge the entire repo to avoid opening separate pull requests for it". if it were any other testing library (like let's say testify), i'd be much more hesitant to copy it over into cilium repo just to avoid having to open separate pull requests.

i acknowledge that the connectivity package is still in a very early stage of development in terms of enabling outside projects to use it as a library. therefore early adopters often find that they need to improve the connectivity package. @jibi has been trailblazing and making many contributions in this area, including:

- https://github.com/cilium/cilium-cli/pull/2050 
- https://github.com/cilium/cilium-cli/pull/1986
- https://github.com/cilium/cilium-cli/pull/1978
- https://github.com/cilium/cilium-cli/pull/1977

my naively optimistic assertion is that this situation will improve over time as the connectivity package matures, and we can get to the point where most people can simply use cilium-cli's [Hook](https://github.com/cilium/cilium-cli/blob/main/internal/cli/cmd/hooks.go) interface to add new tests without having to polish off rough edges in the connectivity package.

if this assertion is wrong, pain points for using the connectivity package should be easily identifiable. and again i'm optimistically confident in cilium community's ability to come up with pragmatic solutions to address these pain points. 

## next steps

- add a test in cilium/cilium using cilium-cli's [Hook](https://github.com/cilium/cilium-cli/blob/main/internal/cli/cmd/hooks.go) interface to see how it might look like. i'll try to find some to do this myself, but ping me if anybody is interested in experimenting with this.

michi-covalent:
sorry for the delay here are proof of concept draft PRs:

- https://github.com/cilium/cilium/pull/31077 moves `connectivity` package from cilium-cli to cilium.
- https://github.com/cilium/cilium-cli/pull/2355 imports back the `connectivity` package from cilium.

we might decide to move more packages later on but i think this is a good first cut. @brb @tklauser please take a look and let me know if the overall approach looks ok 🚀🙏

brb:
@michi-covalent Thanks! Could you TL;DR a new workflow when adding new feature / tests to cilium/cilium? In particular, how they will be run by CLI in the existing GH workflows?

michi-covalent:
yeah so it uses cilium-cli github action https://github.com/cilium/cilium-cli/blob/main/action.yaml to build the Cilium CLI from `pkg/cilium-cli` directory with this main.go: https://github.com/cilium/cilium/blob/5a838dbb895ef3288c9a14f9e997b44bab874105/pkg/cilium-cli/main.go so you can add you tests in `pkg/cilium-cli/connectivity/` and they'll magically get picked up in your PR 🪄

xmulligan:
@michi-covalent is this still up to date with the current ideas?

michi-covalent:
yes coming soon i promise

xmulligan:
Ok, thanks. Was actually just checking if you think it is ready to merge or if anything else needs to be added or updated

joestringer:
One thing I like from the proposed process in https://github.com/cilium/design-cfps/pull/37 ([README preview](https://github.com/katiestruthers/design-cfps/blob/6b8b8310b3ce3e9e4365302a843ecf560f4cb5b9/README.md#status)) is this idea that we can merge a CFP once the ideas are agreed in principle, with the status being "implementable". If the CFP is up to date and the relevant stakeholders approve, we are good to do that even if we are still going through the practical ramifications of how we implement this.

Then we can come back later to update the CFP as "experimental" / "stable" if something changes significantly in the design. Those changes would also be PRs and we can always debate the details at that point too. Merging an "implementable" design does not mean we are hard committed to the specific approach, things are always still flexible.

xmulligan:
I've tried to outline a short status framework [here](https://github.com/cilium/design-cfps/pull/42). Practically speaking, if this CFP has the approval of one committer, then it could be merged as implementable.

cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md
```diff
@@ -0,0 +1,167 @@
...
+
+**Code Name: Mega Ketchup**
+
+**SIG: [sig-cilium-cli]**
+
+**Begin Design Discussion:** 2023-05-25
+
+**Cilium Release:** 1.15
+
+**Authors:** Michi Mutsuzaki `<michi@isovalent.com>`
```
xmulligan, line 11:
```suggestion
**Authors:** Michi Mutsuzaki `<michi@isovalent.com>`

**Status:** Implementable
```

xmulligan: See https://github.com/cilium/design-cfps#status for what this means

michi-covalent:
@brb it would mean a lot to me if you approve this pull request 🥦 https://github.com/cilium/cilium/pull/34178 just got merged 🚀🙏

brb (approved):
:tada:

PR #12, opened 2023-10-16
cfp: Added CFP for blocking traffic to certain endpoints with default allow

tamilmani1989:
Added CFP for blocking traffic to certain endpoints with default allow. 

Github issue: https://github.com/cilium/cilium/issues/26135

### DIFF
--- a/cilium/CFP-26135-default-egress-allow.md
+++ b/cilium/CFP-26135-default-egress-allow.md
@@ -1,0 +1,225 @@
+# CFP-26135: Block Endpoints with default egress allow
+
+**SIG: SIG-Policy**
+
+**Begin Design Discussion:** 2023-06-12
+
+**Cilium Release:** 1.15
+
+**Authors:** Tamilmani Manoharan <tamanoha@microsoft.com>
+
+## Summary
+
+The current implementation of Cilium lacks a mechanism to restrict egress traffic to specific endpoints without blocking all traffic. When the egress policy is enabled for an endpoint, it effectively goes into a default deny state for egress traffic. This proposal talks about options for adding default egress allow while blocking egress to specific endpoints.
+
+
+## Motivation
+
+In cloud provider environments, it is often necessary to enforce firewall policies to restrict access to certain infrastructure components. Historically, before the emergence of eBPF, firewall rules were enforced using IPTables. Cilium primarily utilizes eBPF for policy enforcement. In eBPF host routing mode, Cilium bypasses IPtable rules and rules can be enforced only via eBPF. Also in certain scenarios (eg: L7), cilium prepends some IPTable rules by default which conflicts with cloud provider specific rules. Disabling the "prepend-iptables-chains" option is also not a viable solution as disabling prepend means Cilium can no longer guarantee that its policies are enforced.
+
+This proposal aims to bring the management of cloud provider specific policies within Cilium. The primary goal of this proposal is to integrate the cloud provider-specific policies into the Cilium framework. By doing so, we aim to address potential conflicts and other concerns associated with managing these policies across multiple components.
+
+## Goals
+
+- Block egress traffic to certain endpoints without blocking all
+- Behavior of existing network policy or cilium network policy should not be affected
+
+## Non Goals
+ - Ingress Restriction without blocking all
+ - Applying this policy for specific endpoints
+ - Complete Network Policy support with all features
+
+## Proposal
+
+### Overview
+Allow cloud providers to block egress traffic to certain endpoints without blocking all. This proposal considers to solve this via bpf policy rules as it will work for both legacy host routing and ebpf host routing modes.
+
+### Option 1 (CommandLine Flag)
+
+Add a new command line flag, --egress-deny-tuples=[ip:port:protocol] as stringSlice. Cloud providers would customize this cilium config to specify apriori that traffic from Cilium-managed Pods towards specific address+port+protocol combinations should be blocked. This flag would effectively act independently of all of the NP, CNP, CCNP resources and enforce a deny for such tuples regardless of the user-defined policy. The deny tuples from cilium config are sanitized and converted to cilium egress deny policy rules. Then these rules should get added to policy map of all endpoints. 
+
+#### Adding new field in ruleMetadata
+Based on config received via command line, Cilium Agent would convert it to EgressDeny rule and does policyAdd. PolicyAdd will add the rules to repository and triggers policyreaction event to trigger identity allocation for labels specified in the policy. Cilium by default goes to deny mode if any policy added. To prevent this, this rule can be created with this flag `DefaultAllow` enabled. If a rule has this flag enabled, then cilium will never enforce default deny for the rest of traffic unless if there any egress k8s network policy or cilium network policy rule is enforced. By default, this flag is disabled and it would not regress existing network policy scenarios. 
+
+`api.Rule` can extend `ruleMetadata` in `api:rule` structure to add a new field named `DefaultAllow`. This field is internal and not exposed to user. When evaluating matching policy rules for an endpoint, based on this flag cilium can differentiate this deny rule from other network policy rules. If all matching rules contains `DefaultAllow` to true then default allow will be enforced. If any one of matching rule contains `DefaultAllow` to false, then default deny will be enforced.
+
+`getMatchingRules` function inaddition to ingressMatch and egressMatch, it would also return `adminEgressMatch`.  adminEgressMatch is enabled only if there `DefaultAllow` is set for the rule and there is no other egressMatch without defaultAllow set. 
+
+`Resource` field in `policy.AddOptions` used to indicate the source of policy rule. A new entry of `ResourceKind` named `ResourceKindDaemonConfig` can be added to indicate the source of this rule is via cilium config.
+
+#### Examples
+##### Example 1
+
+**Cloud Provider deny tuples**  
+-–egress-deny-tuples=[“169.254.128.5;80;;]
+
+**Network Policy/Cilium Network Policy**  
+None
+
+**Action**  
+Egress - Default Allow for all. Block access to 169.254.128.5:80 for all  
+Ingress - Default Allow for all
+
+##### Example 2
+
+**Cloud Provider deny tuples**  
+-–egress-deny-tuples=[“169.254.128.5;80;;]
+
+**Network Policy/Cilium Network Policy**  
+```
+apiVersion: "cilium.io/v2"
+kind: CiliumNetworkPolicy
+metadata:
+  name: "l3-egress"
+spec:
+  endpointSelector:
+    matchLabels:
+      role: backend
+  egress:
+  - toEndpoints:
+    - matchLabels:
+        role: frontend
+```
+**Action**  
+Egress - Allow for all except for pods with label “role: backend”. For those pods, egress allowed only to pods with “role: frontend”. Block access to 169.254.128.5:80 for all  
+Ingress - Default Allow for all
+
+### Alternate Options
+#### Option 2 (CCNP CRD with new field)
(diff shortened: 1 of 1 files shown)


cilium/CFP-26135-default-egress-allow.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Non Goals
+ - Ingress Restriction without blocking all
+ - Applying this policy for specific endpoints
+ - Complete Network Policy support with all features
+
+## Proposal
+
+### Overview
+The proposal will talk about preferred option and also about other options that aims to solve the problem by allowing cloud provider to specify the intent of blocking egress to certain endpoints without blocking all. These deny rules can be plumbed either via iptables or bpf policy rules. IPtables rules would not work with ebpf host routing mode. This proposal considers to solve this via bpf policy rules as it will work for both legacy host routing and ebpf host routing modes. It will also ensure that the behavior of existing network policies or Cilium network policies is not regressed.
```
joestringer, line 35: nit, there's a lot of words here that are not giving an overview of the solution. This can make it harder for other contributors to understand the proposal.
```suggestion
Allow cloud providers to block egress traffic to certain endpoints without blocking all. This proposal considers to solve this via bpf policy rules as it will work for both legacy host routing and ebpf host routing modes.
```

cilium/CFP-26135-default-egress-allow.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Option 1 (CommandLine Flag)
+The decision to pass input as command line was taken based on what is more appropriate for this scenario. There is a section below that describes the reason for going with this option.
```
joestringer, line 38: nit, I'd suggest describing the options *before* guiding the reader that an option has been pre-selected. This is good context, but it should probably come later (end of subsection) to help describe the motivations behind selecting this approach.

cilium/CFP-26135-default-egress-allow.md
```diff
@@ -0,0 +1,198 @@
...
+ 
+A new command line flag, --deny-tuples=[ip:port:protocol] could be added as stringSlice. Cloud providers would customize this cilium config to specify apriori that traffic from Cilium-managed Pods towards specific address+port+protocol combinations should be blocked. This flag would effectively act independently of all of the NP, CNP, CCNP resources and enforce a deny for such tuples regardless of the user-defined policy. The deny tuples from cilium config are sanitized and converted to cilium egress deny policy rules. Then these rules should get added to policy map of all endpoints. Several options are considered to implement this and the following section would talk about each option
```
joestringer, line 40: nit, minor wording about what the proposal is - it's easier for reviewers to understand that this is the concrete plan assuming that this option is taken. Otherwise this is a really good summary of this option.

```suggestion
Add a new command line flag, --deny-tuples=[ip:port:protocol] as stringSlice. Cloud providers would customize this cilium config to specify apriori that traffic from Cilium-managed Pods towards specific address+port+protocol combinations should be blocked. This flag would effectively act independently of all of the NP, CNP, CCNP resources and enforce a deny for such tuples regardless of the user-defined policy. The deny tuples from cilium config are sanitized and converted to cilium egress deny policy rules. Then these rules should get added to policy map of all endpoints. 
```

cilium/CFP-26135-default-egress-allow.md
```diff
@@ -0,0 +1,198 @@
...
+
+
+ * Option 1 is preferred as it aligns with existing code flow and does not affect existing network policies schema or behavior in any way.
+
+#### Adding new field in ruleMetadata
+Based on config received via command line, Cilium Agent would convert it to EgressDeny rule and does policyAdd. PolicyAdd will add the rules to repository and triggers policyreaction event to trigger identity allocation for labels specified in the policy. Cilium by default goes to deny mode if any policy added. To prevent this, this rule can be created with this flag `enableDefaultAllow` enabled. If a rule has this flag enabled, then cilium will never enforce default deny for the rest of traffic unless if there any egress k8s network policy or cilium network policy rule is enforced. By default, this flag is disabled and it would not regress existing network policy scenarios.
```
joestringer, line 46: > By default, this flag is disabled and it would not regress existing network policy scenarios.

This is a good goal, is there any planned tests to confirm this or do we think the existing testing coverage should give us reliable feedback about this? I guess the existing network policy tests should continue to work even if the new option is enabled, so that would be key to demonstrating this property.

cilium/CFP-26135-default-egress-allow.md
```diff
@@ -0,0 +1,198 @@
...
+None
+
+* Action
+Egress - Default Allow for all. Block access to 169.254.128.5:80 for all
+Ingress - Default Allow for all
+
+##### Example 2
+
+* Cloud Provider deny tuples 
+```-–d-eny-tuples=[“169.254.128.5;80;;]```
```
joestringer, line 68:
```suggestion
```-–deny-tuples=[“169.254.128.5;80;;]```
```

cilium/CFP-26135-default-egress-allow.md
```diff
@@ -0,0 +1,198 @@
...
+    - matchLabels:
+        role: frontend
+```
+* Action
+Egress - Allow for all except for pods with label “role: backend”. For those pods, egress allowed only to pods with “role: frontend”. Block access to 169.254.128.5:80 for all
+Ingress - Default Allow for all
+
+
+#### Alternate Options
+#### Option 2 (CCNP CRD with new field)
```
joestringer, line 91: nit: If the subsequent options are nested, they should be one step deeper in the document hierarchy. This helps readers to sort through and understand the relationship of this section in comparison to the others. 
```suggestion
### Alternate Options
#### Option 2 (CCNP CRD with new field)
```

cilium/CFP-26135-default-egress-allow.md
```diff
@@ -0,0 +1,198 @@
...
+The proposal suggests introducing a new field called "defaultAllowEgress" in the ClusterwideCiliumNetworkPolicy (CCNP). This field aims to enforce a default allow instead of the default deny for egress traffic within the cluster. By default, the defaultAllowEgress field is disabled, indicating that the default deny is in effect. The scope of the defaultAllowEgress field is at the policy level, meaning it applies to the entire policy, even if there are multiple rules defined within the same policy. Default deny for egress traffic will take precedence if egress enabled in atleast one of Network Policy or Cilium Network Policy or CCNP that does not have the defaultAllowEgress flag set.
```
joestringer, line 92:
```suggestion
Introduce a new field called "defaultAllowEgress" in the ClusterwideCiliumNetworkPolicy (CCNP). This field aims to enforce a default allow instead of the default deny for egress traffic within the cluster. By default, the defaultAllowEgress field is disabled, indicating that the default deny is in effect. The scope of the defaultAllowEgress field is at the policy level, meaning it applies to the entire policy, even if there are multiple rules defined within the same policy. Default deny for egress traffic will take precedence if egress enabled in at least one of Network Policy or Cilium Network Policy or CCNP that does not have the defaultAllowEgress flag set. 
```

cilium/CFP-26135-default-egress-allow.md
```diff
@@ -0,0 +1,198 @@
...
+  endpointSelector: {}
+  egressDeny:
+  - toCIDR:
+    - 169.254.128.5
+    toPorts:
+    - ports:
+      - port: "80"
+        protocol: "TCP"
+  defaultAllowEgress: true
+```
```
joestringer, line 113: This example is presented without context. It's good to see the proposal in context of the policy, but without wording that describes what it does, it's hard to reason about the semantics.

cilium/CFP-26135-default-egress-allow.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Key Question: How cloud provider deny rules are processed with network policy rules for same endpoint?
+There is no change to default deny behavior of existing network policies even if cloud provider specific deny rules configured via cilium config.  As EgressDeny rules are processed first, deny rules from cloud provider are evaluated first and rest of rules are evaluated later as like other network policy egress deny rules.
+
+### Key Question: Why not iptable rules instead of eBPF?
+cilium adds a rule on top of the FORWARD chain to redirect all traffic to a CILIUM_FORWARD chain. So any Cloud Provider specific firewall rules placed after cilium rules doesn't get executed. Disabling the "prepend-iptables-chains" option is also not a viable solution as disabling prepend means Cilium can no longer guarantee that its policies are enforced. Also in eBPF host routing mode, Cilium bypasses IPtable rules and rules can be enforced only via eBPF.
```
joestringer, line 146: nit: These are good context, but for context usually the goal of "Key Questions" is to ask the reviewers to consider the options and provide input on the solutions. These two seem to be more of FAQ items which are probably more suited to the implementatino section.

cilium/CFP-26135-default-egress-allow.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Key Question: How cloud provider deny rules are processed with network policy rules for same endpoint?
+There is no change to default deny behavior of existing network policies even if cloud provider specific deny rules configured via cilium config.  As EgressDeny rules are processed first, deny rules from cloud provider are evaluated first and rest of rules are evaluated later as like other network policy egress deny rules.
+
+### Key Question: Why not iptable rules instead of eBPF?
+cilium adds a rule on top of the FORWARD chain to redirect all traffic to a CILIUM_FORWARD chain. So any Cloud Provider specific firewall rules placed after cilium rules doesn't get executed. Disabling the "prepend-iptables-chains" option is also not a viable solution as disabling prepend means Cilium can no longer guarantee that its policies are enforced. Also in eBPF host routing mode, Cilium bypasses IPtable rules and rules can be enforced only via eBPF. 
+
+
+### Considered Option (CommandLine Flag/Cilium Config):
```
joestringer, line 149: nit:
```suggestion
### Key Question: Which option should be taken to implement the feature?
```

joestringer (commented):
I think that Option 1 is a reasonable tradeoff right now. I see this as the first step to having a stable interface into this sort of configuration. At some point, probably Options (2) and/or (4) will make sense, and those could build on top of the changes that are made to solve this use case under Option (1).

I had some minor wording nits below but it's mostly just to tidy this up and make it easier for someone fresh to the CFP to read and understand the proposal. None of it is substantial to the proposals.

joestringer:
Oh, I meant to highlight that although I saw some example scenarios and how they are expected to perform, there wasn't a dedicated section to describe the intended testing scenarios. For the PR submission we'd need regression tests.

tamilmani1989:
> Oh, I meant to highlight that although I saw some example scenarios and how they are expected to perform, there wasn't a dedicated section to describe the intended testing scenarios. For the PR submission we'd need regression tests.

yep definitely. I'm not aware that I should mention about regression tests in CFP. Let me add separate section for tests.

cilium/CFP-26135-default-egress-allow.md
tamilmani1989: yes will create a separate test with this option enabled and existing network policy should continue to work.

joestringer:
As per https://github.com/cilium/cilium/pull/30060#issuecomment-1955250995, I think that there's two parts to the proposal here:
1. Provide a "non-default-deny" semantic in the policy (now resolved by https://github.com/cilium/design-cfps/pull/16)
2. Allow providers to create such a policy by default, for instance via commandline argument.

For the latter, if you think it would be useful to have a detailed discussion about the form of the solution then we can either repurpose this CFP or create a fresh PR for the CFP to solve that part of the problem. Alternatively I think it's potentially small enough that a CFP is no longer necessary, you could put together a PoC PR on the cilium/cilium repo and we could discuss further from there.

tamilmani1989:
@squeed and I discussed to have a directory with static network policies placed. Agent mounts this directory and load policies on start up . Directory location can be passed via cilium config. I have to do POC on this and will open up PR.

PR #14, opened 2024-01-17
Adding xDS Adapter CFP

robscott:
This is a follow up from the [original proposal doc](https://docs.google.com/document/d/1U4pO_dTaHERKOtrneNA8njW19HSVbq3sBM3x8an4878/edit) I wrote and corresponds with https://github.com/cilium/cilium/issues/30235.

Note: I'm not quite sure how to fill out impacts and key questions yet, it's possible those will come more naturally with more review of this proposal?

/cc @joestringer @youngnick

### DIFF
--- a/cilium/CFP-30235-xds-adapter.md
+++ b/cilium/CFP-30235-xds-adapter.md
@@ -1,0 +1,204 @@
+# CFP-30235: xDS Adapter for ClusterIP Routing
+
+**SIG:** Agent
+
+**Begin Design Discussion:** 2023-01-03
+
+**Cilium Release:** 1.16
+
+**Authors:** Rob Scott <robertjscott@google.com>
+
+## Summary
+
+Add a new xDS adapter in Cilium that could take advantage of some of the
+strengths of xDS, particularly the feedback loop via LRS and the overall
+potential for scalability improvements when adjustments to routing
+configurations don't need to round trip through the Kubernetes API Server.
+This adapter would be an alternative source of endpoints and would not replace
+the existing default behavior of reading directly from Kubernetes APIs.
+
+## Motivation
+
+There are some capabilities that the Kubernetes API Server is not well equipped
+to handle. This includes topology aware routing (or any form of routing that
+requires a feedback loop) and routing to endpoints outside of the cluster.
+
+## Goals
+
+* Describe how an xDS adapter could work in Cilium
+* Highlight new capabilities that xDS could enable
+
+## Non-Goals
+
+* Build anything additional on top of this xDS adapter. Although this could be a
+  foundational technology for several new features, those are out of scope for
+  this proposal.
+
+## Background
+
+Several years ago, Matt Klein published a blog post charting a vision for xDS as
+a [universal data plane
+API](https://blog.envoyproxy.io/the-universal-data-plane-api-d15cec7a). Although
+these APIs are primarily used by Envoy and gRPC today, there has been a broader
+goal in the community to enable more widespread usage as part of a truly
+universal data plane API. xDS is CNCF-governed and is gradually transitioning
+content from the
+[envoyproxy/data-plane-api](https://github.com/envoyproxy/data-plane-api) repo
+to [cncf/xds](https://github.com/cncf/xds) repo as part of that overall vision.
+
+In parallel to these efforts by the xDS community, GKE is planning to introduce
+xDS as an additional data source for DPv2 configuration. This feels sufficiently
+generic and helpful that it could be something that could be contributed to
+upstream Cilium. This could be particularly useful for at least two common use
+cases:
+
+1. Supporting Services and Endpoints from outside of the local cluster.
+1. Supporting advanced routing techniques, such as topology aware routing.
+
+Although it is out of scope for this specific CFP to provide complete solutions
+for either of these use cases, it will demonstrate the benefits of having an
+xDS adapter when developing a solution either of these use cases.
+
+### Use Cases
+
+#### Topology Aware Routing
+
+As the author of both EndpointSlices and the existing Topology Aware Routing
+approach in Kubernetes, I feel like I’m well equipped to discuss the limitations
+of both. What most people want from Topology Aware Routing is essentially to
+fill the local zone to capacity and then spillover to the next closest set of
+endpoints with capacity. Unfortunately that is very difficult to achieve with
+the existing Kubernetes tooling. There are a few key problems here:
+
+1. Kubernetes has no feedback loop to its dataplane and therefore no reliable
+   way to understand if endpoints have reached capacity.
+2. Even if there were a feedback loop, spillover to other zones benefits from
+   some form of centralized orchestration to avoid a thundering herd problem.
+3. All changes to endpoint routing currently have to go through Services or
+   EndpointSlices, and by extension, the Kubernetes API Server. It would be very
+   expensive to implement incremental weighting/spillover adjustments with that
+   overhead. For example, adjusting the weight of any individual endpoint would
+   require a write to the EndpointSlice API which would then need to be
+   distributed to all consumers of that API. That can be problematic from a
+   scalability perspective when there are frequent updates paired with a large
+   number of endpoints and/or nodes.
+
+I believe that xDS is uniquely positioned to address these limitations.
+[LRS](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/load_stats/v3/lrs.proto)
+(Load Reporting Service) provides a straightforward way to provide load reports
+to a centralized control plane. Similarly, [Delta
+xDS](https://www.envoyproxy.io/docs/envoy/latest/configuration/overview/xds_api#delta-endpoints)
+enables xDS to distribute incremental updates that can be smaller in scope than
+comparable Kubernetes APIs. All of this could be combined to avoid adding
+additional load on the Kubernetes API Server.
+
+#### Endpoints From Other Sources
+
+In some cases, it’s useful to be able to route to endpoints from other sources,
+such as other Kubernetes clusters or outside of Kubernetes altogether. This can
+be particularly relevant when implementing Multi-Cluster Service routing.
+Implementations today often rely on one of the following approaches:
+
+1. Mirror endpoints from other clusters into Kubernetes by creating custom
+   Services and EndpointSlices.
+2. Introduce Gateways on the edges of clusters that can be used to route
+   requests from other clusters.
+
+With xDS support, it could be possible to have a centralized xDS control plane
+that was familiar with endpoints in multiple sources, providing a potentially
+more efficient alternative to either of the approaches commonly used today.
+
+## Proposal
+
+### Logic
+
+The introduction of this adapter would mean that Cilium could accept endpoints
+from more than one source. Cilium could be configured to read exclusively from
+Kubernetes (default), xDS, or both, depending on the use case.
+
+When both data sources were in use, precedence in any conflicts would be given
+to data received directly from Kubernetes. As a concrete example, it's possible
+that the same Service could be represented in both Kubernetes and xDS. If that
+were the case, we'd give precedence to the Service from Kubernetes APIs and drop
+the one from xDS, likely surfacing some kind of warning along the way.
+
+### Architecture
+
+This xDS adapter would be similar conceptually to KVStore. Essentially this
+would be another backend that Cilium could get data from. Similar to how Cilium
+can read from Kubernetes APIs and/or KVStore today, we would add a third option
+for xDS as a data source, and develop a shared interface that worked across all
+data sources.
+
+### API Mapping
+
+To keep the initial work as focused as possible, I propose using existing xDS
+APIs to model ClusterIP routing. In the future, we may want to consider either
+additions to those existing xDS APIs or new xDS APIs to cover the entirety of
+Cilium functionality. To start, it seems highly valuable to focus on the areas
+where existing xDS APIs overlap with existing Cilium capabilities.
+
+This shows how we could use existing xDS APIs to represent Cilium capabilities
+for ClusterIP routing:
+
+| Cilium Field | Origin xDS | Comments |
+| - | - | - |
+| Frontend.Address | `filter_chain_match.prefix_ranges` | |
+| Frontend.Port | `filter_chain_match.destination_port` | |
+| Frontend.Protocol | `cluster.metadata.protocol` | |
+| Backend[*].FEPortName | `cluster.metadata.port_name` | |
+| Backend[*].NodeName | ? | TBD - This is only used for logic to determine if an endpoint is local. Instead of sending this through xDS for every endpoint we could just look at the CIDR configured for each Cilium instance. |
+| Backend[*].Addr, Backend[*].Port | `lb_endpoints.endpoint.address.socket_address` | |
+| Backend[*].State | `lb_endpoints.health_status` | |
+| Backend[*].Preferred | N/A | Not used in Kubernetes |
+| Backend[*].Weight | `lb_endpoints.endpoint.load_balancing_weight` | |
+| Type | N/A | Out of scope - only focusing on ClusterIP Services |
+| ExternalTrafficPolicy, InternalTrafficPolicy | `cluster.load_balancing_policy` | Defined in the typed struct `type.googleapis.com/cilium_io_traffic_policy` |
+| NatPolicy | N/A | Not used in Kubernetes |
+| SessionAffinity | `tcpproxy.hash_policy` | |
+| SessionAffinityTimeout | ? | TBD - This represents the maximum session stickiness time. There is no natural place for this in existing xDS APIs. |
+| HealthCheckNodePort | N/A | Out of scope - only focusing on ClusterIP Services |
+| Name | `cluster.metadata.service_name` | |
+| Namespace | `cluster.metadata.service_namespace` | |
+
+## Impacts / Key Questions
+
+_List crucial impacts and key questions. They likely require discussion and are required to understand the trade-offs of the CFP. During the lifecycle of a CFP, discussion on design aspects can be moved into this section. After reading through this section, it should be possible to understand any potentially negative or controversial impact of this CFP. It should also be possible to derive the key design questions: X vs Y._
+
+### Impact: ... 1
+
+_Describe crucial impacts and key questions that likely require discussion and debate._
+
+### Key Question: ... 2
+
+_Describe a key question_
+
+### Option 1:
+
+#### Pros
+
+* ...
+
+#### Cons
+
+* ...
+
+### Option 2:
+
+#### Pros
+
+* ...
+
+#### Cons
+
+* ...
+
+## Future Milestones
+
+_List things that this CFP will enable but that are out of scope for now. This can help understand the greater impact of a proposal without requiring to extend the scope of a CFP unnecessarily._
+
+### Deferred Milestone 1
+
+_Description of deferred milestone_
+
+### Deferred Milestone 2


cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,204 @@
...
+API](https://blog.envoyproxy.io/the-universal-data-plane-api-d15cec7a). Although
+these APIs are primarily used by Envoy and gRPC today, there has been a broader
+goal in the community to enable more widespread usage as part of a truly
+universal data plane API. xDS is CNCF-governed and is gradually transitioning
+content from the
+[envoyproxy/data-plane-api](https://github.com/envoyproxy/data-plane-api) repo
+to [cncf/xds](https://github.com/cncf/xds) repo as part of that overall vision.
+
+In parallel to these efforts by the xDS community, GKE is planning to introduce
+xDS as an additional data source for DPv2 configuration. This feels sufficiently
```
mikemorris, line 50: > DPv2

Has any public high-level summary been written about this that could be linked here?

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,204 @@
...
+
+There are some capabilities that the Kubernetes API Server is not well equipped
+to handle. This includes topology aware routing (or any form of routing that
+requires a feedback loop) and routing to endpoints outside of the cluster.
+
+## Goals
+
+* Describe how an xDS adapter could work in Cilium
+* Highlight new capabilities that xDS could enable
+
```
linsun, line 30: Would a requirement/goal be the proposal could be extended for other scenarios using xDS without major changes?

I echo comments from @mikemorris on slack: `the intersection is relevant to the extent  that we just try to avoid boxing ourselves in by making the ClusterIP xDS routing too narrow/bespoke such that it would prevent further extensibility later`

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,204 @@
...
+| Backend[*].Weight | `lb_endpoints.endpoint.load_balancing_weight` | |
+| Type | N/A | Out of scope - only focusing on ClusterIP Services |
+| ExternalTrafficPolicy, InternalTrafficPolicy | `cluster.load_balancing_policy` | Defined in the typed struct `type.googleapis.com/cilium_io_traffic_policy` |
+| NatPolicy | N/A | Not used in Kubernetes |
+| SessionAffinity | `tcpproxy.hash_policy` | |
+| SessionAffinityTimeout | ? | TBD - This represents the maximum session stickiness time. There is no natural place for this in existing xDS APIs. |
+| HealthCheckNodePort | N/A | Out of scope - only focusing on ClusterIP Services |
+| Name | `cluster.metadata.service_name` | |
+| Namespace | `cluster.metadata.service_namespace` | |
+
```
linsun, line 163: Should we list the https://github.com/istio/ztunnel/blob/master/proto/workload.proto as an alternative per @louiscryan's comment on slack?

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,204 @@
...
+were the case, we'd give precedence to the Service from Kubernetes APIs and drop
+the one from xDS, likely surfacing some kind of warning along the way.
+
+### Architecture
+
+This xDS adapter would be similar conceptually to KVStore. Essentially this
+would be another backend that Cilium could get data from. Similar to how Cilium
+can read from Kubernetes APIs and/or KVStore today, we would add a third option
+for xDS as a data source, and develop a shared interface that worked across all
+data sources.
```
shashankram, line 131: Worth sharing in more detail what this would entail in terms of refactoring the agent functionality to support xDS as an alternative store.
Further, since LRS enables the client to share load information with the server, how would this information be surfaced and interact with rest of the agent functionality, given this capability is unavailable in the agent today?

shashankram (commented):
Overall this seems like a reasonable starting point to enable xDS

cilium/CFP-30235-xds-adapter.md
robscott: Good question, added a "key question" about this + a reference to WDS.

cilium/CFP-30235-xds-adapter.md
robscott: Good point - added the following goal: "Provide room for future growth in xDS support beyond ClusterIP routing."

cilium/CFP-30235-xds-adapter.md
robscott: Sadly don't have anything public yet, we're still in pretty early stages.

cilium/CFP-30235-xds-adapter.md
robscott: I tried to address this in the latest update, adding a new section for LRS, but let me know if I missed anything.

cilium/CFP-30235-xds-adapter.md
linsun: Thank you Rob!!

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+plane via LRS.
+
+## Impacts / Key Questions
+
+### Key Question: Should we use a custom API on top of xDS-TP instead?
+Instead of building on top of existing xDS APIs, we could use a custom API on
+top of xDS-TP (xDS Transport Protocol). This would follow the pattern that
+Istio's Ztunnel project used when they developed
+[WDS](https://github.com/istio/ztunnel/blob/db0a74212c42c66b611a71a9613afb501074e257/proto/workload.proto).
+We could also just choose to send Cilium-specific protos over the wire.
```
shashankram, line 194: There's also the option of using existing Cilium agent API JSON types encoded using the proto Struct type, such that the adapter would be responsible for unmarshalling it to the Cilium API Go types (via JSON). Could you add this?

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+
+#### Pros
+* Very flexible, can specify ~anything
+* Resulting API would likely be very focused on Cilium
```
shashankram, line 198: Extensibility is a big benefit here. Worth noting.

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
+# CFP-30235: xDS Adapter for ClusterIP Routing
+
+**SIG:** Agent
+
+**Begin Design Discussion:** 2023-01-03
```
linsun, line 5:
```suggestion
**Begin Design Discussion:** 2024-01-03
```

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+#### Topology Aware Routing
+
+As the author of both EndpointSlices and the existing Topology Aware Routing
+approach in Kubernetes, I feel like I’m well equipped to discuss the limitations
+of both. What most people want from Topology Aware Routing is essentially to
+fill the local zone to capacity and then spillover to the next closest set of
+endpoints with capacity. Unfortunately that is very difficult to achieve with
+the existing Kubernetes tooling. There are a few key problems here:
+
+1. Kubernetes has no feedback loop to its dataplane and therefore no reliable
```
aanm, line 77: Just to be clear, in Cilium, an "xDS adapter" alone doesn't solve this. We will also need changes into Cilium's control plane to propagate its internal state back to the xDS adapter.

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+2. Even if there were a feedback loop, spillover to other zones benefits from
+   some form of centralized orchestration to avoid a thundering herd problem.
+3. All changes to endpoint routing currently have to go through Services or
+   EndpointSlices, and by extension, the Kubernetes API Server. It would be very
+   expensive to implement incremental weighting/spillover adjustments with that
+   overhead. For example, adjusting the weight of any individual endpoint would
+   require a write to the EndpointSlice API which would then need to be
+   distributed to all consumers of that API. That can be problematic from a
+   scalability perspective when there are frequent updates paired with a large
+   number of endpoints and/or nodes.
```
aanm, line 88: I don't understand this problem. Kubernetes API Server is the "control plane" of the cluster, won't we have this same problem for any "control plane" that receives the load reports from Cilium?

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+interface for interacting with data sources (Kubernetes, KVStore, and xDS). This
+interface will enable the majority of the codebase to be entirely unaware of
+which data sources are in use.
+
+A key part of defining this interface will involve determining a precedence
+order when more than one data source is connected. When any conflicts occur,
+for example if Services from different sources have the same IP, the following
+order will be used for precedence:
+
+1. Kubernetes
```
aanm, line 132: `0. local API` - It's possible to add endpoints and services using the local API.

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+| Type | N/A | Out of scope - only focusing on ClusterIP Services |
+| ExternalTrafficPolicy, InternalTrafficPolicy | `cluster.load_balancing_policy` | Defined in the typed struct `type.googleapis.com/cilium_io_traffic_policy` |
+| NatPolicy | N/A | Not used in Kubernetes |
+| SessionAffinity | `tcpproxy.hash_policy` | |
+| SessionAffinityTimeout | ? | TBD - This represents the maximum session stickiness time. There is no natural place for this in existing xDS APIs. |
+| HealthCheckNodePort | N/A | Out of scope - only focusing on ClusterIP Services |
+| Name | `cluster.metadata.service_name` | |
+| Namespace | `cluster.metadata.service_namespace` | |
+
+### 3. Sample xDS Control Plane
```
aanm, line 173: Food for thought: it might make sense to add this xDS Control plane into Cilium Operator.

linsun: +1 if the community decides to provide this to Cilium users.

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+**Cilium Release:** 1.16
+
+**Authors:** Rob Scott <robertjscott@google.com>
+
+## Summary
+
+Add a new xDS adapter in Cilium that could take advantage of some of the
+strengths of xDS, particularly the feedback loop via LRS and the overall
+potential for scalability improvements when adjustments to routing
+configurations don't need to round trip through the Kubernetes API Server.
```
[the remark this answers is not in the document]
youngnick: It seems like one of the major advantages of an xDS control plane managing this information is that it's _only_ handling this information, rather than being a generic workload-management API like Kubernetes.

For some cases, I don't it will ever make sense to put xDS between Cilium and Kubernetes, but for some use cases, particularly endpoints and endpoint grouping (Clusters in xDS, Services in Kubernetes), it seems more straightforward to map these to xDS objects.

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+  data sources.
+* Provide room for future growth in xDS support beyond ClusterIP routing.
+* Allow the use of any combination of independent data sources (Kubernetes,
+  KVStore, xDS) with Cilium.
+
+## Non-Goals
+
+* Build anything additional on top of this xDS adapter. Although this could be a
+  foundational technology for several new features, those are out of scope for
+  this proposal.
```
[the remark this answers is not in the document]
youngnick: I agree that to validate that this works, we'll need an xDS control plane to talk to. Which is a not-insignificant engineering problem in itself.

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+to [cncf/xds](https://github.com/cncf/xds) repo as part of that overall vision.
+
+In parallel to these efforts by the xDS community, GKE is planning to introduce
+xDS as an additional data source for DPv2 configuration. This feels sufficiently
+generic and helpful that it could be something that could be contributed to
+upstream Cilium. This could be particularly useful for at least two common use
+cases:
+
+1. Supporting Services and Endpoints from outside of the local cluster.
+1. Supporting advanced routing techniques, such as topology aware routing.
```
youngnick, line 60: Seems from the LRS stuff that something like:
```
1. Using load information obtained via bidirectional xDS to steer traffic
```
might be useful here? I don't think that comes under "topology", since topology is inherent to the structure of the network, but the load is contingent on external factors.

cilium/CFP-30235-xds-adapter.md
youngnick: This is where the bidirectional nature of xDS comes in a bit more handy - every xDS communication is a message and response pair, unlike the Kubernetes API where each operation is _either_ a write or a read. The design of the Load Reporting Service uses this to its advantage, effectively making the control plane a _client_ instead of the server, since one "give me your load numbers" request will produce a stream of load numbers (that the control plane will simply acknowledge).

That said, I think it's important to remember that actually building these control planes is _full_ of really hard concurrency problems. More on that under the "incremental" section in a few lines.

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+   number of endpoints and/or nodes.
+
+I believe that xDS is uniquely positioned to address these limitations.
+[LRS](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/load_stats/v3/lrs.proto)
+(Load Reporting Service) provides a straightforward way to provide load reports
+to a centralized control plane. Similarly, [Delta
+xDS](https://www.envoyproxy.io/docs/envoy/latest/configuration/overview/xds_api#delta-endpoints)
+enables xDS to distribute incremental updates that can be smaller in scope than
+comparable Kubernetes APIs. All of this could be combined to avoid adding
+additional load on the Kubernetes API Server.
```
youngnick, line 97: I agree that LRS is designed to do exactly this, but implementing incremental xDS is _not_ straightforward. Unless your control plane is very carefully designed, it's nearly impossible (which is why many simpler Envoy control planes don't do it.)

To be clear, the hard part is on whoever's building the control plane, not inside the Cilium Agent. But given that we'll need an open-source thing to test with at the very least, the engineering cost of building this shouldn't be underestimated.

cilium/CFP-30235-xds-adapter.md
youngnick: yes, I think that the first part of this effort needs to be to understand the existing internal model for endpoints and endpoint groups and how well it aligns with the other models (including xDS).

cilium/CFP-30235-xds-adapter.md
youngnick: I'm -1 on adding it into the Operator, personally, because the operator is currently tightly focussed on its current role: reading from Kubernetes objects, building some internal model, then writing back to Kubernetes. The Operator doesn't have any actual routing logic itself.

An xDS control plane will need to have a lot of routing logic added. I think it's better to keep this as a separate thing that's built with the same tooling as the Agent and Operator (that is, using the Hive and Cell pattern), so that it can be merged in later if necessary.

shashankram: Seems reasonable to have a standlone component that implements the control plane. Adding it to the operator has the benefit of not requiring a new component and it can simply be feature gated. Though from a stability perspective, keeping this out of operator is beneficial.

cilium/CFP-30235-xds-adapter.md
joestringer: Discussing *not* running this in production gives me a bit of concern. How will Cilium users benefit from this functionality if the reference implementation isn't built with production in mind?

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+**Begin Design Discussion:** 2023-01-03
+
+**Cilium Release:** 1.16
+
+**Authors:** Rob Scott <robertjscott@google.com>
+
+## Summary
+
+Add a new xDS adapter in Cilium that could take advantage of some of the
+strengths of xDS, particularly the feedback loop via LRS and the overall
```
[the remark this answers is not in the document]
joestringer: x-ref https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_reporting_service (may want to include the link in the doc)

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+Implementations today often rely on one of the following approaches:
+
+1. Mirror endpoints from other clusters into Kubernetes by creating custom
+   Services and EndpointSlices.
+2. Introduce Gateways on the edges of clusters that can be used to route
+   requests from other clusters.
+
+With xDS support, it could be possible to have a centralized xDS control plane
+that was familiar with endpoints in multiple sources, providing a potentially
+more efficient alternative to either of the approaches commonly used today.
```
joestringer, line 113: What is it about this centralized xDS control plane that is more efficient?

Probably worth also contrasting with kvstoremesh (see CiliumCon or https://arthurchiao.art/blog/trip-first-step-towards-cloud-native-security/#222-kvstoremesh)

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+implementation with Cilium. Of course there will likely be multiple options
+here, but it will be helpful to have at least a reference implementation for the
+purposes of e2e testing.
+
+### 4. Report Metrics via LRS
+One of the significant benefits of xDS is that it contains a load reporting
+mechanism. This can enable advanced routing capabilities at the xDS control
+plane layer, such as the topology aware routing example mentioned above. Within
+Cilium we'd likely want to report new and active connections to an xDS control
+plane via LRS.
```
joestringer, line 185: Just to note (partly coming from the session we had earlier in the week): I think it makes sense to clearly outline what sort of stats we are looking for, format, etc. and also acknowledge that this may require datapath changes. I suspect that this may end up being a logically fairly separate change that could have other debuggability benefits even in existing environments if exposed through prometheus or other relevant interfaces.

cilium/CFP-30235-xds-adapter.md
joestringer, line 1: > Note: I'm not quite sure how to fill out impacts and key questions yet, it's possible those will come more naturally with more review of this proposal?

Yes, that's a big part of the intent. Capture aspects of the proposal that may not end up being directly encoded in the design, but which influence the design.

Impacts are the sorts of tradeoffs that we discover during writeup, discussion, evaluation of the ideas. For instance maybe there'll be an increase in resource usage and we should take that into account while considering whether the balance of functionality vs. cost is worthwhile. Or maybe there are additional API stability commitments the community will need to make. The impacts may be mitigated by choices made during the CFP preparation, in which case these mitigations are worth mentioning as well.

Key questions will often arise through review. Perhaps there is a fork in the road where there are two possible choices and they both have pros/cons. Key questions are intended to facilitate the discussion during the review so that we acknowledge the good and bad things about an aspect of the proposal, consider severity and mitigations, and come to a solution that we believe does not outweigh the benefits of the overall proposal.

(Rough guidelines, I don't believe we have an official writeup on this yet so these may not comprehensively capture the intent behind those sections. But hopefully this gives you a sense of the intent + how to benefit from this structure in the CFP)

cilium/CFP-30235-xds-adapter.md
joestringer: LRS providing feedback on top can only be a net increase in load. I could see maybe an argument there that kube-apiserver protocols are not as well suited to the problem, but before we even consider handling the load feedback, what about the existing load of just configuring K8sEndpoints on all of the cilium-agents?

Perhaps some guiding questions:
- In the target deployment scenario, what percentage of kube-apiserver resources (CPU,memory,network) usage corresponds to service/endpoint handling? Is the bottleneck having a big enough kube-apiserver node, and the consideration is that we can increase the available CPU by moving this functionality to a secondary node with an additional 2x, 5x, 10x CPU to process the events? That's an argument that moving this service handling to a dedicated process can gain some multiplier of available resources to handle the scale.
- In the proposed design, is there something inherent that reduces the number of requests/responses that are being sent around the cluster? Let's say for a ballpark number the naive assumption is that if today's cluster has 1K nodes, then you have 1K events per endpoint update. If you move the service handling to a second process, then kube-apiserver as source of truth will reduce down to only transmitting one event to the new service control plane manager, but then the service control plane manager is responsible for (a) the original 1K events and now also (b) marshalling/unmarshalling between disparate protocols.
- How does this impact availability?

cilium/CFP-30235-xds-adapter.md
shashankram: To clarify my previous comment, I am not suggesting we should build an experimental control plane. Instead, we should agree on the scope of the work as an initial deliverable.  Building a production level control plane that handles all the edge cases can be challenging. There should definitely be a reference implementation that is usable outside of tests.

cilium/CFP-30235-xds-adapter.md
shashankram: https://github.com/cilium/cilium/issues/30283 describes the benefits of centralization in a different context. Just wanted to share this for reviewers not aware of the adjacent proposal.

cilium/CFP-30235-xds-adapter.md
robscott: I think the minimum bar here in terms of initial development would be for a OSS/CNCF reference xDS control plane that is sufficiently reliable to run for e2e testing and development and also at least one that is production ready. Those will probably be the same project, but if we are successful at building a good ecosystem here, it's possible that there will be multiple OSS production-ready options and we won't need the reference/testing implementation to also be production ready. Our e2e tests here could become something resembling conformance tests, but for control plane implementations.

cilium/CFP-30235-xds-adapter.md
robscott: Added, thanks!

cilium/CFP-30235-xds-adapter.md
robscott: Added, thanks!

cilium/CFP-30235-xds-adapter.md
robscott: 🤦  Fixed, thanks!

cilium/CFP-30235-xds-adapter.md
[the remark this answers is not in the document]
robscott: Updated, thanks!

cilium/CFP-30235-xds-adapter.md
robscott: +1, this just provides the foundation to solve this. The whole idea of topology aware routing or really any form of routing that requires a feedback loop is deceptively simple until you start to think through this part of it. What xDS gives us here is an established API + patterns for completing a feedback loop, but like you're saying, we'd still need to connect the dots here.

cilium/CFP-30235-xds-adapter.md
robscott: Let me preface this by saying that scalability was not my primary goal with this CFP, but I do think it could be a nice coincidental benefit of the approach even outside of load reporting. When we get to load reporting, I can't think of any feasible way to do it with existing k8s constructs, and I'd rather not invents something new here when LRS already exists.

I spent a fair amount of time thinking about the Kubernetes endpoint scalability problem as I was working on the EndpointSlice API, and went as far as scale testing large k8s clusters until they broke to really understand what was happening ([relevant slides](https://kccnceu20.sched.com/event/Zeta/scaling-kubernetes-networking-beyond-100k-endpoints-rob-scott-minhan-xia-google)). This is a relatively unique problem in terms of Kubernetes APIs. Here's what we have to deal with:

1. EndpointSlices are updated frequently (essentially any time Pod status changes)
2. Every update needs to be distributed to every node in the cluster
3. There are often thousands of endpoints in a cluster

Some Kubernetes APIs have to deal with the scale of EndpointSlices but are only consumed in one or two places - ie some kind of centralized controller(s). The Pod API is consumed by every Node thanks to Kubelet, but Kubelet is able to filter only the Pods that are local to it, meaning that each Pod is distributed to exactly one Node. 

Let's consider a rolling update of a deployment with 100 Pods. Each of those Pods is going to go through a transition of unready -> ready while the old 100 Pods are transitioning from ready -> unready. Let's say that translates to 400 distinct events to Process (old Pod -> unready, old Pod -> terminated, new Pod -> unready, new Pod -> ready) * 100. In a naive implementation, that would mean that Kubernetes API Server would need to transmit 400 EndpointSlice updates to _every_ node in the cluster. That becomes especially fun when you consider that some providers support [15,000 Nodes per cluster](https://cloud.google.com/kubernetes-engine/quotas). 

Fortunately the EndpointSlice controller uses batching to mitigate that to some extent, but hopefully that sets the stage for the unique problem that endpoints pose to the Kubernetes API Server. Like @joestringer mentions above, from the perspective of the API Server, it would be amazing if it didn't need to distribute all of those updates to each individual Node.

As far as LRS specifically, +1 to everything @youngnick mentioned above. Load reporting would be especially complex with existing Kubernetes constructs but relatively straightforward with LRS. Everything that goes through Kubernetes API Server would need to be persisted through writes to etcd. On the other hand, LRS combined with an xDS control plane could enable us to bypass that entirely.

>In the proposed design, is there something inherent that reduces the number of requests/responses that are being sent around the cluster?

I don't think this necessarily does that. It may have some marginal impact here, but I don't think it will be a huge one. In my opinion, a lot of the value here would be that we'd have a path to moving this load from the API Server to a separate control plane that could be scaled independently and potentially more optimized for this specific purpose. My proposal is not really focused on scalability, but I think it provides the potential for significant improvements here. For example, you might choose to deploy an xDS control plane per-zone as a solution that might improve both availability and scalability. 

> How does this impact availability?

So in the base case where people just don't use this feature, I think there's no impact. Assuming they do, I think it entirely depends on how xDS control planes are deployed. Using the example above, if there were a separate instance of a control plane in each zone it _may_ improve overall availability. On the other hand, if you're adding a single xDS control plane instance to a cluster that has 3 API Server replicas, you might decrease availability.

cilium/CFP-30235-xds-adapter.md
robscott: +1, phase 1 in this proposal will be to define a common interface that can be used for all Service and Endpoint data sources. I think that _mostly_ exists, but it would be good to ensure that _everything_ is going through the same shared path here with an interface we all agree on before going any further. I also added the local API to the list as suggested by @aanm above.

cilium/CFP-30235-xds-adapter.md
```diff
@@ -0,0 +1,213 @@
...
+#### API Mapping
+To keep the initial work as focused as possible, I propose using existing xDS
+APIs to model ClusterIP routing. In the future, we may want to consider either
+additions to those existing xDS APIs or new xDS APIs to cover the entirety of
+Cilium functionality. To start, it seems highly valuable to focus on the areas
+where existing xDS APIs overlap with existing Cilium capabilities.
+
+This shows how we could use existing xDS APIs to represent Cilium capabilities
+for ClusterIP routing:
+
```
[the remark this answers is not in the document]
robscott: This is currently referring to Cilium's Service cache, but I expect that at least some details may change a bit if we're building a common interface around that as part of phase 1: https://github.com/cilium/cilium/blob/0632058f820c05013dfcb010d6bda0911b1269d7/pkg/service/service.go#L73-L97

cilium/CFP-30235-xds-adapter.md
robscott: Good point, added. It was always a hope that this kind of load reporting would enable other forms of traffic steering here, added this a use case so it's clear.

cilium/CFP-30235-xds-adapter.md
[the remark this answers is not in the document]
robscott: Yeah that's a good point. What I'd really intended here was that this CFP was meant to be about foundational building blocks/infrastructure, not features. I agree that an xDS control plane needs to be bundled with this in some way and have already added it as phase 3 of this CFP.

cilium/CFP-30235-xds-adapter.md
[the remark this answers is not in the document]
shashankram: A bit confusing with so many Service types within the code, but can this be mapped 1:1 to loadbalancer.SVC?

DerekTBrown:
Could you provide some more detail around where the proposed xDS Adapter would sit in the Cilium stack? I could see how this would inform some of the interface API design decisions. 

In particular, I would like to hear your thoughts around how the xDS Adapter would interact with the embedded Envoy instance. For instance, does the xDS adapter proxy/cache requests from Envoy to an upstream server, or is Envoy communicating directly with upstream xDS servers?

DerekTBrown:
One other thought: Would it make sense to explore an approach that supports only a single backend (rather than trying to reconcile between several sources). Basically:
- Cilium would have either a KVStore/Local API/xDS Mode
- There could be external tooling that would help migrate between modes (eg. an xDS implementation that reads from existing Cilium CRs).

mikemorris:
> Would it make sense to explore an approach that supports only a single backend

While https://github.com/cilium/cilium/issues/30283 is a separate proposal to build a full alternative xDS interface, my understanding is that the narrow scope of this proposal is part of the attraction, as it allows xDS to be used just where it provides a specific benefit, as @youngnick mentions in https://github.com/cilium/design-cfps/pull/14#discussion_r1464368619

youngnick:
> In particular, I would like to hear your thoughts around how the xDS Adapter would interact with the embedded Envoy instance. For instance, does the xDS adapter proxy/cache requests from Envoy to an upstream server, or is Envoy communicating directly with upstream xDS servers?

The xDS adapter in this CFP is _totally_ orthogonal to the built-in Envoy instance, except in so far as it would take endpoints from somewhere else, and make them available to as Cilium endpoints to be used in Envoy config like any other one. But that's a very indirect link, and code-wise they will be completely separate.

xmulligan:
@robscott we now have statuses for CFPs https://github.com/cilium/design-cfps#status. Are you still trying to move this towards implementable or is it dormant for now?

pchaigno:
FYI, there's an implementation submitted for reviews at https://github.com/cilium/cilium/pull/34484.

joestringer:
Next step on this CFP is a refresh from the latest developments, onus is on @robscott for the update. Marking draft until then.

xmulligan:
@robscott with https://github.com/cilium/cilium/pull/34484 merged, would you mind updating this design doc for merging?

PR #16, opened 2024-02-02
Non-default-deny network policies

squeed:
See contents.

Sample implementation [here](https://github.com/cilium/cilium/pull/30572).

### DIFF
--- a/cilium/CFP-30572-non-default-deny-policies.md
+++ b/cilium/CFP-30572-non-default-deny-policies.md
@@ -1,0 +1,202 @@
+# CFP-30572: Non-default-Deny Network Policy
+
+**SIG: SIG-Policy**
+
+**Begin Design Discussion:** 2024-02-02
+
+**Cilium Release:** 1.16
+
+**Authors:** Casey Callendrello <cdc@isovalent.com>
+
+## Summary
+
+Allow users to create network policies that do not implicitly set endpoints in to a default-deny mode. 
+
+## Motivation
+
+Consider a large, multitenant cluster. The cluster administrator is distinct from
+the workload owners, and the workloads are diverse and pre-existing.
+
+The administrator may wish to create a Cilium Clusterwide Network Policy (CCNP)
+that affects most or all existing workloads. For example, they may wish to
+- Globally grant access to a security scanner
+- Ensure that workloads can access kube-dns
+- Proxy all DNS requests for monitoring purposes
+- Deny access to sensitive IPs, such as cloud provider metadata APIs
+
+All of these can already be accomplished by the existing policy language. However, there
+is an operational risk here: any of these broad-based policies may be the first policy
+that applies to a given endpoint.
+
+By default, all Cilium endpoints have unrestricted network access. If at least one
+network policy applies to an endpoint, then all traffic is denied except that
+explicitly allowed by policy.
+
+This presents a challenge for the cluster administrator. They cannot assume that
+existing, heterogeneous worloads already have network policies provisioned. So, by
+creating an administrative network policy -- even one that only allows traffic --
+they may select endpoints that **previously were in the allow-all mode**. Thus, 
+they may unintentionally block legitimate traffic for an existing workload, causing an outage.
+
+## Goals
+
+* Ability to safely roll-out broad-based network policies, without the risk
+  of disrupting existing traffic.
+
+## Non-Goals
+
+* An iterative, iptables-style firewall rules engine
+* Arbitrary, multi-level rule priority
+
+## Proposal
+
+### Overview
+
+We propose a new field in the policy spec, `EnableDefaultDeny`. This field specifies
+whether or not endpoints selected by a policy should be placed in to default-deny mode
+_on behalf of this policy_.
+
+This field is optional. The default value per-direction is `true`. This matches the existing behavior,
+wherein any policy automatically sets the subject endpoints' mode to default-deny. 
+
+
+Sample YAML & go types:
+```yaml
+enableDefaultDeny:
+  egress: false
+  ingress: true
+```
+```go
+type PolicySpec struct {
+  // ...
+  EnableDefaultDeny DefaultDenyConfig `json:"enableDefaultDeny,omitempty"`
+}
+
+type DefaultDenyConfig struct {
+  Ingress *bool `json:"ingress,omitempty"`
+  Egress  *bool `json:"egress,omitempty"`
+}
+```
+
+Note that this is unrelated to the `IngressDeny` and `EgressDeny` policy expressions. Those dictate the action for selected peer traffic, whereas the `EnableDefaultDeny` specifies what should happen with unselected peer traffic.
+
+### Per-direction mode
+
+Endpoints have a default mode *for each direction*. That is to say, if an endpoint only has policies with ingress peer selectors, then ingress is default-deny, and egress is default-allow.
+
+The rest of this document, for shorthand, assumes all policies operate in the same direction. The logic described is identical but independent for each direction.
+
+### Precedence
+
+When an endpoint is selected by multiple policies, default-deny takes precedence over default-allow. That is to say, if a new policy with `enableDefaultDeny: false` selects an endpoint with existing policies, no new traffic shall be allowed except that selected by the new policy.
+
+For each traffic drection, an endpoint can be in the following states:
+
+* No policies select this endpoint. Default allow.
+* Only `enableDefaultDeny true` policies select. Default deny; only selected peers allowed.
+* `enableDefaultDeny true` and `false` policies select. Default deny; only selected peers allowed.
+* Only `enableDefaultDeny false` policies select. Default allow; only traffic explicitly blocked by an `IngressDeny` / `EgressDeny` selector is blocked.
+
+
+## Examples
+
+### Allow a security scanner full access to the cluster
+
+```yaml
+kind: CiliumClusterwideNetworkPolicy
+spec:
+
+  enableDefaultDeny:
+    egress: false
+
+  endpointSelector: {}
+  ingress:
+  - fromEndpoints:
+    - matchLabels:
+        "k8s:io.kubernetes.pod.namespace": security-team
+```
+
+### Observe all DNS traffic
+
+```yaml
+kind: CiliumClusterwideNetworkPolicy
+metadata:
+  name: all-ur-dns-are-belong-to-us
+spec:
+
+  enableDefaultDeny:
+    egress: false
+  
+  endpointSelector: {}
+  egress:
+    -  toPorts:
+        - ports:
+          - port: "53"
+            protocol: ANY
+          rules:
+            dns:
+              - matchPattern: "*"
+```
+
+## Impacts / Key Questions
+
+### Impact: More permissive policy engine
+
+Making the policy engine more permissive is always a sensitive change. It is important that user intent is respected, and that no traffic is inadvertently allowed. It is critical that existing policies are not made more permissive.
+
+This proposal attempts to respect the user's intent by making the permissive behavior explicitly opt-in and a part of the existing policy API. It does not create any opaque configuration flags or mystery annotations.
+
+As this is a new field, this proposal does not affect any existing policies. Workloads with network policies will not see a change in allowed / denied traffic. Likewise, workloads without any policies will be unaffected.
+
+This proposal does not change the precedence of Deny vs. Allow *peer selectors* -- as always, an `IngressDeny` selector always takes precedence over an `Ingress` selector, independent of the default action.
+
+### Alternative: Agent Configuration
+
+The policy engine already has a tri-state enablement mode:
+
+- Disabled
+- Default
+- Always
+
+The default is as described above -- endpoints without policy are permissive and the creation of policy changes that endpoint to restricted. Administrators may choose to completely disable network policy (in which all traffic is allowed, always), or set all endpoints to default-deny regardless of policy.
+
+This value cannot be changed on a per-endpoint basis; only per-Agent.
+
+### Option 1: Agent-level configuration
+
+We could add a fourth policy enablement mode that changes the defaulting logic for the Cilium agent. When set to this mode, CCNPs would never cause the subject endpoints to be in the default-deny mode. Other policies, however, still would.  
+
+#### Pros
+
+* Does not require any API changes
+
+#### Cons
+
+* Cluster specific. A workload that is migrated between clusters may stop working. This breaks the principle that the existing API objects should describe workloads completely.
+* Diverges the logic between CNP and CCNP. An administrator may be surprised that an otherwise-identical CNP and CCNP would behave completely differently.
+* Opaque. The same CCNP, inadvertently applied to a cluster without this option, will instantly drop all traffic. This violates the principle of least surprise. 
+* Inflexible. One could imagine a scenario in which all new namespaces are required to have a zero-trust-style network policy, but existing namepaces are "grandfathered". The administrator may wish to create a CCNP that emulates a default-deny mode, *but only for namespaces without a "grandfathered" label*. Given the existing policy language, this would be easy. If, however, the default action is not configurable on a per-CCNP basis, this is impossible.
+
+### Option 2: Per-namespace annotation
+
+#### Pros
+
+* Does not require any API changes
+* Non-opaque; workload owners could view enforcement status
+
+#### Cons
+
+* Incomplete. We have a rich policy API type; it should be the complete description of a workload's policy state.
+* Surprising. The same CCNP may behave differently on different clusters, based on an unrelated and opaque namespace annotation.
+* Inconsistent. Network policies operate on labels; annotations are not supposed to affect network policy
+* Imprecise. This only allows for configuration on a namespace basis; where the policy engine can select on a much finer grained basis.
+
+## Future Milestones
+
+### Kubernetes NetworkPolicy equivalent
+
+One could imagine an annotation on a Kubernetes NetworkPolicy that had a similar effect. 
+
+### Hard-coded network policies
+
+Other end-users would like the ability to configure super-CCNPs that cannot be changed by cluster administrators. These would be provided either by ConfigMap or similar mechanism.


cilium/CFP-30572-non-default-deny-policies.md
```diff
@@ -0,0 +1,175 @@
...
+### Overview
+
+We propose a new field in the policy spec, `DefaultAction`. This field specifies
+whether or not an implicit default-deny rule should be created _on behalf of this policy_.
+
+This field is optional. The default value is `Deny`. This matches the existing behavior, wherein any network policy implicitly blocks all other traffic.
+
+The value `Pass` indicates that this policy should not create a default-deny rule.
+
+Note that this is unrelated to the `IngressDeny` and `EgressDeny` policy expressions. Those dictate the action for *matching* traffic, whereas the `DefaultAction` specifies what should happen with *non-matching* traffic.
```
joestringer, line 60: I found myself falling into a bit of a trap with the wording around `non-matching`. In the newer ANP terminology they clearly define the _subject_ of the policy and separately the rules that apply to that subject. In this case, the "non-matching traffic" is all traffic that is part of the direction in the rules but not explicitly described in the rules, but also *still part of the subject*. Ie if you use an endpointSelector to apply policy for `app: foo` within namespace `foo`, then this has no impact on the policy for `app: bar` in namespace `bar`. The non-matching traffic is the traffic to/from the subject which is not explicitly described in the policy.

I think that this existing description is probably clear enough, but we might consider mentioning the subject:
```suggestion
Note that this is unrelated to the `IngressDeny` and `EgressDeny` policy expressions. Those dictate the action for *matching* traffic for the endpoint subject to the policy, whereas the `DefaultAction` specifies what should happen with *non-matching* traffic for the subject endpoint.
```

(Not entirely sure if this helps but wording is always a bit tough. Feel free to adopt, shuffle around this suggestion or reject it altogether)

cilium/CFP-30572-non-default-deny-policies.md
```diff
@@ -0,0 +1,175 @@
...
+## Non-Goals
+
+* An iterative, iptables-style firewall rules engine
+* Arbitrary, multi-level rule priority
+
+## Proposal
+
+### Overview
+
+We propose a new field in the policy spec, `DefaultAction`. This field specifies
```
joestringer, line 53: My first impression was that expressing this feature as a `DefaultAction` creates some dissonance with the current language we use in CNP/CCNP. Specifically, I think of the current policy expressions in the CNP/CCNP as _user intent_ and not _orders for a thing to do something_. We tend to avoid using any sort of verbs or actions. The nice property of this is that we end up with a bunch of composable statements that have no reference to a specific time or place. When I read a policy, I read passive statements like:
- "when `app: foo` sends traffic to endpoints with `app: bar` on port 80, it is allowed"
- "when `app: foo` sends HTTP requests to `/public` on port 80 to `app: bar`, it is allowed".

Then, if we need to evaluate the way that each of these policies applies, we can tell individually by looking at one rule or another that the specific target traffic should be allowed. There's no sense of whether rules in one policy apply before or after rules in another policy. I think ultimately this is how we end up being able to process these expressions of user intent efficiently down into BPF for fast lookups, because there is no inherent ordering in the policy language. Of course, in the case of deny policies there is a slight deviation from this, and in the implementation details of L7 policies or so on we do end up needing to evaluate the semantics of the rules with respect to each other, but we're talking about a relatively small set of guidelines around how sets of two rules can be combined. It's not a general purpose multi-level rule priority system.

With that in mind I thought it's worth at least throwing around some ideas that could allow this feature to be expressed as a property of the policy itself without introducing verb language:

- `mode: PassByDefault/DenyByDefault`: Models the policy as having a mode of operation which is more like a property and less like a specific action at a time and place.
- `defaultAllow: true`: This is a bit specific rather than having a mode string, but has the benefit of being false by default without specific code to handle it, and I think that it clearly composes with the individual rules in the policy in a similar way to the current proposal.
- `nonMatchingTraffic: Pass/Drop`: A bit more explicit, this spells out what "default" means in the previous proposals.
- `zeroTrust: true`: assume a zero trust posture, ie do not allow anything unless it is explicitly allowed. Cons: potential for this to have overloaded meaning given the marketing origins of the expression "zero trust". This is also pretty opinionated, it implies anyone using this requested policy feature is not following zero trust. Then again, is that so far from reality....? If every user of the cluster declared their policy intent ahead of time, this feature would not be necessary at all I think?

cilium/CFP-30572-non-default-deny-policies.md
```diff
@@ -0,0 +1,175 @@
...
+whether or not an implicit default-deny rule should be created _on behalf of this policy_.
+
+This field is optional. The default value is `Deny`. This matches the existing behavior, wherein any network policy implicitly blocks all other traffic.
+
+The value `Pass` indicates that this policy should not create a default-deny rule.
```
joestringer, line 58: One thing I like about the idea of a string for this property is the possibility for us to reuse this same field in future for a per-policy / per-namespace audit mode feature. May be worth considering whether we want to leave that option open here with a new enum value for the field or whether we would expect such a selective audit mode feature to be implemented separately.

joestringer (commented):
Overall the feature here makes sense, my review below is mostly quibbling about naming/semantics and ensuring the wording is precise in order to minimize confusion as much as possible.

I don't anticipate many concerns in general with the proposal here, happy to approve once we're on the same page for semantics. Obviously when it comes to implementation details we'll need to think through how the docs can express this feature in clear terms, and also the sorts of test cases we will have to validate the feature. The four precedence examples already provide some hints as to where some of the testing effort should be spent.

cilium/CFP-30572-non-default-deny-policies.md
joestringer: I think there are elements of the same concern in @learnitall's feedback over [here](https://github.com/cilium/cilium/pull/30572#pullrequestreview-1867389946), where a "defaultAction" could be interpreted as a "slot" in the overall architecture which is populated by this policy, rather than a property of this specific policy which then implies that it doesn't impact any other policy in any way.

joestringer: Whichever pattern we choose, I think it's also worth reading through the UX of this feature / field and express it in human terms the way that @learnitall did at the bottom of his linked comment above. If it's natural to go from the field name to a human-friendly description of the behaviour then that will serve well for documentation and will help users to read and understand the policy when they see this field.

squeed: I agree that the naming is awkward -- hence this CFP :-). We're trying to find a way to express - arguably - the most complicated part of k8s / Cilium network policy. Namely, that creating the first policy for a subject is really doing *two* things -- allowing X, and blocking ^X. I don't have any good answers :-).

Coming at this backwards, from the perspective of the developer, `nonMatchingTraffic: Pass/Drop` seems the closest fit. "Pass" as a term already has a meaning in the context of firewall rules, and it pretty closely matches what this PR does. The only awkwardness is that it - from a certain reading - could imply that all other traffic would be dropped, even if allowed by other network policies. Hm.

cilium/CFP-30572-non-default-deny-policies.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Per-direction state
+
+Endpoints have a default rule *for each direction*. That is to say, if an endpoint only has rules that select ingress traffic, then ingress is default-deny, and egress is default-allow.
+
+The rest of this document, for shorthand, assumes all rules operate in the same direction. The logic described is identical but independent for each direction.
+
+### Precedence
+
+When an endpoint is selected by multiple rules, `Deny` shall take precedence over `Pass`. That is to say, if a new `Pass`-type rule selects an endpoint with existing policies, no new traffic shall be allowed except that selected by the new rule.
```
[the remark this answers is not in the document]
squeed: Oh, good point. We use "Rule" and "Policy" somewhat interchangeably in the codebase, since a `CiliumNetworkPolicy.Spec` is just a `struct Rule`. And then `Rule.Ingress []IngressRule` further complicates matters.

This should be a policy-level field. I'll update the doc.

cilium/CFP-30572-non-default-deny-policies.md
```diff
@@ -0,0 +1,175 @@
...
+
+Making the policy engine more permissive is always a sensitive change. It is important that user intent is respected, and that no traffic is inadvertently allowed. It is critical that existing policies are not made more permissive.
+
+This proposal attempts to respect the user's intent by making the permissive behavior explicitly opt-in and a part of the existing policy API. It does not create any opaque configuration flags or mystery annotations.
+
+As this is a new field, this proposal does not affect any existing policies. Workloads with network policies will not see a change in allowed / denied traffic. Likewise, workloads without any policies will be unaffected.
+
+This proposal does not change the precedence of Deny vs. Allow *peer selectors* -- as always, an `IngressDeny` rule always takes precedence over an `Ingress` rule, independent of the default action.
+
+### Alternative: Agent Configuration
```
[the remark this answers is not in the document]
squeed: Agreed, once I sat down to think about agent config flags (which had been proposed somewhere), I realized I preferred API types over config.

cilium/CFP-30572-non-default-deny-policies.md
squeed: So [upstream k8s](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.29/#networkpolicyspec-v1-networking-k8s-io) has the `policyTypes` array, which solves a similar problem as this one. That said, I don't like the connotations implied by "Type", which seems wrong for this kind of field.

Perhaps
```yaml
spec:
  ...
  nonMatchingTraffic:
    ingress: Drop
    egress: Pass
```

is the right shape. We can compute the defaults neatly, based on the presence or absence of rules in each direction.

joestringer: Incidentally the one thing I find a little ambiguous about the `nonMatchingTraffic` proposal is exactly what the "non-matching" aspect relates to. If you just look at the hierarchy proposed above with `spec: nonMatchingTraffic: ...`, the field is nested directly under the specification. As such it's not immediately clear the relationshp between `nonMatchingTraffic` and the `endpointSelector` since they're sibling fields. I think we all mean that `nonMatchingTraffic` still must match the `endpointSelector`. A couple of thoughts further on that would be either to embed that understanding directly into the naming ("subjectDefaultPosture: Pass/Drop"?), to move this field to become a property of the `EndpointSelector`, or to express this feature as a rule under the `ingress`/`egress` stanzas (not exactly great, because then it's just one more rule in a list of rules and it's much easier to lose this highly important policy-wide semantic in a big list of rules).

I actually think that the use of `default` wording provides just enough abstraction to cover this ambiguity. Default relies on the user understanding the notion that "a policy creates a default deny posture", but once they have that knowledge they can quickly jump from a field named something like `PassByDefault` to an intuitive understanding of how that changes the policy behaviour.

Since these options are probably fractal, maybe to put forth what seems to me one of the better options mitigating most of the downsides discussed above, how about:

```
spec:
  endpointSelector:
    matchLabels:
      org: empire
      class: deathstar
    defaultPosture: Pass
```

joestringer:
```
  nonMatchingTraffic:
    ingress: Drop
    egress: Pass
```

Just to expand a bit more specifically on explicitly defining the `ingress` / `egress` aspects here. It is more explicit, which can make it clearer to reason about. To explore the semantics a bit we can probably iterate through various permutations of policies with the fields specified or unspecified and the implied meaning.

## Policy behaviour today

```
spec:
  endpointSelector:
    matchLabels:
      org: empire
      class: deathstar
  ingress:
    ...
```

Ingress rules are specified. The subject "deathstar" endpoints are put into default deny at ingress. Some specific traffic is also allowed, as specified in the rules. Other policies may further extend the allowed/denied traffic at ingress. This policy does not impact egress.

---

```
spec:
  endpointSelector:
    matchLabels:
      org: empire
      class: deathstar
  ingressDeny:
    ...
```

IngressDeny rules are specified. The subject "deathstar" endpoints are put into default deny at ingress. Some specific traffic is also denied, as specified in the rules. Other policies may further extend the allowed/denied traffic at ingress. No other policy may allow traffic that is specified in the deny rules here. This policy does not impact egress.

---

```
spec:
  endpointSelector:
    matchLabels:
      org: empire
      class: deathstar
  ingress:
  - {}
```

Ingress rules are specified, and there is an ingress rule which specifies no aspects of traffic to match. Due to the existance of the `ingress` stanza, the endpoint goes into default deny, and furthermore since the only rule does not match any traffic, this implicitly denies all traffic. Other policies may further extend the allowed/denied traffic at ingress. This policy does not impact egress.

## Split defaults proposal

I'll use the `nonMatchingTraffic` expression of this CFP idea to reason about the ingress/egress aspect below, but I think this will equally apply if we choose to express the feature in other terms.

```
spec:
  endpointSelector:
    matchLabels:
      org: empire
      class: deathstar
  ingress:
    ...
```

Like today, we must not change the semantics of this. Due to the lack of `nonMatchingTraffic` field and the fact that `ingress` is specified, the "deathstar" endpoints are put into default deny at ingress. This policy does not impact egress. Implicitly this is treated as though the above policy also has the following `nonMatchingTraffic` specification:

```
  nonMatchingTraffic:
    ingress: Drop
    egress: Pass
```

We implicitly construct this nonMatchingTraffic behaviour based on the existence of the `ingress` field and the lack of existance of the `egress` field.

---

```
spec:
  endpointSelector:
    matchLabels:
      org: empire
      class: deathstar
  nonMatchingTraffic:
    ingress: Pass
  ingress:
    ...
```

Now that the `nonMatchingTraffic` field explicitly specifies the `ingress` aspect, the `ingress` field no longer implicitly creates a default deny posture. Of course given that there is no explicit mention of egress, this policy has no impact on egress.

---

```
spec:
  endpointSelector:
    matchLabels:
      org: empire
      class: deathstar
  nonMatchingTraffic:
    egress: Deny
  ingress:
    ...
```

There is no `nonMatchingTraffic` specification on ingress, so given that there are ingress rules the default posture on ingress is deny. The `nonMatchingTraffic` field specifies the posture for egress though, so now egress is set to a default deny (which may be overridden by other policies).

---

```
spec:
  endpointSelector:
    matchLabels:
      org: empire
      class: deathstar
  nonMatchingTraffic:
    ingress: Pass
  ingress:
  - {}
```

Although there is a rule in the `ingress` stanza, that rule doesn't select any specific traffic, because the entire rule is empty. That implies that all traffic on ingress to this endpoint is non-matching and hence subject to the `nonMatchingTraffic` behaviour.

What's interesting about this case is that if we dropped/ignored the `nonMatchingTraffic` field, this would be the typical expression of an [ingress default deny policy](https://docs.cilium.io/en/stable/security/policy/language/#ingress-egress-default-deny). 

Strictly speaking I think this is actually consistent and I think that if someone was to read through such a policy, given the clear `nonMatchingTraffic` field I think the behaviour could be intuitively understood.

---

Based on the examples above, there will still need to be default behaviour for the cases where the user does not explicitly specify the new field. However, the benefit would be that it's possible to express the variations on default behaviour explicitly as fields directly in the policy. We could even provide tooling for existing users to convert their underspecified policies directly into explicitly specified policies, ie take existing policies and populate the new field to make the policies clearer. I don't feel strongly about this either way but I think that the explicit split of the directions here is a reasonable proposal.

joestringer: Recording back to this thread: @nathanjsweet , @learnitall , @squeed and I discussed the "split defaults" aspect of this for splitting ingress/egress specifications for the behaviour and generally agreed that it's better to explicitly split this feature by direction.

We discussed through some of the variations on naming that I brought up at the start of this thread and discussed nuances on semantics. We didn't conclusively come to an agreement on that but I think we're more in sync about the concerns so hopefully we can drive agreement soon.

cilium/CFP-30572-non-default-deny-policies.md
joestringer: From discussion out-of-band with @nathanjsweet , @learnitall , @squeed we acknowledged the potential overlap for describing audit mode in this way, but also identified that there are nuances in the way that audit mode applies which raise some uncertainty about whether embedding audit mode into individual policies is a good idea. Specifically, audit mode is typically applying on a per-endpoint basis rather than selecting specific traffic to apply to. While it's plausible for us to express this concept in the policy in this area, we don't want to tie the overall proposal here to the possibility of audit mode, so it makes sense to put this idea back on the shelf for now. If and when we have stronger motivation to address a more dynamic audit mode feature for individual endpoints/namespaces or even policies, we can revisit the solution space specifically for that feature.

cilium/CFP-30572-non-default-deny-policies.md
squeed: I changed the wording a bit, using the terms "subject" and "peer". This is at least vaguely consistent with the k8s terminology.

cilium/CFP-30572-non-default-deny-policies.md
```diff
@@ -0,0 +1,175 @@
...
+### Precedence
+
+When an endpoint is selected by multiple rules, `Deny` shall take precedence over `Pass`. That is to say, if a new `Pass`-type rule selects an endpoint with existing policies, no new traffic shall be allowed except that selected by the new rule.
+
+For each traffic drection, an endpoint can be in the following states:
+
+* No policy selected. All traffic is allowed
+* `Deny`-type rule selected. Default deny; only selected traffic allowed
+* Both `Deny`-type and `Pass`-type rules. Default deny; only selected traffic allowed
+* Only `Pass`-type rules. Default allow; only traffic explicitly blocked by an `IngressDeny` / `EgressDeny` selector is blocked
```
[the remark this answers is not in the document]
squeed: Right, the documentation is a bit confusing as well, since it mentions "deny policies". Deny is not a property of a policy, it's a property of a rule. The go types are, loosely,
```go
type PolicySpec struct {
  Ingress []IngressRule
  IngressDeny []IngressRule
  //...
}
```

I'll give the CFP a once-over to make it consistent, but from the perspective of implementation, a Policy consists of a set of Rules, and Rules have a verdict. Kind of.

cilium/CFP-30572-non-default-deny-policies.md
joestringer: The "vetoing" property is the critical semantic here, I agree. If I visualize the network policy engine as having a map of rules and then a "slot" at the end of processing which decides whether or not to drop all traffic that didn't already receive a verdict, then I would say that today each network policy populates that slot. This field is supposed to decide whether or not this specific policy will populate that slot. As long as any one policy populates that slot, there will be a default deny. 

I don't love that omitting the field is the same as setting it to true, but I don't see a great way to reverse the expression of the feature and retain the veto property. Certainly not `defaultAllow`, because that would imply that this policy may override other policies. Similarly for `defaultPass`. Although I had some fondness for the `PassByDefault`, `DenyByDefault` proposals I made above, I do not think that this feature can be expressed as a "mode" in the policy. There is a subtlety here in the way that each individual policy interacts with one another and with the core policy engine to potentially overlap with static agent flags. I guess that's the price of fixing API mistakes from years ago :)

learnitall: I think that's a good call @squeed. It's very clear and there probably isn't a great way to express this idea without using a verb. I don't know if there is a great way to make this field false by default either, since reversing the logic would add some bloat:

```yaml
spec:
  doesNotCreateDefaultDeny:
    ingress: false
    egress: false
```

The last bit of feedback I have here is that there might be a synonym for "creates" that sounds less verb-y but isn't necessarily in the form of third person singular:

* Establish
* Enact
* Initiate
* Trigger (my favorite)

joestringer: I agree `doesNot...` is too clumsy.

Maybe one more for the list, `EnableDefaultDeny`. That would probably be my pick out of all of the verbs above including `Create`.

Initiate/Trigger seem too temporal to me, they suggest an even more timing-sensitive aspect to the semantic which I don't think reflects the reality of the feature. Enact is a bit more nuanced, perhaps less accessible for non-English speakers. Establish is fine. Create is OK, it fits CRUD terminology better, not sure if that's a good thing or bad thing since we don't have the rest of the CRUD verbs.

joestringer: Maybe one other subtle nuance to throw into the ring, replace `Default` with `Implicit`. This wording comes directly from one of the users that was requesting the feature. So something like `CreateImplicitDeny` / `EnableImplicitDeny`.

cilium/CFP-30572-non-default-deny-policies.md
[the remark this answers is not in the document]
joestringer: Cool yep. We can always improve these things over time. If we're at least on the same page for how this CFP treats these words and that we should try to use this terminology more going forward then things should become clearer :+1:

cilium/CFP-30572-non-default-deny-policies.md
squeed: I like "enable", seems easy enough. I do prefer "Default" over "Implicit", since firewalls have had the notion of a "default action" since "time immemorial".

I'm going to update the PR with `enableDefaultDeny:...` and then this should be ready to go.

nathanjsweet: > I like "enable", seems easy enough. I do prefer "Default" over "Implicit", since firewalls have had the notion of a "default action" since "time immemorial".

So is the default value "true"?

squeed: The default value is true for each direction with rules. Same as before -- creating a policy with just Ingress rules blocks all other ingress traffic, but leaves egress untouched.

nathanjsweet: I agree with the overall semantics, but my preference is that default boolean values are always false. I know there are exceptions all over the place in the Cilium code base, but I would prefer `DefaultAllow` or `enableDefaultAllow` over `enableDefaultDeny` so that the default boolean value can be "false" as I think most API users expect it. However, I'm not going to hold up this CFP over syntax.

joestringer: @nathanjsweet I sympathize with the argument around boolean values, but specifically if we look at `enableDefaultAllow`, it implies that this policy creates an "allow" rule by default, which does not compose as clearly with other policies. If you have three policies, two with `enableDefaultAllow: true` and one without, the behaviour may surprise users. The policy without the field will implicitly have `enableDefaultAllow: false`, and this results in the actual behaviour being to enable default deny. The (omitted) `false` boolean would be overriding the explicit positive from the other policies.

cilium/CFP-30572-non-default-deny-policies.md
```diff
@@ -0,0 +1,202 @@
...
+## Summary
+
+Allow users to create network policies that do not implicitly set endpoints in to a default-deny mode. 
+
+## Motivation
+
+Consider a large, multitenant cluster. The cluster administrator is distinct from
+the workload owners, and the workloads are diverse and pre-existing.
+
+The administrator may wish to create a Cilium Clusterwide Network Policy (CCNP)
```
christarazi, line 20: Very nit: 
```suggestion
The administrator may wish to create a Clusterwide Cilium Network Policy (CCNP)
```

cilium/CFP-30572-non-default-deny-policies.md
```diff
@@ -0,0 +1,202 @@
...
+  EnableDefaultDeny DefaultDenyConfig `json:"enableDefaultDeny,omitempty"`
+}
+
+type DefaultDenyConfig struct {
+  Ingress *bool `json:"ingress,omitempty"`
+  Egress  *bool `json:"egress,omitempty"`
+}
+```
+
+Note that this is unrelated to the `IngressDeny` and `EgressDeny` policy expressions. Those dictate the action for selected peer traffic, whereas the `CreateDefaultDeny` specifies what should happen with unselected peer traffic.
```
christarazi, line 81:
```suggestion
Note that this is unrelated to the `IngressDeny` and `EgressDeny` policy expressions. Those dictate the action for selected peer traffic, whereas the `EnableDefaultDeny` specifies what should happen with unselected peer traffic.
```

christarazi (approved):
Two very minor nits, LGTM

cilium/CFP-30572-non-default-deny-policies.md
joestringer: Cilium comes first in the acronym in the vast majority of matches in the codebase, I think the wording is correct as-is.

```
$ git grep ClusterwideCi | wc -l
2
$ git grep CiliumClusterwide | wc -l
500
```

joestringer:
@squeed I believe that as a committer, you should be able to merge this PR. I'll leave it up to you to make any last minute wording/typo changes and do the honors.

cilium/CFP-30572-non-default-deny-policies.md
christarazi: Oh, you're right, I didn't check before posting and I've always thought it was Clusterwide first and then the CNP part after, but I guess I had the reverse in my head all this time :)

PR #17, opened 2024-02-12
Select ToFQDN prefixes via a special label

squeed:
This proposes a new label source: `dns`. This would then tag prefixes with the ToFQDN selector(s) that match this prefix. E.g. `192.0.2.42/32 -> dns:*.cilium.io`

Post-merge follow-up items:
- [ ] Consider a flag limiting the number of `fqdn:` identities in the local allocator
- [ ] Add alerts for excessive local identities
- [ ] Chunk ipcache updates.

### DIFF
--- a/cilium/CFP-28427-fqdn-selectors.md
+++ b/cilium/CFP-28427-fqdn-selectors.md
@@ -1,0 +1,404 @@
+# CFP-28427: ToFQDN policy performance improvements
+
+**SIG: SIG-Policy**
+
+**Begin Design Discussion:** 2024-02-12
+
+**Cilium Release:** 1.16
+
+**Authors:** Casey Callendrello <cdc@isovalent.com>, Sebastian Wicki <sebastian@isovalent.com>
+
+## Summary
+
+Improve FQDN-based policy performance by changing how identities and labels are managed. Rather than propagating FQDN -> IP mappings to selectors, label IPs by the FQDN selector(s) that return them.
+
+e.g. `18.192.231.252 -> (dns:*.cilium.io)`
+
+## Motivation
+
+ToFQDN scaling limitations are a source of pain for many, many Cilium end users. When a domain name returns a large and constantly varying set of IPs (e.g. S3), this causes significant identity churn and increased memory usage.
+
+At the time of writing, Cilium allocates a unique security identity for each IP address, even if all of those addresses are selected by a single selector. All these identities then need to be pushed in to every relevant PolicyMap as well as Envoy -- all before returning the DNS response to the waiting endpoint.
+
+## Goals
+
+- Reduce resource consumption by large ToFQDN policies
+- Improve response latency for proxied DNS requests
+
+----
+
+
+## Background
+
+First, let us understand the existing components at work for ToFQDN policies.
+
+### Lexicon
+
+**Label:** A key-value pair, optionally tagged with a source. For example, `k8s:app:database` is equivalent to the Pod label `app=database`
+**Identity:** A unique set of labels. Every identity has a numeric identifier.
+**Selector:** A policy entity that selects identities. 
+**Label Selector:** A selector that references by label key and/or value.
+**CIDR Selector:** A selector that references by CIDR. CIDR selectors are converted to a special form of label selectors.
+**ToFQDN Selector:** A selector that dynamically selects based on the result of intercepted DNS requests and responses.
+
+### CIDR labels
+
+The BPF policy engine relies on identities -- it does not understand CIDRs or IPs directly. So, special CIDR labels are used, and special _local identities_ are allocated on-demand to represent these external IP blocks.
+
+The IPCache aggregates multiple sources of labels for external CIDRs. As necessary, it will generate the complete set of labels and allocate a numerical identity.
+
+Consider a ToCIDR block that selects `192.0.2.0/24`. The policy subsystem will upsert the prefix-label mapping `192.0.2.0/24 (cidr:192.0.2.0/24)`. The IPCache will allocate a local identity (e.g. 16777216) for this set labels. It will also convert the ToCIDR selector to a label selector, selecting the identical CIDR label. The SelectorCache will then select this numeric identity when generating the policy maps.
+
+If another ToCIDR block selects `192.0.2.128/25`, a similar preifx-label mapping will be created `192.0.2.128/25 -> (cidr:192.0.2.128/25)`. Likewise, a new identity will be allocated for this unique set of labels. The SelectorCache will be updated, and the label selector `cidr:192.0.2.0/24` will **now match two identities**.
+
+### Components & Flow
+
+```mermaid
+graph TD;
+sc[SelectorCache]
+pol[Network Policy]
+bpf[/BPF\]
+ipc[IPCache]
+nm[NameManager]
+dnsp{{DNS packet}}
+
+dnsp-- UpdateGeneratedDNS() -->nm
+sc-- IdentitySelectionUpdated() -->pol
+nm-- UpsertPrefix() -->ipc
+ipc-- UpdateIdentities() -->sc
+nm-- UpdateFQDNSelector() -->sc
+pol-- SyncPolicyMap() -->bpf
+```
+
+#### The SelectorCache
+
+The SelectorCache maintains a list of all "active" selectors for policies. If, for example, there is a network policy that grants access to label `team: database`, this selector will "live" in the SelectorCache. The policy will ask the SelectorCache for all identities that match this selector. Furthermore, the policy will "register" for incremental updates to this selector.
+
+#### NameManager
+
+The NameManager maintains two key lists:
+1. The mapping from DNS names to IPs
+2. The set of active FQDN selectors
+
+As DNS responses come in, the NameManager determines if the name is selected by any ToFQDN policies. If so, it inserts the IP in to the ipcache (if needed) and updates the relevant ToFQDN selectors.
+
+#### IPCache
+
+The IPCache stores the mapping from prefix to labels. It also allocates identities for labels, and manages the mapping from prefix -> numeric identity.
+
+### Incremental updates
+
+   1. An identity is allocated for the IP:
+      1. `ipc.UpsertPrefixes()` is called, associating the complete set of CIDR labels with this IP address.
+      2. The IPCache determines the IP's full set of labels, allocates an identity, and informs the SelectorCache of this new IP
+   2. The selector is updated:
+      1. The NameManager calls `sc.UpdateFQDNSelector()`, updating the set of selected IPs for this given FQDNSelector
+      2. The SelectorCache determines the complete set of identities selected
+      3. Any differences are pushed to network policies
+   3. Endpoints are updated
+      1. Any pending incremental updates are batched up and applied to the endpoints' BPF PolicyMap. If needed, Envoy is also updated
+ 
+Only at this point can the DNS packet be returned to the waiting endpoint.
+
+## Performance bottlenecks
+
+### Many identities
+
+Some DNS names (e.g S3) will return many different IP addresses. S3 even has a TTL of 5 seconds, so the set of IPs will quickly be very broad. This causes a large number of identities to be allocated -- one per IP. These identities then need to plumbed through to all endpoints.
+
+### Envoy updates
+
+Adding a new indentity to a selector requires an xDS roundrip to Envoy. This has a high tail latency, and in practice can take up to 100ms. This may have to happen multiple times for a single proxied FQDN request.
+
+----
+
+## Proposal
+
+This CFP proposes 3 key changes. Together, they preserve correctness while radically restructuring how prefixes are selected.
+
+1. Labels on prefixes propagate downwards. That is to say, if prefix `192.0.0.0/22` has label `foo:bar`, then prefix `192.0.2.0/24` also has label `foo:bar`, unless overridden. The IPCache aggregates labels accordingly.
+2. Only selected CIDR labels are inserted in to the ipcache. If a selector references CIDR `192.0.2.0/24`, only the prefix -> label mapping `192.0.2.0/24 -> cidr:192.0.2.0/24` is added to the IPCache.
+3. When the NameManager learns about a new relevant `name -> IP` pair, it upserts the label `IP -> (dns:selector)` It **does not** create any CIDR labels.
+
+
+### 1: CIDR Label down-propagation
+
+Currently, prefixes in the ipcache are independent. Instead, labels should flow downwards from prefixes. For example, consider an IPCache with two prefixes: 
+
+```
+192.0.2.0/24 -> reserved:world, k8s:foo:bar, cidr:192.0.2.0/24
+192.0.2.2/32 -> reserved:world, cidr:192.0.2.2/32, reserved:kube-apiserver
+```
+
+Then, the complete set of labels for `192.0.2.2/32` would be `reserved:world, k8s:foo:bar, cidr:192.0.2.2/32, reserved:kube-apiserver`. As a special case, shorter CIDR labels can be omitted from descendant prefixes, as the selector logic will match longer prefixes correctly.
+
+This is a cornerstone for implementing another feature, non-k8s label sources. It is also required for correctness within the context of this proposal, especially when ToFQDN and CIDR selectors overlap.
+
+#### Lifecycle
+
+With label down-propagation, it is important that updates are propagated accordingly. When a prefix P is updated or deleted, then updates must be also triggered for all prefixes contained in P.
+
+This update may be calculated if we use a bitwise LPM trie. Alternatively, scanning the entire ipcache to determine affected prefixes may be more efficient.
+
+
+### 2: Only generate terminal CIDR labels
+
+The label selector logic is already cidr-aware, and does not require cidr label expansion for shorter-prefixes to correctly match descendant prefixes.
+
+So, if selector A selects `cidr:192.168.0.0/24`, there will be a single identity, `id1 -> (cidr:192.168.0.0/24)`. If selector B selects `cidr:192.168.0.0/25`, then there will be two identities: `id1-> (cidr:192.168.0.0/24)` and `id2 -> (cidr:192.168.0.0/25)`. The label matching logic ensures that selector A will now select _id1_ and _id2_, and selector B will select just _id2_.
+
+Thus, when generating the set of labels for a prefix, we should discard any intermediate CIDR labels, keeping only the longest prefix.
+
+### 3: FQDN selector labels
+
+Currently, FQDN selectors (as stored in the SelectorCache) understand the set of IPs that they select. They determine the relevant set of identities by generating the equivalent `cidr:x.x.x.x/32` labels and looking for matching identities.
+
+Presently, when a new `(name: IP)` mapping is learned, the NameManager inserts that IP in the IPCache and updates the selectors' set of desired labels to include that IP's CIDR label.
+
+Instead, the NameManager should label IPs with the selector(s) that match that name. So, if a DNS answer `www.cilium.io: 52.58.254.253, 18.192.231.252` is seen, and there is a ToFQDN policy that allows `*.cilium.io`, two new entries are inserted in to the IPCache metadata layer:
+
+```
+18.192.231.252/32 -> (dns:*.cilium.io)
+52.58.254.253/32 -> (dns:*.cilium.io)
+```
+
+If no identity exists for the label set `(dns:*.cilium.io)`, it would be allocated and the selector would be updated. Then, the updated mapping from prefix to identity would be inserted for both IPs in the IPCache.
+
+This optimizes a common case in the FQDN response path: learning a new IP. If the new IP does not require an identity allocation, then a only a single BPF write to the IP -> identity cache is needed -- neither the PolicyMaps nor Envoy need to be updated.
+
+#### FQDN identity pre-allocation
+
+We would like to optimize the dynamic path, so that learning a new IP generally does not require an identity allocation or policy update. One way to do this is to pre-allocate an identity for FQDN selectors, even if there are no corresponding IPs.
+
+This has advantages and disadvantages. It could cause unnecessary BPF PolicyMap entries, but reduces the critical-path latency. It also reduces the chance that packets are dropped on agent restart. We plan to enable pre-allocation.
+
+## Impacts / Key Questions
+
+### Question: Are overlapping FQDN selectors handled correctly?
+
+Consider two selectors: one selects `*.cilium.io`, another selects `www.cilium.io`. Imagine DNS responses have been seen for `www.cilium.io: IP1, IP2` and `dev.cilium.io: IP2, IP3`.
+
+The IPCache label map would have:
+
+```
+IP1: (dns:*.cilium.io, dns:www.cilium.io)
+IP2: (dns:*.cilium.io, dns:www.cilium.io)
+IP3: (dns:*.cilium.io)
+```
+
+There would, then, be two identities allocated:
+```
+ID1: (dns:*.cilium.io, dns:www.cilium.io)
+ID2: (dns:*.cilium.io)
+```
+
+and the IP -> ID mapping would be
+```
+IP1: ID1
+IP2: ID1
+IP3: ID2
+```
+
+The selector `*.cilium.io` selects `ID1, ID2`, and `www.cilium.io` selects `ID1`. Every selector derives to the correct set of IPs, thus overlapping selectors are handled correctly.
+
+
+### Question: Are overlapping IPs handled correctly?
+
+It is possible for the same IP to map to multiple names. Consider two names, `foo.com: IP1, IP2` and `bar.com: IP2, IP3`. If selector A selects `foo.com`, and selector B selects `bar.com`, the state of the IPCache should be
+
+```
+IP1: (dns:foo.com)
+IP2: (dns:foo.com, dns:bar.com)
+IP3: (dns:bar.com)
+```
+
+There would be 3 identities allocated for the 3 unique sets of labels, and the selectors would select these identities accordingly.
+
+So, overlapping IPs are handled correctly.
+
+### Question: Are overlapping ToFQDN and CIDR selectors handed correctly?
+
+Consider a selector, `*.cilium.io`, which currently selects one IP address, `52.58.254.253`. If, separately, a ToCIDR selector selects `52.58.254.0/24`, the state of the IPCache will be
+
+```
+52.58.254.0/24: (cidr:52.58.254.0/24)
+52.58.254.253/32: (dns:*.cilium.io) 
+```
+
+However, when the IP -> Identity mapping is calculated, because of label propagation, the complete set of labels for `52.58.254.253/32` is `(cidr:52.58.254.0/24, dns:*.cilium.io)`. So, the selector for `52.58.254.0/24` correctly selects both identities. Therefore, overlapping ToFQDN and CIDR selectors are handled correctly.
+
+
+### Impact: Identity churn
+
+While this proposal reduces operations for the common case (adding a new IP to an existing selector), it can increase identity churn for other cases. Identity churn is the allocation of a new identity for an existing prefix.
+
+- IPs that are selected by multiple ToFQDN selectors will churn identities unless the set of selectors is quiescent. Frequent garbage collection will also exacerbate churn.
+- Adding broad-based ToFQDN selectors may cause a large number of reallocations
+
+Identity churn is not a correctness issue, as policy updates are carefully sequenced so to not drop traffic. It is, however, a general load on the system.
+
+In general, this proposal should be a significant performance improvement. CIDR selectors are relatively static, whereas FQDN updates are highly dynamic (and have latency guarantees). But, it is a tradeoff. A cluster with dynamic, overlapping CIDR selectors may find the identity churn to be more costly than expected.
+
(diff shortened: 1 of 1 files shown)


cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,277 @@
...
+
+CIDR labels, especially IPv6 labels, are large. For example, the full set of labels for the CIDR `2001:db8:cafe::cab:4:b0b:0/112` is 2KiB. Every extra-cluster identity always has the full set of CIDR identities calculated for it.
+
+----
+
+## Proposal
+
+This CFP proposes 3 key changes. Together, they preserve correctness while radically restructuring how prefixes are selected.
+
+1: Labels on prefixes propagate downwards. That is to say, if prefix `192.0.0.0/22` has label `foo:bar`, then prefix `192.0.2.0/24` also has label `foo:bar`, unless overridden. The IPCache aggregates labels accordingly.
```
squeed, line 119: nit: GitHub doesn't recognize `1:` numbered lists, fix this.

joestringer: If I apply this more concretely to the current implementation, I think it translates to the following:

When the IPcache allocates a local Identity for an IP, it inspects all shorter prefixes in the IPCache and accumulates the labels in each shorter prefix until it gets the full set of CIDR (+ potentially other) labels. Then it allocates an Identity for that full label set.

Is that accurate? That's what I had in mind for the CIDRPolicy implementation originally.

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,277 @@
...
+2: Only selected CIDR labels are inserted in to the ipcache. If a selector references CIDR `192.0.2.0/24`, only the prefix -> label mapping `192.0.2.0/24 -> cidr:192.0.2.0/24` is added to the IPCache.
```
joestringer, line 120: sidenote: To implement CIDR policy, the Identity must be allocated after the CIDR policy is processed and that Identity must be (a) inserted into the bpf ipcache and (b) the Identity must be evaluated against the policies applied to local endpoints and inserted into the corresponding relevant policymaps. This process is unrelated to the FQDNs functionality described in this CFP, but already occurs and must still occur to correctly implement the CIDR policy feature.

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,277 @@
...
+3: When the NameManager learns about a new relevant `name -> IP` pair, it upserts the label `IP -> (dns:selector)` It **does not** create any CIDR labels.
```
joestringer, line 121: Ooo, nice. I do like that. Means that the new Identities are only allocated iff there is an overlapping CIDR policy statement. There will still be footguns, but less of them :pleading_face:

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,277 @@
...
+192.0.2.2/32 -> cidr:192.0.2.2/32, reserved:kube-apiserver
+```
+
+Then, the complete set of labels for `192.0.2.2/32` would be `k8s:foo:bar, cidr:192.0.2.0/24, cidr:192.0.2.2/32, reserved:kube-apiserver`.
+
+This is a cornerstone for implementing another feature, non-k8s label sources. It is also required for correctness within the context of this proposal, especially when ToFQDN and CIDR selectors overlap.
+
+#### Lifecycle
+
+With label down-propagation, it is important that updates are propagated accordingly. When a prefix P is updated or deleted, then updates must be also triggered for all prefixes contained in P.
```
joestringer, line 139: Incidentally in the worst case this can potentially mean reallocating every single local identity on the node, and of course we would need to stage that transition correctly to avoid packet loss during the migration.

We've had this idea in mind at different times during implementation - for instance the kube-apiserver IP can always migrate at some point, but I point this out because stringent testing will be necessary to really prove that these lifecycle operations work correctly. Also worth noting that in the absolute worst case we run out of local identity space to express the new Identities. Or short of that, we could run out of policymap space to express reachability.

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,277 @@
...
+
+This update may be calculated if we use a bitwise LPM trie. Alternatively, scanning the entire ipcache to determine affected prefixes may be more efficient.
```
joestringer, line 141: Why would scanning the entire ipcache be more efficient than doing a descendents iterator?

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,277 @@
...
+
+
```
joestringer, line 143: Not sure if you're collecting more lifecycle operations here but another case would be when a new name comes in and gets associated with the same IP. Similar Identity reallocation concern to above.

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,277 @@
...
+### 2: Only generate selected CIDR labels
+
+Currently, every selected prefix gets the full set of CIDR labels. In other words, the prefix `192.168.0.42/32` will have 32 labels generated. However, this is not truly necessary. The key insight is that the ipcache already knows about all selected CIDRs, since selectors are required to upsert relevant CIDRs with the IPCache before referencing them.
+
+So, if selector A selects `cidr:192.168.0.0/24`, there will be a single identity, `id1 -> (cidr:192.168.0.0/24)`. If selector B selects `cidr:192.168.0.0/25`, then there will be two identities: `id1-> (cidr:192.168.0.0/24)` and `id2 -> (cidr:192.168.0.0/25, cidr:192.168.0.0/24)`. Selector A will now select _id1_ and _id2_, and selector B will select just _id2_.
```
joestringer, line 148: The downside here is potential to increase the amount of dynamic state recalculation when a new policy is created.

I think we could alternatively explore an option where we avoid encoding each prefix's labels in the Identities and instead have an optional CIDR association in the local Identities. Then extend the label selector so that when it evaluates selections against an Identity with an embedded CIDR, it can do a CIDR subset check using straight math rather than label comparisons.

cilium/CFP-28427-fqdn-selectors.md
joestringer: Oh yeah you already mentioned this concern below under impacts.

cilium/CFP-28427-fqdn-selectors.md
squeed: I prototyped this in a different context, and discovered that trie iteration causes heap allocation, whereas a full map scan did not. In my benchmark, even though I reduced regular expression evaluations by a factor of 10, it was a 20% slow-down once GC time was added in.

Since scanning a small-ish map and checking for prefix coverage is alloc-less and cheap, I would expect it to outperform more complex data structures until map sizes grow very large.

In any case, this will be very easy to test.

cilium/CFP-28427-fqdn-selectors.md
squeed: Yes, that's what I had in mind for the implementation as well. And, as you mention, there's nothing special about `cidr:` labels; we will someday have "arbitrary" labels on prefixes.

The interesting bit -- and not immediately relevant to this bullet point -- is that multiple CIDRs may share identities.

cilium/CFP-28427-fqdn-selectors.md
squeed: Does ordering matter? I don't think so; I'm pretty sure the SelectorCache does the right thing. What if a CIDR policy selects a `/32` that has already been allocated by FQDN policy? It should work fine.

cilium/CFP-28427-fqdn-selectors.md
squeed: Bingo. Huge reduction in identity usage. Plus, adding an IP to an existing identity is cheap cheap cheap; just a single BPF write. Don't even need to update Envoy.

cilium/CFP-28427-fqdn-selectors.md
squeed: > The downside here is potential to increase the amount of dynamic state recalculation when a new policy is created.

Right, I had a section musing about this, but it never went anywhere and I axed it. Because of FQDN policies, we need to have a dynamic policy engine. I think it's important to _match impedances_, that is to say, try and have more frequent and latency-sensitive operations be less expensive, and infrequent operations be more expensive. 

Right now, ingesting policy and adding new IPs to an FQDN selector are near equivalent to the SelectorCache and IPCache. However, these take place in totally different contexts; one happens very occasionally and has latency measured in seconds (k8s strives for a pod creation time of ~2 seconds), and the other happens constantly and has a hard deadline of 100ms.

From this perspective, it makes sense to shift as much computation as possible to the slow, latent path. We can't take this too far, however, since the occasional slow operation still blocks any fast ones that happen to be waiting in line.

I am a bit concerned that a pathological sequence of ToCIDR policy updates could cause a surge of load. On the whole, however, I think this would be a huge improvement in experience. We essentially have that pathological sequence of updates now with S3, so it's unlikely to be any worse ;-).

cilium/CFP-28427-fqdn-selectors.md
joestringer: There's a few permutations to consider (sample, not comprehensive):

- No policy exists, create a new CIDR policy for this IP
- Presuppose a broader CIDR policy, then a new more specific CIDR policy is created
- Presuppose a more specific CIDR policy, then a new broader CIDR policy is created
- Presuppose an FQDN policy, then a CIDR policy with the same IP is created

I think that each of these cases more or less boils down to the same set of operations that need to occur in the correct order to prevent connection disruption due to (temporary) packet loss:

1. Cilium processes the CIDR policy
2. Cilium plumbs the policy down into the PolicyRepository, triggers Endpoint regeneration, leading to generation of the SelectorPolicy and generally preparing the SelectorCache so that when there *is* an identity, we can figure out how to plumb the policymaps correctly.
3. Somewhere along the way Cilium needs to inform the IPCache about the IP<->label mapping
4. The IPCache will react to the label mapping and allocate an Identity for the IP.
   1. If this happens before (2), then this will plumb the ipcache bpf map without updating the policymaps, and may lead to packet loss. Newly arriving traffic will hit the datapath, hit the ipcache lookup, resolve to the new identity, then subsequently hit the policy lookup and the policymap is not prepared for this new Identity which leads to drops.
      1. Given that this happens first, once (2) eventually _does_ occur, it will automatically calculate the new policy against the latest set of Identities including the one just allocated here in step (4) and the policymaps will converge to be "eventually consistent" and allow the traffic again. However by that point we've already disrupted the connection.
   2. If (2) happens before (4), then the ipcache will ensure correct ordering by first notifying the selectorcache about the newly allocated Identity, and waiting for the corresponding policymaps to be plumbed, then subsequently update the bpf ipcache.

Looking through these steps, if there is another policy already in place such as a broader CIDR policy or an FQDN policy, then step (2) is already in place and so we can guarantee case (2.ii) and all is fine. The tricky cases are when there is no specific policy for this IP yet, when the newly created policy is broader than the existing one, or for extra spice when there are overlapping properties (consider for instance a policy that is broader on L3 but more specific on L4 with L7 aspects or auth). Let's say for example if there is a /32 policy already, then we create a new /24 policy on port 80 with L7 filtering. Given the dynamic Identity creation proposed in this PR, there will already be Identity X for the /32 (no /24 label), then once all is processed there must be two identities, a new Identity X′ (/24 and /32 label) and Y (/24 label), the SelectorPolicy must be recalculated to account for the distinct subsets of traffic {/32 and port 80, /32 and not port 80, not /32 and port 80}, and of course the new bpf ipcache and policymap content must be calculated based on the new SelectorPolicy and Identities.

I think we actually _don't_ provide this level of ordering guarantee when calculating CIDR policies today; we only guarantee that policy calculation is _triggered_ before informing the ipcache, but don't guarantee that it _completes_ before informing the ipcache:

https://github.com/cilium/cilium/blob/12e3ae9936bd82924a995cbf2a2eb6284d6db0cb/daemon/cmd/policy.go#L464-L471

cilium/CFP-28427-fqdn-selectors.md
joestringer: I agree in principle with that path. I guess the idea I'm broaching in the counter-proposal above is whether we need to internally encode the IP Prefixes as strings at all. Probably for external interfaces like when we serialize flows on the wire we would need to include string representations of the CIDR labels, but maybe we don't have to internally track IP Prefixes in strings at all (if we just make the selectors, ipcache, etc a bit smarter wrt prefix handling and encoding the prefixes using netip.Prefix).

squeed: Thats a really interesting avenue of exploration. It does allow for short-cutting certain cases in this proposal that would require re-allocation.

I wonder if this is basically equivalent to having mutable labels on local identities. Have to stare at the wall for a bit and decide.

joestringer: My initial inclination is that mutable labels is a separate, more complex problem. We could start by stating that if we must encode an IP prefix into an Identity, there would only be one prefix and it would not change. If the prefix needs to change, we would allocate a new Identity. This is closer to what we have today.

squeed: Oooh I wonder if we can make this even easier: make the `cidr:xxx` label selector smarter and match longer contained prefixes.

squeed: > make the cidr:xxx label selector smarter and match longer contained prefixes.

Ah right, I remember why I didn't propose this: the lifecycle gets ugly with multiple overlapping ToFQDN and CIDR selectors. Label down-propagation solves so many problems so neatly.

joestringer: Ah. I guess part of this is if we try to visualize the ideal representation of the following cases:

- Only ToFQDN rules exist
- There is a ToFQDN rule and a specific CIDR rule or multiple CIDR rules
- There is a ToFQDN rule and a broad CIDR rule

Then work backwards from those "steady states" to figure out how we might represent these in the ipcache, identity and selector... and work through the lifecycles of how to transition into these states (eg starting from no policy -> tofqdn, then introduce CIDR policy, remove CIDR policy etc.).

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,277 @@
...
+**CIDR Selector:** A selector that references by CIDR. CIDR selectors are converted to a special form of label selectors.
+**ToFQDN Selector:** A selector that dynamically selects based on the result of intercepted DNS requests and responses.
+
+### CIDR labels
+
+The BPF policy engine relies on identities -- it does not understand CIDRs or IPs directly. So, special CIDR labels are used, and special _local identities_ are allocated on-demand to represent these external IP blocks.
+
+The IPCache aggregates multiple sources of labels for external CIDRs. As necessary, it will generate the complete set of labels and allocate a numerical identity.
+
+Consider a ToCIDR block that selects `192.0.2.0/24`. The policy subsystem will upsert the prefix-label mapping `192.0.2.0/24 (cidr:192.0.2.0/24, cidr:192.0.2.0/23, cidr:192.0.0.0/22, ..., cidr:0.0.0.0/0)`. The IPCache will allocate a local identity (e.g. 16777216) for this set labels. It will also convert the ToCIDR selector to a label selector, selecting the identical CIDR label. The SelectorCache will then select this numeric identity when generating the policy maps.
```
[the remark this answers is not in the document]
squeed: Fair question. Ultimately, the policy engine only works by selecting labels. However, we offer a much richer grammar, such as `toCIDRs:` and `fromEntities`, etc. We then convert the broader peer selection API to the equivalent pure-label selector.

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,277 @@
...
+## Proposal
+
+This CFP proposes 3 key changes. Together, they preserve correctness while radically restructuring how prefixes are selected.
+
+1: Labels on prefixes propagate downwards. That is to say, if prefix `192.0.0.0/22` has label `foo:bar`, then prefix `192.0.2.0/24` also has label `foo:bar`, unless overridden. The IPCache aggregates labels accordingly.
+2: Only selected CIDR labels are inserted in to the ipcache. If a selector references CIDR `192.0.2.0/24`, only the prefix -> label mapping `192.0.2.0/24 -> cidr:192.0.2.0/24` is added to the IPCache.
+3: When the NameManager learns about a new relevant `name -> IP` pair, it upserts the label `IP -> (dns:selector)` It **does not** create any CIDR labels.
+
+
+### 1: CIDR Label down-propagation
```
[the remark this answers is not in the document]
squeed: Iterating over all known prefixes is much more efficient than you might think :-). Burning a little extra CPU is much cheaper than allocating heap memory.

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,277 @@
...
+#### Lifecycle
+
+With label down-propagation, it is important that updates are propagated accordingly. When a prefix P is updated or deleted, then updates must be also triggered for all prefixes contained in P.
+
+This update may be calculated if we use a bitwise LPM trie. Alternatively, scanning the entire ipcache to determine affected prefixes may be more efficient.
+
+
+### 2: Only generate selected CIDR labels
+
+Currently, every selected prefix gets the full set of CIDR labels. In other words, the prefix `192.168.0.42/32` will have 32 labels generated. However, this is not truly necessary. The key insight is that the ipcache already knows about all selected CIDRs, since selectors are required to upsert relevant CIDRs with the IPCache before referencing them.
```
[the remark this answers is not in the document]
squeed: Good question. Policy updates and CIDR label updates are already serialized.

squeed:
I just had another idea: identity pre-allocation.

One of the biggest causes of FQDN tail latency is updating Envoy with new identities for a given selector. What if we pre-allocate an identity for every selector? Then we **don't need to update Envoy** in most cases! (This is because marking an IP as having an existing identity is just an ipcache update, not a policy update). The only time an incremental policy update would be required is if an IP was newly selected by two selectors.

gandro:
> I just had another idea: identity pre-allocation.
> 
> One of the biggest causes of FQDN tail latency is updating Envoy with new identities for a given selector. What if we pre-allocate an identity for every selector? Then we **don't need to update Envoy** in most cases! (This is because marking an IP as having an existing identity is just an ipcache update, not a policy update). The only time an incremental policy update would be required is if an IP was newly selected by two selectors.
> 
> This would be a huge improvement in tail latency.

I thought about per-allocation as well. It really depends how much overlap there is between selectors. If you have a `matchPattern: *` selector, it will be part of every identity. So if you just pre-allocate an identity for `dns:cilium.io`, that identity itself will never be used. Instead, all observed IPs will have two labels: `dns:cilium.io,dns:*`

squeed:
@gandro 

> I thought about pre-allocation as well. It really depends how much overlap there is between selectors.

Indeed, and we may with wish to be clever when there is a wildcard selector. Pre-allocation, however, doesn't have any relevant downsides (1 identity per FQDN selector, so O(N) = 100), and it optimizes the policy update case when there is no overlap. Which would be a huge win for tail latency; anything we can do to prevent a policymap + envoy update is an improvement.

squeed:
To all reviewers: this has been updated based on initial prototype implementation. It is now ready for final review.

cilium/CFP-28427-fqdn-selectors.md
squeed: For existing CIDR selectors, if a given IP changes identities due to a label being added, that change can be rolled out without dropping traffic. This is because the ipcache is very careful about the order in which it interlaves operations.

For *new* selectors, traffic may be dropped. This is unrelated to the proposed FQDN refactor. I actually have a branch that fixes this, but it's not yet ready for merging.

cilium/CFP-28427-fqdn-selectors.md
squeed: One minor thing we should consider doing is breaking up the ipcache work loop in to chunks, on the very unlikely case we temporarily need too many local identities. That way we only spike identity usage by the chunk size, not by doubling the number of existing identities.

cilium/CFP-28427-fqdn-selectors.md
squeed: update: this has currently merged with simple scanning. If we need to move this to prefix-aware selection, we can do so as needed.

cilium/CFP-28427-fqdn-selectors.md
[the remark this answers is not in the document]
squeed: I removed this section; we no longer expand CIDR identities. Let me know if you have more questions.

squeed:
We've not been able to come up with a circuit breaker, since the scope for identity allocation is already so small. The existing limitation on the maximum number of names per endpoint is probably sufficient.

cilium/CFP-28427-fqdn-selectors.md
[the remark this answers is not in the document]
gandro: We ran the numbers here - even for large clusters, this is acceptable: https://github.com/cilium/cilium/pull/32578#issuecomment-2133794216

Marking as resolved

squeed:
@joestringer I think this CFP is probably ready to merge :-). If you'd like to take a final pass, feel free.

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,409 @@
...
+
+#### Example
+
+The IPCache has three prefixes, each with a corresponding identity.
+
+- `192.168.0.0/24` -> `(cidr:192.168.0.0/24)`
+- `1.1.1.1/32` -> `(fqdn:*, fqdn:one.one.one.one)`
+- `2.2.2.2/32` -> `(fqdn:*)`
+
+If, on agent restart, all prefixes have the same set of labels, then the identity mapping is stable. If, however, the FQDN checkpointing is not exact (which does happen when an IP is near expiration), then `1.1.1.1/32` may change labels to `(fqdn:*)`. However, since the identity for `fqdn:*` is already allowed by policy maps, no traffic to `fqdn:*` is dropped.
```
joestringer, line 332: side comment: I suppose checkpointing is also not 100% foolproof either. IIRC the current name tracking per-endpoint is done every 5s or so, so there's always a race condition of whether the most up-to-date info is checkpointed to disk. Depending on the checkpointing mechanism there could be a similar blip with this as well.

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,409 @@
...
+
+Thus, transitioning to label-based, rather than prefix-based allocation should preserve policy stability on restarts.
+
+### Key question: Circuit breakers?
+
+There is currently a circuit-breaker, `--tofqdns-endpoint-max-ip-per-hostname`, which is intended to prevent names such as S3 causing identity churn. In this proposal, large names such as S3 should not allocate more than one or two identities, making this less critical.
+
+However, this is still a security-sensitive area, since we are allocating resources based on external, untrusted packets. We will need a sane circuit-breaker to ensure we do not allocate a large number of identities based on malicious input.
```
joestringer, line 340: I think it's fair to say that the trust in the DNS server is also quite critical in general with the ToFQDNs feature, in that if the path to the DNS server or the DNS server itself is somehow compromised, or if the user allows DNS requests to arbitrary destinations and used these as input to the FQDN engine, then these situations could each cause adverse behavior. 

Then there's the 'selector combinations' side of things which is also subject to the network policies configured by the k8s namespace users.

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,409 @@
...
+
+The maximum number of identities allocated by ToFQDN selectors is the number of unique selector combinations seen for given IPs. The theoretical maximum is, therefore, $2^N-1$[^1]. 10 distinct selectors would be OK (1024). More than 16 selectors could *theoretically* exceed identity space.
```
joestringer, line 342: Do you think there's something here we could export as a metric to at least measure the risk / likelihood of running into an issue with this?

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,409 @@
...
+existing labels based on certain upgrade heuristics.
+
+
+#### Downgrade:
+
+The constraints for downgrades are somewhat reduced; it is acceptable for brief policy drops while the agent finishes converging.
+
+On downgrade to v1.15, the restoration logic will, again, upsert (cidr:xxx/32) labels and reserved numeric identities in to the ipcache. However, for v1.16 more than one IP may share the same local identity. Thus, one IP will "capture" the numeric ID, while the others will change. Connections to those IPs may be briefly interrupted only during endpoint regeneration. Once regeneration is complete, there should be no further disruption.
+
+As a precautionary measure, v1.15.tip should be updated
```
joestringer, line 371: Updated how?

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,409 @@
...
+
+The constraints for downgrades are somewhat reduced; it is acceptable for brief policy drops while the agent finishes converging.
+
+On downgrade to v1.15, the restoration logic will, again, upsert (cidr:xxx/32) labels and reserved numeric identities in to the ipcache. However, for v1.16 more than one IP may share the same local identity. Thus, one IP will "capture" the numeric ID, while the others will change. Connections to those IPs may be briefly interrupted only during endpoint regeneration. Once regeneration is complete, there should be no further disruption.
+
+As a precautionary measure, v1.15.tip should be updated 
+
+## Future Milestones
+
+TODO
```
joestringer, line 375: Maybe drop the `TODO` or flesh this out?

cilium/CFP-28427-fqdn-selectors.md
```diff
@@ -0,0 +1,409 @@
...
+
+1. An identity is allocated for the IP:
+  a. `ipc.UpsertPrefixes()` is called, associating the complete set of CIDR labels with this IP address.
+  b. The IPCache determines the IP's full set of labels, allocates an identity, and informs the SelectorCache of this new IP
+2. The selector is updated:
+  a. The NameManager calls `sc.UpdateFQDNSelector()`, updating the set of selected IPs for this given FQDNSelector
+  b. The SelectorCache determines the complete set of identities selected
+  c. Any differences are pushed to network policies
+3. Endpoints are updated
+  Any pending incremental updates are batched up and applied to the endpoints' BPF PolicyMap. If needed, Envoy is also updated
```
joestringer, line 99: nit: Markdown formatting here didn't quite render out well. I suspect a little more indentation would help.

joestringer (approved):
Overall this looks pretty comprehensive, I like the way that various cases were called out like the combinations of potentially overlapping label sets, as well as upgrade/downgrade consequences. I dropped a few notes and discussion points below, but I agree this is good enough to merge (and implement) :+1:

cilium/CFP-28427-fqdn-selectors.md
gandro: Thanks, fixed!

cilium/CFP-28427-fqdn-selectors.md
gandro: I've dropped it. There are no major future milestones that I think need to be part of the CFP. Minor future work can be tracked in issues.

cilium/CFP-28427-fqdn-selectors.md
gandro: I've removed this, since I don't think this is necessary.

However, I've added a note to in the upgrade section that users need to upgrade to v1.15.tip for dropless upgrades (c.f. https://github.com/cilium/cilium/pull/32872)

cilium/CFP-28427-fqdn-selectors.md
squeed: OK, added chunking to the CFP (and filed a PR for it :-p)

cilium/CFP-28427-fqdn-selectors.md
squeed: Thinking very hard about this, I don't think this proposal significantly changes the security scope or attack surface of the DNS proxy. If an FQDN selector selects an attacker-controlled domain, then the attacker can insert any IPs they want in to policy (waves hand).

Perhaps this is worth a better writeup (@ferozsalam :smile_cat: ).

cilium/CFP-28427-fqdn-selectors.md
squeed: We do have a metric for the total number of local identities. We can easily add an alert for agents with a large number of local identities. This would be a very good start.

joestringer: I see, that is probably good enough. I was wondering about this "multiple selectors selecting the same IP" aspect since that can end up generating double the number of identities for each new selector selecting the IP. But this is _probably_ getting too much into the weeds for metrics?

squeed: From an end-user perspective, the interesting thing is memory usage and policy map size. It doesn't matter what leads to those identities, just that they're there -- 10,000 fqdn selectors vs. a pathological combination of 16 of them, that's all the same to the admin.

AIUI we already have alerts for overfull policy maps (I hope!). An alert for excessive local identities would be a good addition to this, I'll look in to adding this.

gandro: I will add an optional follow-up item to think about if there are metrics we can add. Having some metric that indicates that the user chose selectors that are risking combinatorial explosion would be nice to have, as it would be a more targeted signal than "policy map is full" or "allocator is out of identities"

cccsss01:
Still having problems w/ v1.16.0

joestringer:
@cccsss01 could you please file a full issue on https://github.com/cilium/cilium ?

PR #27, opened 2024-03-14
Add CFP-28986 with details of mutual auth updates

youngnick:
This PR adds a CFP design for updates to mutual authentication to address reported security issues, as per cilium/cilium#28986.


cilium/CFP-28986-mutual-auth-updates.md
```diff
@@ -0,0 +1,256 @@
...
+* Attacker has to be able to delete the source pod so it releases it’s IP. Otherwise attacker has to wait for the pods to scale down by the autoscaler or some other reason and has to identify IPs which have been released. All of this has to happen while the destination node remains detached from the apiserver.
+* Attacker must have access to the IPAM pool (when in use) of the original source pod
+* Attacker has to be able to schedule enough pods to re-use the released pod IP before anyone else does. If any pods get scheduled by other users in parallel, any other new pod may pick up the IP. All of this needs to happen while the destination node continues to remain without access to the apiserver.
+* In non trivial clusters using per node PodCIDRs, attacker typically needs to be able to use a nodeSelector to avoid spreading pods on arbitrary nodes with different PodCIDRs.
+* No egress policies deny policies affecting the scheduled source pod may exist
```
lizrice, line 87:
```suggestion
* No egress deny policies affecting the scheduled source pod exist
```

cilium/CFP-28986-mutual-auth-updates.md
```diff
@@ -0,0 +1,256 @@
...
+* As soon as the destination node regains apiserver access, any ongoing, incorrectly allowed packet flow gets dropped even for established connections.
```
lizrice, line 88: Before this bullet point, for clarity I think we could say something like "If all these steps happen, the attacker can send traffic from the new source to the destination pod, as authentication will continue to use the cached identity from the old source pod"

cilium/CFP-28986-mutual-auth-updates.md
```diff
@@ -0,0 +1,256 @@
...
+## Motivation
+
+All of the above advantages are why the mutual authentication feature was built this way to begin with, but it's become apparent that there are a set of circumstances where the current design will fail when attacked.
+
+Those circumstances are:
+- The cluster must be using native routing or some other method that does not encapsulate the packets (where the encapsulation carries the CiliumIdentity, as in VXLAN and Wireguard)
+- _Some_ nodes in the cluster must lose connection to the apiserver somehow. In one demonstration, this was performed using iptables rules on a host, which would require elevated privileges, but it's also possible, if unlikely, that this situation would arise in some other network failure scenarios.
+- An attacker must be able to launch new pods on one of the nodes while the above two conditions hold.
+
+Then, as Thomas Graf laid out [in the Cilium Slack](https://cilium.slack.com/archives/C02QKQDTVDX/p1701466203218329?thread_ts=1701466174.930209&cid=C02QKQDTVDX)
```
lizrice, line 79: Rather than repeat it verbatim I think there are some re-wordings possible for clarity

cilium/CFP-28986-mutual-auth-updates.md
```diff
@@ -0,0 +1,256 @@
...
+
+* Attacker has to identify a source / destination pod pair
```
lizrice, line 81:
```suggestion
* Attacker has to identify an existing, mutually authenticated source / destination pod pair. Attacker is going to replace the source pod with a replacement pod that will be able to communicate with the destination.  
```

cilium/CFP-28986-mutual-auth-updates.md
lizrice: Based on text from below, would it also be true to say: 

This traffic can flow until either 
*  the destination node reconnects to the API server, causing the IP cache to be updated, or 
* a mutual authentication handshake is re-initiated because one of the certificates expires (by default certificates are rotated every 30 minutes)

cilium/CFP-28986-mutual-auth-updates.md
```diff
@@ -0,0 +1,256 @@
...
+
+The first idea makes cache manipulation irrelevant, because the inbound packets include the security identity in a header.
+
+However, in the existing mode, mutual authentication handshakes are only initiated when there is a change in either endpoint state (that is, a new node to node connection from a new pod), or when a certificate is rotated. So, in the event that an attacker _did_ manage to compromise mutual authentication mechanisms somehow, the window is, at maximum, the certificate rotation period, which defaults to 30 minutes.
+
+Adding a connection-based mutual authentication mode significantly reduces the timescale that an attack is exploitable for - since the mutual auth handshake is performed on each new connection, and connections _generally_ last much less that 30 minutes.
+
+It also makes performing the exploit harder, since the full 5-tuple (source IP, source port, destination IP, destination port, protocol) must match, rather than just source and destination IP, _and_ must also migrate the actual connection from one application to another. Still possible, but orders of magnitude harder.
```
lizrice, line 123: I don't find this very convincing. We know that the source and destination IP address match as that's a prerequisite for the attack. Matching the protocol is trivial. Is matching ports that difficult?

cilium/CFP-28986-mutual-auth-updates.md
```diff
@@ -0,0 +1,256 @@
...
+
+This CFP lays out a proposal to mitigate this risk in two ways:
+* Require the use of either encryption (recommended because it prevents man-in-the-middle attacks) or encapsulation (which prevents cache manipulation but not MITM). In both of these cases, the Cilium identity is included in every packet, and so the ipcache is not used to impute identity to packets.
+* Add an additional mutual authentication mode that will use the connection tracking ("conntrack") table in the conntrack BPF map to store information about what _connections_ need authentication, instead of what _identity-identity-node pairings_ need authentication.
+
+The first idea makes cache manipulation irrelevant, because the inbound packets include the security identity in a header.
+
+However, in the existing mode, mutual authentication handshakes are only initiated when there is a change in either endpoint state (that is, a new node to node connection from a new pod), or when a certificate is rotated. So, in the event that an attacker _did_ manage to compromise mutual authentication mechanisms somehow, the window is, at maximum, the certificate rotation period, which defaults to 30 minutes.
```
lizrice, line 119: I think we can talk about this earlier, in the description of the attack

cilium/CFP-28986-mutual-auth-updates.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Overview
+
+This CFP proposes two changes: enforcing every packet contains an identity using either encryption or encapsulation and adding a per-connection mutual auth mode. These changes are relatively orthogonal and could be implemented as two separate changes.
+
+#### Requiring encryption or encapsulation resolves cache manipulation
+
+This assertion depends on two things that may not be clear:
+
+- Wireguard encryption now (as of Cilium 1.15) also encapsulates packets into VXLAN as they transit the tunnel, adding a Cilium identity into the header (as in unencrypted VXLAN)
```
lizrice, line 144: IPSec as well?

cilium/CFP-28986-mutual-auth-updates.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Connection-based mutual authentication method
+
+The connection-based mutual authentication method will:
+* Be selectable as an entirely new authentication method (called `connection-spiffe` or similar)
+* Use the conntrack table to keep track of connections
+* Add two bits of marking into conntrack table entries, "Auth required", and "Auth completed"
+* Use the existing TLS handshake code, but call it on a different event (new connection instead of new identity-identity-node tuple).
+
+The existing conntrack table performs standard connection tracking functions, tracking connections based on the 5-tuple of source IP, source port, destination IP, destination port, and protocol. TCP is handled using its connection mechanisms, and UDP is handled using the same time-based pseudo connection that is standard across most networking gear.
```
lizrice, line 177: I don't know enough about UDP pseudo connection, but if we do per-connection auth for UDP, what are the chances that we drop so many packets during mutual auth that the UDP traffic is ineffective?

cilium/CFP-28986-mutual-auth-updates.md
```diff
@@ -0,0 +1,256 @@
...
+
+Those circumstances are:
+- The cluster must be using native routing or some other method that does not encapsulate the packets (where the encapsulation carries the CiliumIdentity, as in VXLAN and Wireguard)
+- _Some_ nodes in the cluster must lose connection to the apiserver somehow. In one demonstration, this was performed using iptables rules on a host, which would require elevated privileges, but it's also possible, if unlikely, that this situation would arise in some other network failure scenarios.
+- An attacker must be able to launch new pods on one of the nodes while the above two conditions hold.
+
+Then, as Thomas Graf laid out [in the Cilium Slack](https://cilium.slack.com/archives/C02QKQDTVDX/p1701466203218329?thread_ts=1701466174.930209&cid=C02QKQDTVDX)
+
+* Attacker has to identify a source / destination pod pair
+* Attacker has to cut apiserver access of the destination node or otherwise delay any updates from the apiserver to that node for a prolonged time while the source node remains connected to the apiserver. The entire apiserver not being available is not enough as no pods could be deleted or scheduled (next step)
```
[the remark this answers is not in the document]
youngnick: I think this would depend on how the container runtime considers that a container stops running. But yeah, I'll grant that if you can start something else in the process and network namespace without apiserver access, then you could bypass this. I'll update the wording.

cilium/CFP-28986-mutual-auth-updates.md
```diff
@@ -0,0 +1,256 @@
...
+* Attacker has to be able to delete the source pod so it releases it’s IP. Otherwise attacker has to wait for the pods to scale down by the autoscaler or some other reason and has to identify IPs which have been released. All of this has to happen while the destination node remains detached from the apiserver.
+* Attacker must have access to the IPAM pool (when in use) of the original source pod
```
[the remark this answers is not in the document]
youngnick: Yes, the attacker needs to be able to schedule another pod that will end up with the same IP address.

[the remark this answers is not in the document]
youngnick: Updated.

cilium/CFP-28986-mutual-auth-updates.md
youngnick: I added a quick summary and then tidied up the more detailed steps.

cilium/CFP-28986-mutual-auth-updates.md
youngnick: It's not the destination port that's hard, it's actually the source port. The source port for connections is usually randomly chosen by the network stack, so simply starting a new connection will probably not work. The attacker will need to inspect existing flows and ensure that a new flow matches the existing source port.

In addition, the creation of a new connection will (for TCP connections) involve a new handshake, which should trigger the partially-closed parts of the datapath to clear the connection properties (including the mutual auth bits, as mentioned below).

cilium/CFP-28986-mutual-auth-updates.md
youngnick: Added as per your other suggestion.

cilium/CFP-28986-mutual-auth-updates.md
youngnick: I'm not sure - the IPSec code has been going through signfiicant refactoring in the last few months, so I thought it was better to leave it out.

cilium/CFP-28986-mutual-auth-updates.md
youngnick: Reasonable, depending on how the application using UDP treats its UDP packets. Many UDP applications essentially build their own connection handling on top of UDP (HTTP3 is a good example).

A UDP pseudo-connection is created by most connection-tracking tools by associating UDP packets matching a 5-tuple (and the equivalent return traffic) into a "connection", even though it's only a guess that they actually are a connection. These pseudo-connections usually have a timer that has them be removed once no packets have matched for a certain period of time.

What this means in this case is that UDP traffic will also be captured by the conntrack table, even though there's not actually any connection involved.

cilium/CFP-28986-mutual-auth-updates.md
youngnick: I added some explanation of the latter part.

youngnick:
@lizrice or anyone else, any further updates needed here for now?

cccsss01:
Would love this feature!

xmulligan:
@youngnick do you think we should merge this as is?

youngnick:
Yes, I think we should merge as-is, and update again the next time we update mutual auth.

xmulligan (approved):

xmulligan:
Once we merge #42 we should also add the status here and update it as needed.

PR #29, opened 2024-04-03
Add CFP-30788 Managing Hubble Metrics cardinality

vakalapa:
This proposal aims at adding a new dynamic mechanism to manage Hubble metric cardinality.

### DIFF
--- a/hubble/CFP-30788-managing-hubble-metrics-cardinality.md
+++ b/hubble/CFP-30788-managing-hubble-metrics-cardinality.md
@@ -1,0 +1,160 @@
+# CFP-30788: Managing Hubble Metrics cardinality
+
+**SIG: SIG-HUBBLE-API/Observability**
+
+**Begin Design Discussion:** 2024-03-28
+
+**Cilium Release:** -
+
+**Authors:** Vamsi Kalapala <vakr@microsoft.com>, Neha Aggarwal <neaggarw@microsoft.com>
+
+## Summary
+
+Hubble metrics are very powerful and can be used to monitor a wide variety of network traffic. However, the cardinality of the metrics can grow very quickly and can lead to performance issues. This CFP proposes a mechanism to manage the cardinality of Hubble metrics.
+
+## Motivation
+
+Consider a large scale cluster with thousands of pods in a Cloud Environment. Cloud environment managed prometheus offerings have costs associated at multiple levels, including storage, query, and ingestion. With auto-scaling enabled in the cluster and need for in-depth monitoring, the cardinality of the metrics can grow very quickly and can lead to performance, cost, and scalability issues.
+
+Usually, Cluster admins would be able to control cardinality of the metrics ingested by applying custom ingestion rules on the prometheus configuration. However, there are scenarios where this may not be ideal, for example:
+
+- When the cluster is using a managed prometheus offering where the prometheus configuration is managed by the cloud provider and custom ingestion rules cannot be applied.
+- When a high scale cluster is generating high cardinality of metrics putting pressure on the prometheus ingestion, storage and query performance.
+- When the cluster admin wants to dynamically control the cardinality of the metrics based on the cluster state without having to restart cilium agent or prometheus server. 
+
+Such issues present a unique challenge to cluster admins and developers. By creating a mechanism to manage the cardinality of Hubble metrics dynamically based on the cluster state would help in addressing the above challenges. Recent change for [Hubble Flows dynamic exporter](https://docs.cilium.io/en/latest/observability/hubble-exporter/) has inspired this proposal.
+
+## Goals
+
+* Ability to manage the cardinality of Hubble metrics dynamically.
+* Should be able to control labels and dimensions of the metrics.
+
+
+## Non-Goals
+
+* Recommendations for Prometheus configuration at ingestion time.
+
+## Proposal
+
+### Overview
+
+We propose a new config map wrapped functionality in Hubble metrics which can be used to manage the cardinality, dimensions and targets of the metrics. Drawing from [Hubble Flows dynamic exporter](https://docs.cilium.io/en/latest/observability/hubble-exporter/) which uses a standalone config map to manage exporting of flows, we propose a similar mechanism. 
+
+Existing behavior of reading cilium-config for `hubble-metrics` be preserved for backward compatibility. 
+
+A new config on cilium agent called `enable-dynamic-hubble-metrics` will be introduced which when set to true, will enable the dynamic management of Hubble metrics. When this config is set to true, the cilium agent will read the `hubble-metrics-config` config map to get the configuration for the metrics. When this config is set to false, the cilium agent will read the `cilium-config` for `hubble-metrics` configuration.
+
+### Config Map
+
+A new config map called `hubble-metrics-config` will be introduced which will have the following fields:
+
+- `metrics`: List of metrics to be managed. If no metrics are defined in this config map then essentially Hubble should not export any metrics. Each metric will have the following fields:
+  - `name`: Name of the metric.
+  - `contextOptions`: List of [context options](https://docs.cilium.io/en/latest/observability/metrics/#context-options) to be applied to the metric.
+  - `includeFilters` : List of include flow filters to be applied to this particular metric.
+  - `excludeFilters` : List of exclude flow filters to be applied to this particular metric.
+
+### Example
+    
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+    name: hubble-metrics-config
+    namespace: kube-system
+data:
+    hubblemetrics.yaml: |
+        hubbleMetrics:
+            metrics:
+            - name: flow
+                contextOptions: 
+                    - name: sourceEgressContext
+                      values:
+                        - pod
+                    - name: destinationIngressContext
+                      values:
+                        - pod
+                includeFilters: 
+                    - []
+                excludeFilters:
+                    - source_pod: ["namespace1/"]
+                    - destination_pod: ["kube-system/"]
+            - name: dns
+                contextOptions: 
+                    - name: labelsContext
+                      values:
+                        - source_namespace
+                        - source_pod
+                    - name: destinationIngressContext
+                      values:
+                        - pod
+                includeFilters: 
+                    - source_pod: ["namespace1/"]
+```
+
+This will make sure we will have a dynamic way to manage each metric's cardinality, dimensions and targets.
+
+
+## Impacts / Key Questions
+
+### Impact: Managing Hubble Metrics Cardinality
+
+This proposal intents to give more control of metrics to the cluster admins and developers. This will help in managing the cardinality of the metrics dynamically based on the cluster state without having to restart cilium agent or prometheus server. 
+
+### Impact: Dashboarding and Alerting
+
+### Will this proposal affect the existing dashboards and alerting rules?
+
+By keeping the existing behavior of reading cilium-config for `hubble-metrics` for backward compatibility, this proposal will not affect the existing dashboards and alerting rules.
+
+### Will this proposal affect dashboards built using the dynamic metrics config?
+
+Yes there is a possibility that the dashboards built using the dynamic metrics config may be affected if the properties of the metrics are changed dynamically. This will be a trade-off between the dynamic management of metrics and the stability of the dashboards. The cluster admins and developers should be aware of this trade-off.
+
+Also, this problem exists even with the existing behavior of reading cilium-config for `hubble-metrics`. New context options could be added to the metrics which may not be supported by the existing dashboards. 
+
+## Implementation Alternatives
+
+### Option 1: Using CRD approach instead of ConfigMaps 
+
+CRDs are good alternative to approach solving this problem, there was good amount of discussion done previously [here](https://github.com/cilium/cilium/pull/28220#issuecomment-1741109225). For this proposal I built upon the existing precedence so there is uniformity on how Hubble is configured for dynamic flow logs and metrics.
+
+#### Pros
+
+* Versioned and changes are preserved. 
+* Can implement complex validation logic
+
+#### Cons
+
+* Unlike ConfigMaps, CRDs are heavy-weight. In real-world scenarios, my assumption is that the dynamic hubble metrics config map will not be changed often, so CRD based continuous watcher might be overkill.
+* Based on existing precedence set by Hubble dynamic flow logs configmap, users of Cilium do not have a re-learn a new way of configuring Hubble via CRD. 
+
+Open to discuss more on this. 
+
+### Option 1: Move only the flow filters to the dynamic metrics config
+
+
+#### Pros
+
+* This will reduce the re-learning aspect of existing way to configure hubble metrics.
+
+#### Cons
+
+* It creates a split brain scenario where some of the same metric properties are configured in cilium-config and some in hubble-metrics-config.
+
+### Option 3: Introduce annotations to manage the cardinality of the metrics
+
+#### Pros
+
+* Easy to apply the annotations to the pods/namespaces
+
+#### Cons
+
+* No central place to manage which pods/namespaces have the annotations applied.
+
+
+## Future Milestones
+
+### Deprecation of existing hubble-metrics in cilium-config
+
+This proposal is focusing on discussing new way of configuring hubble-metrics out-of-band from cilium-config. Deprecation of existing configuration method is necessary to reduce different code paths and complexity in configuring Hubble metrics. Current consensus is to re-visit deprecation of hubble metrics in cilium-config after 2 major versions of cilium. This gives enough time for new configMap approach to be validated and tested enough.


cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,130 @@
...
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+    name: hubble-metrics-config
+    namespace: kube-system
+data:
+    hubblemetrics.yaml: |
+        hubbleMetrics:
+            enabled: true
```
anubhabMajumdar, line 70: Do we need the `enabled` field? There's already controls to disable Hubble entirely. So, even if this CM is applied and Hubble is disabled, the reconcile is basically no-op.

chancez: I think it's useful to have an enabled/disabled option here, since the only other way to disable is to not configure metrics in the first place. Having a config option to enable/disable means something else managing the configuration out-of-band (eg: a separate add-on control plane for integrating with a hosted K8s solution), means that it could be toggled on/off without restarting cilium. (in theory)

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,130 @@
...
+A new config called `enable-dynamic-hubble-metrics` be introduced which when set to true, will enable the dynamic management of Hubble metrics. When this config is set to true, the cilium agent will read the `hubble-metrics-config` config map to get the configuration for the metrics. When this config is set to false, the cilium agent will read the `cilium-config` for `hubble-metrics` configuration.
+
+### Config Map
+
+A new config map called `hubble-metrics-config` will be introduced which will have the following fields:
+
+- `enabled`: Boolean value to enable/disable the dynamic management of Hubble metrics.
+- `metrics`: List of metrics to be managed. Each metric will have the following fields:
+  - `name`: Name of the metric.
+  - `contextoptions`: List of [context options](https://docs.cilium.io/en/latest/observability/metrics/#context-options) to be applied to the metric.
```
chancez, line 54: Should probably be camel case.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,130 @@
...
+  - `labels`: List of labels to be applied to the metric.
```
chancez, line 55: How do you expect this to work? Just hardcoded static labels?

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,130 @@
...
+            metrics:
+            - name: flow
+                contextoptions: sourceEgressContext=pod;destinationIngressContext=pod
+                includeFilters: []
+                excludeFilters:
+                    destination_pod: ["kube-system/"]
+            - name: dns
+                contextoptions: labelsContext=source_namespace,source_pod
+                includeFilters: 
+                    source_pod: ["namespace1/"]
```
chancez, line 80: List of filters here too
```suggestion
                   - source_pod: ["namespace1/"]
```

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,130 @@
...
+    
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+    name: hubble-metrics-config
+    namespace: kube-system
+data:
+    hubblemetrics.yaml: |
+        hubbleMetrics:
```
chancez, line 69: Do we need this top level field?

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### Pros
+
+* This will reduce the re-learning aspect of existing way to configure hubble metrics.
+
+#### Cons
+
+* It creates a split brain scenario where some of the same metric properties are configured in cilium-config and some in hubble-metrics-config.
+
+### Option 2: Introduce annotations to manage the cardinality of the metrics
```
chancez, line 117: Not exactly the same, but I believe with some well-crafted `FlowFilters`, you could in theory do something similar with pod labels using your main proposal. Eg: you could write a config like:
```
- name: flow
  includeFilters:
    - source_label: ["hubble-flow-metrics=true"]
```

Which would then mean any pod with the label `hubble-flow-metrics` and a value of `true`  will produce flow metrics. 

That's one reason I like the config file approach; admins can basically centralize config; or craft configs that let users manage their metrics (or some combination of the two).

chancez:
Overall idea makes sense to me. We've got a few people on the hubble team who are on holiday this week; so hopefully next week they can take a look and we can move forward.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
vakalapa: This the field for the optional fields each metrics support ? https://docs.cilium.io/en/latest/observability/metrics/#id12
Are these defined in contextOptions itself today ?

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,130 @@
...
+metadata:
+    name: hubble-metrics-config
+    namespace: kube-system
+data:
+    hubblemetrics.yaml: |
+        hubbleMetrics:
+            enabled: true
+            metrics:
+            - name: flow
+                contextoptions: sourceEgressContext=pod;destinationIngressContext=pod
```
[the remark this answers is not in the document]
vakalapa: This is definitely a quality of life improvement from today. Configuring the metrics via existing configmap is not easy and very error prone. A structure data structure would bring ease of use. Let me try to expand this.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
vakalapa: That is a good idea, so this proposal infact is more customizable per admins' requirements on how they want to expose it.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
vakalapa: Removing, it makes sense to fold them under contextOptions.

chancez: `labels` isn't a context option; the link your referrencing is describing the labels the metric defines; they're not configurable.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
chancez: Fair enough; I'm not to picky here

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
chancez: Exactly

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
vakalapa: Can you expand on this a bit, i am thinking this enabled/disabled could be a redundant field, source of truth for if this feature is enabled will be in `cilium-config` right ? Even if admin disables here in this configmap, hubble willl read the `enable-dynamic-hubble-metrics` in `cilium-config` and keep the feature enabled?

I am inclining towards removing this flag from new configmap.

chancez: Basically, imagine your a K8s as a service provider and you have cilium as part of your offering. Users cannot control the cilium install via helm, or by modify the cilium-config, etc. 

So as part of your hosted k8s offering, you have a separate microservice which creates the `hubblemetrics.yaml` configmap based on the settings the user has set in your web UI. One option is to disable metrics which would set `hubbleMetrics.disabled=true` in the configmap and cilium will not configure any Hubble Metrics. Later on, the user can enable Hubble Metrics, which will change the configmap to have `hubbleMetrics.disabled=false`. 

This means the Hubble Metrics feature can be completely controlled without having to restart cilium or anything.

glibsm: Is this enable/disable of metrics overall, or _dynamic_ metrics?

vakalapa: below is my assumption on how the ordering of config maps will work:

1.  `cilium-config` : `enable-hubble `- will enable Hubble itself
2. `cilium-config` :  `enable-dynamic-hubble-metrics`  - will make any metrics related config in cilium-config obsolete and will consider `hubble-metrics-config` as the only source of truth
3. `hubble-metrics-config`: `enabled/disabled` will enable/disable hubble metrics 

Correct me if my thinking is incorrect.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
vakalapa: I removed it.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
neaggarwMS: Can we also add a global source label to allow all metrics in a given pod. If that global source label is set as a Pod annotation then we observe all metrics

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
chancez: That matches my expectations

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,135 @@
...
+kind: ConfigMap
+metadata:
+    name: hubble-metrics-config
+    namespace: kube-system
+data:
+    hubblemetrics.yaml: |
+        hubbleMetrics:
+            enabled: true
+            metrics:
+            - name: flow
```
sgargan, line 71: Awesome to see this being addressed, its a major pain point right now and I was just asking about it on [slack](https://cilium.slack.com/archives/CQRL1EPAA/p1712692929740089). Happy to help if there's some way I can be useful.

One thing I would love to see would be different levels of granularity for each of the metrics (possibly appended as a suffix to the metric name) where each granularity could have it's own set of context options. This would avoid having to opt out completely from visibility into a source/destination via excludes as you could use labels with a coarser granularity (like app) to reduce the cardinality. The annotation driven suggestion below would make a lot of sense for this kind of scheme.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
sgargan: Big :+1: for using annotations to configure metrics. Long term this is going to be the most flexible. There are a number of other configuration options that make sense at this level and would be easy to opt into with annotations 
- unidirectional flows - right now flow metrics are captured for both directions in and out of a workload. For highly connected workloads it would be useful to potentially capture only metrics in one direction. This would reduce the cardinality by half while still maintaining most the visibility.
- verdict filtering - again for high cardinality workloads, it might be useful only to track metrics for given verdicts.  omitting forwarded verdicts and only shipping metrics for dropped verdicts that might be useful to have alerts on.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Proposal
+
+### Overview
+
+We propose a new config map wrapped functionality in Hubble metrics which can be used to manage the cardinality, dimensions and targets of the metrics. Drawing from [Hubble Flows dynamic exporter](https://docs.cilium.io/en/latest/observability/hubble-exporter/) which uses a standalone config map to manage exporting of flows, we propose a similar mechanism. 
+
+Existing behavior of reading cilium-config for `hubble-metrics` be preserved for backward compatibility. 
+
+A new config called `enable-dynamic-hubble-metrics` be introduced which when set to true, will enable the dynamic management of Hubble metrics. When this config is set to true, the cilium agent will read the `hubble-metrics-config` config map to get the configuration for the metrics. When this config is set to false, the cilium agent will read the `cilium-config` for `hubble-metrics` configuration.
```
kaworu, line 45: typo
```suggestion
A new config called `enable-dynamic-hubble-metrics` will be introduced which when set to true, will enable the dynamic management of Hubble metrics. When this config is set to true, the cilium agent will read the `hubble-metrics-config` config map to get the configuration for the metrics. When this config is set to false, the cilium agent will read the `cilium-config` for `hubble-metrics` configuration.
```

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Config Map
+
+A new config map called `hubble-metrics-config` will be introduced which will have the following fields:
+
+- `enabled`: Boolean value to enable/disable the dynamic management of Hubble metrics.
```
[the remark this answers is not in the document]
kaworu: @chancez While I agree conceptually, `disabled: false` is difficult to read and would be inconsistent with the Hubble flow exporter.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,135 @@
...
+                    - destination_pod: ["kube-system/"]
+            - name: dns
+                contextOptions: 
+                    - labelsContext=source_namespace,source_pod
+                    - destinationIngressContext=pod
+                includeFilters: 
+                    - source_pod: ["namespace1/"]
+```
+
+This will make sure we will have a dynamic way to manage each metric's cardinality, dimensions and targets. The flow filters can  be applied on each metric's `ProcessFlow` function save CPU cyles.
```
kaworu, line 88: IMO implementation details should be discussed in the follow up PRs rather than at the design stage.

```suggestion
This will make sure we will have a dynamic way to manage each metric's cardinality, dimensions and targets.
```

kaworu (changes requested):
Hi @vakalapa and thanks for the PR!

I left a couple of minor comments, and have two high level questions:
1. Did you consider a CRD instead of a ConfigMap to configure metrics? [The question was discussed for the Hubble exporter](https://github.com/cilium/cilium/pull/28220#issuecomment-1741109225), and I feel like an explanation of why the CFP propose one or the other (or something else) would make sense in the "Implementation Alternatives" section.
2. It's unclear to me whether the CFP plan to deprecate/remove the cilium's config `hubble-metrics`. IMO we don't want to maintain the two long-term, and thus we're missing a migration path (or at least suggestion) from the current `hubble-metrics` mechanism to the proposed `hubble-metrics-config` mechanism in the "Impacts / Key Questions". Are we going to ship / document `hubble-metrics-config` for the existing dashboards?

The second question goes back to the first where if the new system is CRD driven we could have a CRD per dashboard in addition to user defined metrics CRD, although maybe we would generate too many metrics and be back to square one, hence the request to clarify in the CFP.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
vakalapa: @sgargan if i understand you correctly, is your scenario to have multiple granularities for a given metrics based on the pod combinations ? Or different metrics with different context options (this is supported). 


Example:

```
 - name: flow
    -   contextOptions: 
            - sourceEgressContext=pod
            - destinationIngressContext=pod
        includeFilters: 
            - [namespace1/]
  - contextOptions:
            - sourceEgressContext=workload
            - destinationIngressContext=workload
        includeFilters: 
            - [test/]
 
```

If it is the former of having same metric like (as shown above) `flow` generating different labels based on a combination of contextOption + filter then I will defer the possibility to @chancez  or whoever knows Hubble code in depth, if this is possible and is a desired scenario ?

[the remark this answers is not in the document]
vakalapa: this is interesting like a TTL of a metrics and then delete the metrics after a while ? I would welcome this change, but it might be our of scope for this CFP. it would be a all new enhancement for Hubble metrics to "track" dimensions and delete dimensions after a certain timeout. Happyt o call this out in the CFP as a potential future expansion ?

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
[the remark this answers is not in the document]
vakalapa: Proposal was to add a new config on the cilium agent (via cilium-config). Added this clarity on the doc.

vakalapa:
> > @kaworu: @chancez While I agree conceptually, disabled: false is difficult to read and would be inconsistent with the Hubble flow exporter.
> 
> Fair enough, so then should we just remove the option from the dynamic metrics config entirely? Then to disable a given metric option it just needs to not be configure in the list all? Eg to have the feature enabled, but no metrics configured:
> 
> ```
>        hubbleMetrics:
>             metrics: [] # no metrics enabled currently. 
> ```

That sounds reasonable, check for the presence of this new configmap and/or the lenght of metrics to determine if the feature is enabled. Removing the enabled/disabled section from CFP.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
vakalapa: Resolving this comment, since this is following previous precedence.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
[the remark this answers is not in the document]
vakalapa: Removed the enabled/disabled options, we will go via the metrics list defined. No metrics defined == disabled.

vakalapa:
@kaworu  Thank you for the review, 

> 1. Did you consider a CRD instead of a ConfigMap to configure metrics? [The question was discussed for the Hubble exporter](https://github.com/cilium/cilium/pull/28220#issuecomment-1741109225), and I feel like an explanation of why the CFP propose one or the other (or something else) would make sense in the "Implementation Alternatives" section.

Ack, i want building on the previous precedence of config maps, but open to discussion on this. I believe configmaps are better solution compared to CRDs as lined out in @chancez 's comment. Added my views to alternatives section. 

> 2. It's unclear to me whether the CFP plan to deprecate/remove the cilium's config `hubble-metrics`. IMO we don't want to maintain the two long-term, and thus we're missing a migration path (or at least suggestion) from the current `hubble-metrics` mechanism to the proposed `hubble-metrics-config` mechanism in the "Impacts / Key Questions". Are we going to ship / document `hubble-metrics-config` for the existing dashboards?

That is a great point, we should deprecate the older path, but i am not familiar about the deprecation policy/process of cilium? If there are any pointers on how this is normally handled, i am happy to add it to "Future Milestones". My assumption is that it will take atleast one or two minor version upgrades for existing behavior get deprecated. I.e. if this support lands in 1.16 cilium, by 1.18 existing configuration via cilium-config will be deprecated?

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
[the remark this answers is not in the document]
chancez: Yeah I think managing metrics deletion is a separate feature that's more nuanced and separate from the configuration of the metrics. I believe the config as proposed should support it; if it does need some kind of configuration.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,151 @@
...
+metadata:
+    name: hubble-metrics-config
+    namespace: kube-system
+data:
+    hubblemetrics.yaml: |
+        hubbleMetrics:
+            metrics:
+            - name: flow
+                contextOptions: 
+                    - sourceEgressContext=pod
```
chancez, line 71: I think the values themselves for the options should be their own values too. 

```
- name: flow
  contextOptions:
    - name: sourceEgressContext
       values: 
         - pod

- name: dns
  contextOptions:
    - name: labelsContext
       values:
         - source_namespace
         - source_pod
```

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
chancez: I'm not entirely sure if it's a desired scenario, but with the current proposal and the fact that we're introducing structure to the `contextOptions`, it should be possible to these fields/behavior later, if it makes sense. I don't think we should block on adding new functionality unless the current proposal prohibits the possibility of adding it later, so I think we're fine here.

chancez (approved):
CFP looks good to me. I'm pinging others to see what else is needed before merging.

cilium/CFP-30788-managing-hubble-metrics-cardinality.md
```diff
@@ -0,0 +1,160 @@
...
+CRDs are good alternative to approach solving this problem, there was good amount of discussion done previously [here](https://github.com/cilium/cilium/pull/28220#issuecomment-1741109225). For this proposal I built upon the existing precendence so there is uniformity on how Hubble is configured for dynamic flow logs and metrics.
+
+#### Pros
+
+* Versioned and changes are preserved. 
+* Can impliment complex validation logic
+
+#### Cons
+
+* Unlike ConfigMaps, CRDs are heavy-weight. In real-world scenarios, my assumption is that the dynamic hubble metrics config map will not be changed oten, so CRD based continous watcher might be overkill.
```
kaworu, line 129: typo
```suggestion
* Unlike ConfigMaps, CRDs are heavy-weight. In real-world scenarios, my assumption is that the dynamic hubble metrics config map will not be changed often, so CRD based continous watcher might be overkill.
```

kaworu (approved):
Thanks for the update @vakalapa!

> My assumption is that it will take at least one or two minor version upgrades for existing behavior get deprecated. I.e. if this support lands in 1.16 cilium, by 1.18 existing configuration via cilium-config will be deprecated?

This is a fair assumption, I think  a couple of version should be enough to validate the configMap approach and implementation, and at that point we could start on deprecating the cilium-config metrics configuration.

vakalapa:
> This is a fair assumption, I think a couple of version should be enough to validate the configMap approach and implementation, and at that point we could start on deprecating the cilium-config metrics configuration.

thank for the review, I added the same at the bottom of the proposal stating, we will re-visit deprecation in 2 versions' time.

xmulligan:
Last thing before merging, can you create and put this in a /hubble folder rather than the /cilium folder?

chancez:
Good idea. @xmulligan @vakalapa once it's moved I'll merge the CFP.

vakalapa:
Done, moved CFP to hubble folder. @chancez

joestringer:
Heads up that I moved this under `cilium/` to match the documented process: https://github.com/cilium/design-cfps/pull/35 .

joestringer:
Ah @xmulligan I didn't see your request there. If we're going to split PRs up like that, then it'll make sense to update the README.md to describe exactly how we expect developers to make those decisions so we can consistently apply the rule :)

xmulligan:
ahh you are right sorry about that. I guess I expected the issue to be in the Hubble repo. Thanks for changing

PR #31, opened 2024-04-04
Add CFP-31752 Active Connection Tracking

AwesomePatrol:


cilium/CFP-31752-active-connection-tracking.md
```diff
@@ -0,0 +1,200 @@
...
+
+* Some events may be missed by datapath so the information must be combined with
+  Agent's (i.e. counting service entries that were GCed)
+* `uint32` will overflow, so metrics reader will need to adjust for that
+
+### Alternatives
+
+* Use event-based mechanism (for example: trace monitoring / Hubble) -
+  unreliable, too slow
+* Dump conntrack table - too slow
```
christarazi, line 106: Have you considered batch ops? I assume the "too slow" comment here is because it is slow to dump the map element by element. 

The CT BPF map pressure metric uses batch ops to dump the map and count the elements for the metric already.

AwesomePatrol: This looks promising. I need to explore it further:
1. In the initial testing it shows:
    ```
    48791 entries -> 36.65ms
    48803 entries -> 36.38ms
    ```
    which gives ~0.7µs per entry which completely satisfies the scalability requirement.
2. Does this warning apply? https://github.com/cilium/ebpf/blob/main/map.go#L980-L982
3. Are there any other downsides (i.e. am I guaranteed to get the full state? do I get only current/valid entries?)

xmulligan:
@AwesomePatrol are you still working on this CFP? We now have a few different statuses that CFPs can fall into https://github.com/cilium/design-cfps#status

AwesomePatrol:
> @AwesomePatrol are you still working on this CFP? We now have a few different statuses that CFPs can fall into https://github.com/cilium/design-cfps#status

Implementation is complete. So this CFP can be merged or dropped. It is no longer needed.

cilium/CFP-31752-active-connection-tracking.md
```diff
@@ -0,0 +1,200 @@
+# CFP-31752: Active Connection Tracking
+
+**SIG: SIG-DATAPATH**
+
+**Begin Design Discussion:** 2024-03-13
+
+**Cilium Release:** 1.16
+
+**Authors:** Aleksander Mistewicz <amistewicz@google.com>, Rob Scott <robertjscott@google.com>
```
xmulligan, line 9:
```suggestion
**Authors:** Aleksander Mistewicz <amistewicz@google.com>, Rob Scott <robertjscott@google.com>

**Status:** Released Cilium 1.16
```

xmulligan:
Would you mind making any updates that you think are needed so we can then merge it as "released"?

cilium/CFP-31752-active-connection-tracking.md
AwesomePatrol: Only the first two PR are part of v1.16 release. The other two - surfacing metrics in prometheus and counting connection errors - are only on main. Do you want to backport them to v1.16 or should it say `Released Cilium 1.17`?

xmulligan: You're right. Updated to 1.17

pchaigno (approved):

pchaigno:
The status update in the CFP was added, addressing the last review. Merging.

PR #33, opened 2024-04-11
CFP-033: Move cilium/hubble code into cilium/cilium repo

michi-covalent:


chancez:
Overall, looks good. I think we may want to elaborate slightly more on phase 2, and if there's anything else to be done after that. 

For example:
- Do we continue publishing to cilium/hubble forever?
- Do we start releasing CLI binaries to cilium/cilium at any point?

Basically, you outlined the improvements we want to make as part of the goals, but the goals don't actually include the final desired state.

michi-covalent:
phase 1 in cilium/cilium repo ready for review: https://github.com/cilium/cilium/pull/31893
 
> Do we continue publishing to cilium/hubble forever?
> Do we start releasing CLI binaries to cilium/cilium at any point?

i was thinking we'll continue using cilium/hubble repo to publish releases and host existing docs.

michi-covalent:
btw thank you chance for adding these reviewers. i need an approval from a committer before this CFP can be merged.

cilium/CFP-033-move-hubble-into-cilium.md
```diff
@@ -0,0 +1,120 @@
...
+
+* Enable developers to make Hubble server and CLI changes in a single pull
+  request, and immediately make the new Hubble CLI available in the CI
+  `cilium-agent` images without having to wait for the next Hubble CLI release.
+* Build Hubble CLI release binaries from Cilium release tags to simplify the
+  release process.
+
+## Non-Goals
+
+* Archive or delete the cilium/hubble repo.
```
gandro, line 75: Would it be fair to say that it is a non-goal of this CFP to change the versioning schema of Hubble? 

Because I assume this is probably the potentially biggest point of discussion, how we want to version and maintain the Hubble CLI version after Phase II.

cilium/CFP-033-move-hubble-into-cilium.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Phase 2: Update the release process
+
+In the second phase, coordinate with `@cilium/sig-hubble` and update the Hubble
+CLI release process. Instead of having a separate GitHub workflow in
+`cilium/hubble` repository, pushing a release tag in `cilium/cilium` repo can
+trigger a workflow to build Hubble CLI release binaries and publish a release
+in `cilium/hubble` repo.
+
+## Impacts / Key Questions
```
gandro, line 98: I think another impact of Phase I is that users importing `cilium/hubble` as a go module will now have to vendor in Cilium. Impact seems rather low, mainly affecting cilium/cilium-cli, cilium/fake, cilium/hubble-ui and the already archived hubble-otel:

https://pkg.go.dev/github.com/cilium/hubble@v0.13.2/pkg/printer?tab=importedby
https://pkg.go.dev/github.com/cilium/hubble@v0.13.2/pkg/time?tab=importedby
https://pkg.go.dev/github.com/cilium/hubble@v0.13.2/pkg/defaults?tab=importedby

The only notable known import which is not a fork and not a Cilium org member is https://github.com/stleox/seeflow

kaworu (approved):
Thanks @michi-covalent LGTM.

To make sure I understand: In Phase 1 the Hubble CLI in the Cilium image is built as part of the Cilium release process, and then we manually do a Hubble CLI release to be published at `cilium/hubble` for download (i.e. all platforms). Phase 2 is about automating the later. Does that sound correct?

gandro (commented):
Thank you! Excellent write-up and the two-phase plan allows us to take it one step at a time. A bit of feedback attached inline.

cilium/CFP-033-move-hubble-into-cilium.md
gandro: Another impacted user is the Homebrew recipe: https://github.com/Homebrew/homebrew-core/blob/master/Formula/h/hubble.rb

chancez:
> i was thinking we'll continue using cilium/hubble repo to publish releases and host existing docs.

So forever/indefinitely? Let's make that explicit in the doc. 

And what about publishing to cilium/cilium? Is the plan to not publish to c/c, ever?

michi-covalent:
> To make sure I understand: In Phase 1 the Hubble CLI in the Cilium image is built as part of the Cilium release process, and then we manually do a Hubble CLI release to be published at cilium/hubble for download (i.e. all platforms). Phase 2 is about automating the later. Does that sound correct?

yeah that's what i'm thinking. let me make these 2 phases a bit clearer in the CFP.

> So forever/indefinitely? Let's make that explicit in the doc.

yeah let me update the docs. btw i'm not against publishing releases in cilium/cilium instead of cilium/hubble at some point. personally i haven't found any strong reasons to make the switch, but i am kind of curious what other sig-hubble folks think 💭

cilium/CFP-033-move-hubble-into-cilium.md
michi-covalent: actually i was thinking the versioning will change as a part of the CFP, basically use the same version as cilium version. there is a section called `Question: How do you version Hubble CLI releases in Phase 1?` below. let me know what you think 💭

cilium/CFP-033-move-hubble-into-cilium.md
michi-covalent: yeah i was just thinking of mentioning the homebrew stuff. i'll update the CFP 📝

michi-covalent:
thank you all i updated the CFP based on your feedback. i'll wait until the end of this week before merging the PR in case people have more feedback / questions.

cilium/CFP-033-move-hubble-into-cilium.md
gandro: Ah, right. I overlooked that part, it implies we just use the Cilium version. However, I prefer to have a section about versioning and supported versions, since I believe this does have user impact that I believe we want to make sure we agree on the the path forward. Namely:

- Hubble CLI versions will now be coupled to Cilium versions, i.e. Hubble CLI will jump from 0.13.x to 1.16.y

What I think is less clear to me:

Also, since we would match the Cilium version, what will the maintained versions of the Hubble CLI be? Currently, Hubble only supports one stable branch, where as Cilium maintains three stable branches. Will we release Hubble CLI artifacts for all three stables branches? Do we only release artifacts for the latest stable branch? If we do the latter (i.e. maintain only latest stable), there is going to be the oddity where the "hubble" binary inside the cilium-agent pod effectively is going to be a maintained build of old-stable, i.e. there will be a Hubble CLI 1.16.20 binary inside the cilium-agent 1.16.20 docker image, where as the latest Hubble CLI release on GitHub will be 1.16.8 (or whatever was the latest when we released Hubble CLI v1.17.0). Not necessary a problem, but something to be aware of that this might confuse users.

Another alternative would be to use separate versions for the Hubble CLI inside the cilium-agent pod, and keep having 0.x.y versions for the release artifacts on cilium/hubble (at least feasible during Phase I)

Again, I'm fine to punt those questions to a separate CFP, but I believe we want to have agreement on that at some point, since I do think there is a non-trival decision space here. IMO, we don't have to change the versioning for Phase I, but we probably should change it for Phase II.

michi-covalent: yeah let me add a section regarding versioning. i have my thoughts 💭

michi-covalent: updated the CFP. please take a look 🚀🙏

gandro: Thanks, I think that answers all my questions :pray:

chancez (approved):

PR #35, opened 2024-04-30
Rearrange CFPs to match documented process

joestringer:
A couple of the CFPs were filed in ways that don't follow the current
instructions. Move them to make the repo more consistent.


michi-covalent (approved):
oh yeah i missed the step 1. thanks for fixing it 🥰

PR #40, opened 2024-07-08
Ordered policy engine.

squeed:
Note: There is a HackMD for this PR with the diagrams fully rendered as well as allowing interactive editing: https://hackmd.io/@squeed/ordered-policy/edit

### DIFF
--- a/cilium/CFP-23380-ordered-policy.md
+++ b/cilium/CFP-23380-ordered-policy.md
@@ -1,0 +1,329 @@
+# CFP-23380: Ordered network policy
+
+**SIG: sig-policy**
+
+**Begin Design Discussion:** 2024-07-01
+
+**Cilium Release:** 1.17
+
+**Authors:** Casey Callendrello <cdc@isovalent.com>, You, and you too.
+
+## Summary
+
+Add explicit, user-supplied rule ordering to the policy engine.
+
+## Motivation
+
+Most firewall engines present an _iterative_ interface to their users. Users provide an _ordered_ list of rules. In comparison, Cilium policy is _commutative_, which is to say, the ordering of rules does not matter.
+
+Iterative policy languages are more expressive. For example, consider the rules:
+
+> 1. `env=qa, role=grafana`: allow access to `env=prod, role=artifacts`
+> 2. `env=qa`: deny access to `env=prod`
+
+These sorts of rules are awkward or impossible to express via the Cilium policy language, where Deny-based rules always take priority.
+
+## Goals
+
+* Allow expressing L3/L4 rules with order
+* Add an explicit priority field to policy Rules
+* Keep the datapath O(1)
+
+## Non-Goals
+
+* Add an ordered engine to the datapath
+* Add an ordered engine to Envoy / L7 proxies
+
+## Existing mechanism
+
+Before outlining the proposed changes, it is important to understand the current policy resolution mechanism in detail. The basic shape:
+
+1. From API objects, policy is lightly translated to an `[]api.Rule`
+2. These rules are converted to a `L4Policy`, which is a `map[(proto,ports)]L4Filter`. The `L4Filter` is a `map[selector]PerSelectorPolicy`. `L4Filter` contains some bookkeeping information. The "policy judgement" is the `PerSelectorPolicy`. Informally, this can be considered a `map[(proto, port(s), selector)]PerSelectorPolicy`.
+3. The `L4Policy` is then converted to a `MapState`, which is (logically) a `map[port+proto]map[identity]MapStateEntry`. The `MapStateEntry` is the actual configuration for the datapath.
+
+
+### Existing logic: PerSelectorPolicy generation
+
+When there are no existing entries, the translation from an `api.Rule` to its `L4Policy` and `PerSelectorPolicy` is straightforward. Most of the logic is in how to merge two `PerSelectorPolicies`. The `PerSelectorPolicy` merge logic primarily evaluates per-field precedence. For example, deny overrides all, auth overrides no-auth, L7 overrides no L7. In the event of a conflict (e.g. different L7 parsers for the same port), policy resolution fails. L7 rules are the exception: they are vectors that support merging.
+
+Envoy directly consumes the `L4Policy`, the distilled per-selector policy. However, the BPF policy engine expects a different structure: the PolicyMap.
+
+### Existing logic: Generating the PolicyMap
+
+The policy map as stored in BPF is a map from `(identity, proto, port) -> MapStateEntry`. Thus, the `L4Policy` must be partially inverted to a per-identity policy result. This translation is not purely mechanical; it must also be precedence aware.
+
+Why must PolicyMap generation be precedence aware? After all, `PerSelectorPolicy` merging *should* have included all the precedence logic. However, disjoint selectors may select the same identity. Thus, when resolving the `MapStateEntry` for a `(id, proto, port)` key, it may be that a tiebreaker is needed between two `MapStateEntries`.
+
+## Proposal
+
+This proposal outlines a way to add explicit order-based precedence to L3/L4 policies. This proposal explicitly does not permit L7 policies with a specified order. See below for how ordering may interact with L7 rules.
+
+There are two key sections for this proposal. The first is the technical implementation of policy ordering. The second is an outline of the algorithm for generating an endpoint-specific BPF PolicyMap from an ordered set of policies.
+
+### Ordered PerSelectorPolicy resolution
+
+Quick summary:
+- Generate a partial order of PerSelectorPolicies, based on specified order and tie-breakers.
+- Merge PerSelectorPolicies with the same order number, using existing precedence logic.
+
+The first step is to generate an ordered list of PerSecectorPolicies. This should be sorted according to explicit order, then tie-broken according to the existing precedence logic.
+
+Then, PerSelectorPolicies with the same order should be *merged*. This means there will be exactly one PerSelectorPolicy for the key `(proto, port, selector, priority)`. This prevents conflicts while determining L7 policies, and preserves existing behavior, where each policy can be considered to have priority 0.
+
+Finally, the full set of PerSelectorPolicies is flattened, retaining the first PerSelectorPolicy per `(proto, port(s), selector)` key.
+
+
+### Generating the MapState
+
+Quick summary:
+- Apply the ordered list of PerSelectorPolicies to the MapState
+- Resolve conflicts, preferring existing entries
+
+Once the ordered, flattened set of PerSelectorPolicies has been created, it must be translated to the MapState. Recall that the MapState is almost equivalent to `map[proto + port]map[identity]MapStateEntry`, where `MapStateEntry` is
+
+```go
+ProxyPort uint16
+Listener string
+AuthType AuthType
+IsDeny bool
+```
+
+This step must *also* be priority aware, as **selectors can overlap**. 
+
+The sketch of the algorithim is:
+- For every PerSelectorPolicy, highest priority first:
+    - For every selected identity:
+        - Insert in to the MapState *if this would not affect an existing entry*
+        - Propagate authentication mode to descendants
+
+The MapState *isn't* a map, however. It is a longest-prefix-match trie, whose key is chosen to match the datapath. In other words, MapState's precedence matches the datapath's precedence. A longer-prefix MapState key will be preferred over a shorter prefix match.
+
+This step must allow for deconfliction, to cover two cases:
+1. A higher-priority PerSelectorPolicy has been inserted
+2. An overlapping PerSelectorPolicy has been inserted
+
+### MapState conflict resolution
+
+The key observation is this: **we cannot insert a lower-priority, higher-precedence entry in to the MapState**. Since MapState precedence mirrors datapath precedence, determining this is relatively straightforward: to insert a given key, we scan all ancestor keys, checking if the same identity is selected. If a lower-precedence entry is found, we cannot insert the higher-precedence, lower-priority entry.
+
+#### Example
+
+A MapState contains the entry `(tcp, port 0-1023, identity *)`.
+
+**Example 1a:** Can we insert `(tcp, port 0-1023, identity 5)`? No, because this would cover a parent key.
+
+```mermaid
+---
+title: "Example 1a: invalid"
+---
+graph TD
+    rule1[prio 1, tcp, 0-1023, id *]
+    rule2[prio 2, tcp, 0-1023, id 5]
+    root[[root]]
+    
+    root-->rule1
+    rule1-. invalid .-> rule2
+```
+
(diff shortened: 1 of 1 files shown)


cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+
+For the time being, we will not permit L7 policies to specify an order. This is so that we do not need to implement an ordered L7-aware merge. It is likely this restriction will be lifted, once we gain familiarity with ordered policies.
+
+However, we still need to handle ordering between L4 and L7 policies.
+
+In the unordered case, any L7 policy overwrites a L4 policy. Consider two policies with the same selector:
+> - prio -1: allow access to `role=web, port 80` 
+> - prio 0: allow access to `role=web, port 80, host=example.com`
+
+In this case, the L7 policy would be ignored, since it is "overridden" by the L4 policy.
```
christarazi, line 316: What happens if the user supplies an explicit priority that inverts the core policy assumption of "any L7 policy overwrites an L4 policy"? In other words, what happens if the user does

```
> - prio 1: allow access to `role=web, port 80` 
> - prio 0: allow access to `role=web, port 80, host=example.com`
```

christarazi (commented):
Thanks for the write up. I've made a first pass at it, but haven't fully digested all of the little details (I will go back through this again), but I noticed that there seems to be an entire class of problem that isn't addressed in the CFP. How will we handle cases where the user supplies an explicit priority that goes against the assumptions of the policy model? See my comment below as an example. Another example would be if the user priorities a deny lower than an allow (for the same exact selector).

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+    root[[root]]
+    
+    root --> rule1
+    rule1 --> rule2 & rule1a
+    rule2 -. covers .-> rule1
+```
+
+#### Merging mutual authentication
+
+Whether or not a connection requires mututal authentication is *also* managed by the network policy engine. However, authentication is not a "traffic verdict". (Traffic verdicts are `allow`, `deny`, and `proxy`). Rather, it is applied to all traffic, regardless of selectors.
```
joestringer, line 230: nit: I won't go over them all but a spellchecker could pick up on some things like this. I also noticed some cases like `PerSecectorPolicies` that a spellchecker won't pick up on by default but would also be nice to fix.
```suggestion
Whether or not a connection requires mutual authentication is *also* managed by the network policy engine. However, authentication is not a "traffic verdict". (Traffic verdicts are `allow`, `deny`, and `proxy`). Rather, it is applied to all traffic, regardless of selectors.
```

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+
+There are two options for resolving mutual authentication. The first is to push auth down to all mapstate entries as part of resolving conflicts. The other is to maintain a parallel overlay selector trie for authentcation, which is then merged when generating the BPF entries. In order to keep the merge algorithm simpler, an overlay trie is our current proposal.
```
joestringer, line 232: Given the place where the auth field exists within the policy (ie applying to specific L3+L4 rule, where either of these fields may represent a wildcard), I don't understand the first proposal here. But sure if we can merge an overlay trie when turning these into map entries I can see how that could retain mutual auth semantics.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+
+We will need to ensure that SelectorCache updates are processed in a transactional manner -- that is to say, an incremental update must be delivered to all affected selectors *before* distributing this to the MapState.
+
+#### Example 1: no wildcards
+
+Consider a set of rules:
+
+1. tcp 0-1023, ID (5,6)
+2. tcp 80, ID (7)
+3. tcp *, ID (8)
```
joestringer, line 252: Are the numbers at the start of the lines here representing priority? This should be more explicit to describe the meaning of the ordered list.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+
+Incremental updates are aggregated by selectors and then inserted in to the MapState. The MapState is responsible for deconflicting overlapping selectors. This remains the same in the ordered version.
+
+The important change is that incremental updates must be inserted in to the map state *in priority order*. That is to say, higher-priority PerSelectorPolicies must apply their updates first. Then, the usual MapState deconfliction rules apply, and there is no risk of an incremental update diverging from policy intent.
+
+It is safe to insert incremental updates in this manner because identities and selectors are immutable. As a direct consequence, two selectors S1 and S2 might overlap, a new identity I1 can never change state from non-overlapping to overlapping.
+
+We will need to ensure that SelectorCache updates are processed in a transactional manner -- that is to say, an incremental update must be delivered to all affected selectors *before* distributing this to the MapState.
+
+#### Example 1: no wildcards
```
joestringer, line 246: What does `no wildcards` mean here given that rule 3 has `tcp *`?

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+    
+    root --> rule3
+    rule3 --> rule1
+    rule1 --> rule2
+    end
+    
+    subgraph After
+    direction TB
+    rule1a["prio 1, tcp, 0-1023, id {5, 6, 9}"]
+    rule2a["prio 2, tcp, 80, id {7, 9}"]
```
joestringer, line 279: The text above says that:

> 2. tcp 80, ID (7)
...
> 2. Rule 2 checks ID 9. ID 9 is present in an ancestor rule. ID 9 is **not added**.

And yet the diagram has id 9 included here. Why?

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+#### Example 2: Wildcards
+
+
+## Impacts / Key Questions
+
+### KQ1: How should unordered and ordered policies interfact?
+
+To preserve existing behavior, all policies without an explicit order will have order number 0. Since policies with the same numerical order are merged, this preserves existing behavior.
+
+Only CCNPs are allowed to have a negative order number. This prevents namespace administrators from overriding cluster-wide policies.
```
joestringer, line 303: Expanding on the feedback earlier about defining the meaning of the values, this should probably be a bit more explicit to state that negative priority values take precedence over positive, and the more negative the number, the higher the precedence.

joestringer: Separately though this also raises the question about how CCNP and CNP rules combine. Currently they're all at the same precedence, then when combined with the deny taking precedence over allow, means that administrators can either include policies that allow common infra for everyone (eg allow DNS), or deny traffic that no-one should have access to (forbidden IP ranges for instance).

I feel like the right expectation is that administrators can write CCNP policies without worrying about the precedence vs. CNP, and they will know that the CCNP always applies. Whether we implement this by forbidding CCNP from configuring precedence, or forbid precedence values in the ranges that CNP configures, or define that the priority numbers are distinct spaces between the two so that CCNP always applies (all CCNP priority is higher than all CNP priority), or some other option... not sure.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+For the time being, we will not permit L7 policies to specify an order. This is so that we do not need to implement an ordered L7-aware merge. It is likely this restriction will be lifted, once we gain familiarity with ordered policies.
+
+However, we still need to handle ordering between L4 and L7 policies.
+
+In the unordered case, any L7 policy overwrites a L4 policy. Consider two policies with the same selector:
+> - prio -1: allow access to `role=web, port 80` 
+> - prio 0: allow access to `role=web, port 80, host=example.com`
+
+In this case, the L7 policy would be ignored, since it is "overridden" by the L4 policy.
+
```
joestringer, line 317: I'd like to see a dedicated section discussing deny policies given the way they currently take precedence over everything, and this is critical to ensuring they work as expected. The answer might be as simple as matching the L7 approach to say that priority is not allowed in conjunction with deny policy.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+
```
joestringer, line 318: Impact: It's probably fair to say that if we enable users to configure a priority value for their policies, then there's a decent chance that users will start configuring an order of magnitude more policies at the same time. Not sure whether there's anything particular we want to do about this (maybe more testing to validate performance), but might be worth calling out.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+## Impacts / Key Questions
+
+### KQ1: How should unordered and ordered policies interfact?
+
+To preserve existing behavior, all policies without an explicit order will have order number 0. Since policies with the same numerical order are merged, this preserves existing behavior.
+
+Only CCNPs are allowed to have a negative order number. This prevents namespace administrators from overriding cluster-wide policies.
+
```
joestringer, line 304: Not necessarily an immediate question here, but related to priorities.. upstream k8s AdminNetworkPolicy is introducing priorities as well, so at some point we'll want to define how the priority spaces interact between the space here and the space in ANP. Do the numbers mean the same thing? So you could intermix priorities between CNP, ANP, etc.? Might be worth noting down any thoughts we might have about this. Not sure it directly impacts this CFP, could maybe put it into future milestones.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+
```
joestringer, line 305: Might want to document the impact on `enableDefaultDeny: false`. I don't think it's an issue - the same principle that "all policies applying to the endpoint must set this to false to avoid generating the default deny", so priority doesn't interact at all. But the question could come up about how these two interact so it'd be nice to clearly define how it's not an issue.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+## Motivation
+
+Most firewall engines present an _iterative_ interface to their users. Users provide an _ordered_ list of rules. In comparison, Cilium policy is _commutative_, which is to say, the ordering of rules does not matter.
+
+Iterative policy languages are more expressive. For example, consider the rules:
+
+> 1. `env=qa, role=grafana`: allow access to `env=prod, role=artifacts`
+> 2. `env=qa`: deny access to `env=prod`
+
```
joestringer, line 23: I don't know if there's a good place to discuss this, but IMO cloud-native has substituted iteration for label selectors to express selection. Combining both is significantly complicating the solution over traditional iterative approaches, because you _can't_ just look at the priorities, you still have to know how the selectors apply to workloads based on their labels. This is part of what makes implementing this proposal so tricky, and what will make debugging environments with this feature also very tricky.

joestringer: Come to think of it, we should probably highlight that debuggability impact in its own impact section.

nathanjsweet (approved):

squeed:
I still need to respond to Joe's feedback -- this isn't ready to go just yet.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+There are two key sections for this proposal. The first is the technical implementation of policy ordering. The second is an outline of the algorithm for generating an endpoint-specific BPF PolicyMap from an ordered set of policies.
+
+### Ordered PerSelectorPolicy resolution
+
+Quick summary:
+- Generate a partial order of PerSelectorPolicies, based on specified order and tie-breakers.
+- Merge PerSelectorPolicies with the same order number, using existing precedence logic.
+
+The first step is to generate an ordered list of PerSecectorPolicies. This should be sorted according to explicit order, then tie-broken according to the existing precedence logic.
```
bimmlerd, line 70: typo:
```suggestion
The first step is to generate an ordered list of PerSelectorPolicies. This should be sorted according to explicit order, then tie-broken according to the existing precedence logic.
```

When you say that these should be tie-broken according to the existing precedence order, I'd expect there to no longer be ties afterward. However, in the very next sentence you refer to policies with the same order. Does this mean that the tie-breaking process is part of the merging process? I think this could be clarified a little.

cilium/CFP-23380-ordered-policy.md
bimmlerd: > lower precedence 

I'm also getting confused about priority and precedence. In my mind, priority was the explicit order number in the policy, whereas precedence was ~proportional to the specificity of the rule. Hence I'd have agreed with Jarno's comment _except_ for wording it as 'lower priority' not 'lower precendence'

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,329 @@
...
+
+Then, PerSelectorPolicies with the same order should be *merged*. This means there will be exactly one PerSelectorPolicy for the key `(proto, port, selector, priority)`. This prevents conflicts while determining L7 policies, and preserves existing behavior, where each policy can be considered to have priority 0.
```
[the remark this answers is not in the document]
jrajahalme: It could be that insertion in sorted order would reduce churn on MapState, especially if we could deduce that the new entry is not needed without allocating it first.

cilium/CFP-23380-ordered-policy.md
joestringer: I understand the expressibility aspect of the example above, what I'm worried about is that this is just not how we designed denies to work, so if we're not careful we could break existing users policies. Even if we don't go so far as to break existing policies, we could introduce surprising behavior for users where they expect denies (particularly with CCNP) to work one way and it turns out that we changed the rules and they work a different way. The [public docs on the policy language](https://docs.cilium.io/en/stable/security/policy/language/#deny-policies) currently define that deny policies always deny regardless of allow rules. Of course those docs are subject to change as long as we have backwards compatibility in mind. 

This ties also into the other [thread above about CCNP vs CNP priority](https://github.com/cilium/design-cfps/pull/40#discussion_r1676505326), because currently an administrator can create a CCNP deny-all, and there is no way for an app developer in a specific namespace to override that deny. Maybe this is more of a problem for the CCNP + CNP combination model when also introducing priority

xmulligan:
@squeed we now have statuses for CFPs https://github.com/cilium/design-cfps#status. Please add the one that you think fits here

cilium/CFP-23380-ordered-policy.md
squeed: That is the intention, yes. An L4 rule with higher priority would cause the L7 rule to be disregarded.

This CFP currently makes no statement as to the ordering of L7 rules. Logically, the expectation of

> - prio 0: allow access to host=example.com
> - prio 1: allow access to host=example.org

would be that example.com and example.org would be allowed. But that is not currently on the table. L7 policy must be priority 0.

PR #44, opened 2024-07-12
 CFP 33716:  Packet Tracing

Bigdelle:
Add PR for [CFP:  Packet Tracing](https://github.com/cilium/cilium/issues/33716).


gandro:
cc @cilium/sig-hubble

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,424 @@
...
+    // Remaining struct members
+    ...
+};
+```
+
+```go
+type Flow struct {
+    // Existing struct members
+    ...
+    IpTraceID uint64
```
chancez, line 157: A few thoughts:
- We already have a `uuid` for the Flow, and I wonder if we should consider using that.
- Since this is part of the IP packet, perhaps it should be part of the `IP` field instead?

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,424 @@
...
+
+## Goals
+- For tagged packets, add the ID to emitted flow events.
+- For tagged packets, emit datapath debugging events regardless of the debug-verbose setting.
+- Support both IPv4 and IPv6.
+- Add hubble filter for trace ID.
+- Allow users to define which flow IDs they want to monitor for events.
+
+## Non-goals
+- Automatically apply IP option to packets (see Future Milestones).
```
chancez, line 30: So when I read the title it mentions _distributed_ packet tracing, but this really just seems like "regular" packet tracing, since there's to trace ID propagation. Generally when I think of "distributed" tracing, I think of having observability beyond the initial source and destination. This doesn't tell you anything beyond the connection between the initial source and destination. It might be worth mention that somewhere, or rephrasing the title of the CFP to reflect this.

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,424 @@
...
+- **Optional:** Provide a flag to optionally remove the tag before egressing Node NIC, addressing targets that do not support IP options.
+- **Challenge:** This affects the available MTU.
+  - [Preferred] Only apply to packets that have the length available (e.g., packets with no data; SYN w/o fast open, RST, etc.).
+  - Modify the MTU/route MTU of the container (i.e., subtracting the IP option + trace ID lengths).
+
+### Deferred Milestone 3: Tag packets via CR
+- **Audience:** Network specialist with knowledge of network topology.
+- Develop a K8s API where the user specifies traffic to tag using common selectors: 4-tuple, pod identity, etc.
+
+## Appendix
```
chancez, line 386: It might be worth investigating or declaring how we would expect this to work with some of Cilium's other features:
- L7 policies/Envoy
- Egress gateway
-  IPSec/Wireguard.

It's okay to not know, but we should at least explicitly mention it's unsupported with certain other features that we could expect it to not work with, so we can document that if the CFP is accepted and implemented.

cilium/CFP-33716-distributed-packet-tracing.md
sypakine: Agreed, makes sense to drop the term distributed as this is only a part of what could be integrated into a larger system tracing system (e.g. with fabric integration). 

Although nested virtualized nodes (e.g. via kubevirt) may also capture flows in the path and then it becomes a bit greyer 😛

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: Creating a message type for the IpTraceID makes a lot of sense; having the IP option recorded would be helpful additional information. These traces are used to correlate flows, so I think creating a new message type and having a field in the Flow struct is an appropriate way to structure this. This has been updated in the new commits. If there are any additional thoughts, please let me know. If this answered your comment, please resolve the comment. Thank you!

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: This has been fixed and will be closed as resolved.

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: There is now a section addressing this. More investigation is ongoing, but this is what we expect based on current context/information. Please take a look and feel free to ask follow-ups/any clarifications.

Bigdelle:
cc @cilium/sig-datapath

julianwiedmann:
On a very high-level - you're proposing that the datapath matches some specific IP option in the packet (containing some `trace_id`), and transfers the `trace_id` value to userspace via the `trace_notify` struct ?

As an alternative thought - afaik the datapath is already punting some(?) packet headers to userspace. If you increase the number of punted bytes (so that it includes the IP options), would that allow you to move all the needed IP option-parsing into hubble?

Bigdelle:
> On a very high-level - you're proposing that the datapath matches some specific IP option in the packet (containing some `trace_id`), and transfers the `trace_id` value to userspace via the `trace_notify` struct ?
> 
> As an alternative thought - afaik the datapath is already punting some(?) packet headers to userspace. If you increase the number of punted bytes (so that it includes the IP options), would that allow you to move all the needed IP option-parsing into hubble?

We want the `ip_trace_id` available at the Kernel level. This PR sets up the infrastructure to expand on this feature as we continue to add other expansions; the next milestone involves restoring the `ip_trace_id` for reply packets using conntrack. This will ensure that subsequent packets in the same connection maintain their traceability even if they do not carry the ip_trace_id explicitly in their IP options. If we push the IP option parsing to Hubble, we lose this ability.

Also, by pushing this to Hubble, we lose integration with the cilium monitor for events. 

Is there a specific concern you have about including this in the trace_notify and drop_notify structs (these fields would only be populated if the feature is enabled and there is an embedded `ip_trace_id`?

Bigdelle:
> > On a very high-level - you're proposing that the datapath matches some specific IP option in the packet (containing some `trace_id`), and transfers the `trace_id` value to userspace via the `trace_notify` struct ?
> > As an alternative thought - afaik the datapath is already punting some(?) packet headers to userspace. If you increase the number of punted bytes (so that it includes the IP options), would that allow you to move all the needed IP option-parsing into hubble?
> 
> We want the `ip_trace_id` available at the Kernel level. This PR sets up the infrastructure to expand on this feature as we continue developing it; the next milestone (linked in the `Future Milestones` section of the CFP) involves restoring the `ip_trace_id` for reply packets using conntrack. This will ensure that subsequent packets in the same connection maintain their traceability even if they do not carry the `ip_trace_id` explicitly in their IP options. If we push the IP option parsing to Hubble, we lose this ability.
> 
> Also, by pushing this to Hubble, we lose integration with the cilium monitor for events.
> 
> Is there a specific concern you have about including this in the trace_notify and drop_notify structs? (these fields would only be populated if the feature is enabled and there is an embedded `ip_trace_id`).

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
+# CFP-2024-06-12: Packet Tracing
+
+## Meta
+**SIG:**
```
bowei, line 4: Put the sig.

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+**SIG:**  
+**Sharing:** Public  
+**Begin Design Discussion:** Jun 12, 2024  
+**End Design Discussion:**  
+**Cilium Release:**  
+**Authors:** Ben Bigdelle <bigdelle@google.com>, Mark St. John <markstjohn@google.com> based on the work of Nathan Perkins <nathanperkins@google.com>  
+**Reviewers:**  
+
+## Summary
+We aim to enhance cross-cluster and multi-system observability for flows that contain an `ip_trace_id` IP Option (referred to as a tagged flow), thus supporting tracing across networks. These tagged flows will trigger adding the ID value to Flow events messages and client (Hubble / Cilium Monitor) representation. This allows correlating flow events across systems which integrate with the span observability.
```
bowei, line 13: If this is a standardized IP option, please add a link to the relevant RFC where the option is defined.

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+
+Additionally, within Cilium, these flows will generate datapath debugging events regardless of the global debug-verbose setting. This will improve debuggability of production systems.
+
+## Motivation
+There are currently limitations for gathering and correlating flow events:
+- NAT transformations and DSR can all change the 4-tuple, so that a single Hubble or tcpdump filter cannot be used to trace the packet along the entire path – even when using a system such as Hubble Relay.
+- Enabling verbose datapath debugging is not possible in production environments due to traffic volume which leads to dropping flows.
+
+## Goals
+- For tagged packets, add the ID to emitted flow events.
```
bowei, line 23: Somehow this is less clear for me, suggest:

- Enable dataplane to watch for `ip_trace_id` option tagged packets and emit flow events for tagged packets.

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+- For tagged packets, emit datapath debugging events regardless of the debug-verbose setting.
```
bowei, line 24: - `ip_trace_id` can be used without enabling debug-verbose in the data path.

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+1. **Avoid MTU Concerns:**
+   - Automatically tagging packets can cause them to exceed the MTU, leading to fragmentation and performance issues.
+
+2. **Prevent Performance Loss from GRO:**
+   - GRO engine does not support packets with IP options.
+
+3. **Avoid Packet Loss:**
+   - Some network devices and paths may not support IP options, causing tagged packets in the path to be dropped.
+   - Targets that don't support IP Options:
+      - L4 ELB
```
bowei, line 60: Is this for a specific cloud provider? Or is this with the Cilium implementation of ELB support? Would be good to clarify here.

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+
+### Section 1: Packet Tagging 
+### Packet Tagging Responsibility
+
+This feature does not implement the packet tagging (see Deferred Milestone 1: Tag Packets). Instead, it is the client's responsibility to tag packets via IP option embedding. Therefore, the key assumption for this feature is that tagged packets arrive at ingress points with an `ip_trace_id` already embedded in the IP options.
+
+#### Reasons for Not Implementing Automatic Packet Tagging:
+
+1. **Avoid MTU Concerns:**
+   - Automatically tagging packets can cause them to exceed the MTU, leading to fragmentation and performance issues.
```
bowei, line 52: For an item like this, it would be good to give a reason of {mitigation || why this is not a problem}, etc. so the reader can understand the thought process here.

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+
+2. **Prevent Performance Loss from GRO:**
+   - GRO engine does not support packets with IP options.
```
bowei, line 55: Same here. Example:

"... This should not be an issue as this is a manually triggered debug option on selected flows generated by the user and the user can control the amount of performance degradation by selectively tagging packets as needed for debugging."

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+      - L4 ELB
+   - Targets that support IP Options:
+      - L4 ILB
+      - Services
+      - Pod-to-pod (including across VPC)
+
+4. **IP Checksum Recalculation:**
+   - When IP options are added or modified, the IP checksum must be recalculated, adding additional processing overhead and complexity.
+
+By making packet tagging the client's responsibility, we ensure that Cilium remains unaffected by these concerns, leaving it to the clients to incorporate these key considerations into their use of this feature and decide how best to utilize it to suit their specific needs.
```
bowei, line 69: It would be good to lift this to be a security section by itself, just so people reading can't miss the context here.

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+
+### Section 2: Feature Enablement
+`--enable-ip-option-tracing=<type>` to define the IP option type used to embed the trace ID value
+- `<type>` corresponds to the IP option value ([https://www.iana.org/assignments/ip-parameters/ip-parameters.xhtml](https://www.iana.org/assignments/ip-parameters/ip-parameters.xhtml))
```
bowei, line 73: What is the recommended value for `type`? Is this a user selected number?

I went to the IANA page and couldn't find a reference to "trace". Which value should be set here?

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+```
+
+### Section 3: Parsing
+
+Parsing at the ingress point of the TC hook allows for early detection and tracing of a flow.
+
+```c
+#ifdef ENABLE_PACKET_IP_TRACING
+    check_and_store_ip_trace_id(ctx, ENABLE_PACKET_IP_TRACING);
+    // Do parsing and storing of ip_trace_id
```
bowei, line 95: nit: this comment probably be before the `check_and_store_ip...()`

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+#endif
+```
+```c
+// Function to parse trace ID and store it if valid
+static __always_inline void check_and_store_ip_trace_id(struct __ctx_buff *ctx, __u8 ip_opt_type_value) {
```
bowei, line 100: (minor) Seems like the #define is both a CPP define and passed as a param into the function. Any particular reason why it is doing double duty?

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+```
+At ingress, the function `trace_id_from_ctx` is called to parse the IP options from incoming packets. If the packet is valid, it proceeds to parse the IP options to extract the trace ID.
+
+We loop over the IP options to extract the trace ID. This loop is unrolled to a maximum of three iterations to minimize performance impact and meet BPF requirements. The maximum of three iterations accounts for any combination of trace ID, user-provided options, and either egress NAT or DSR options. If we read an IP option from the trace type which we configured (see Section 2), then this is stored as the `ip_trace_id`.
+
+Several edge cases are considered during parsing:
+- **No IP Options:** If the packet header indicates there are no options, parsing stops early, avoiding unnecessary processing.
+- **Invalid `ip_trace_id`:** invalid IDs are handled by setting the trace ID to 0.
+
+### Section 4: `ip_trace_id` storing
```
bowei, line 120: (nit) storing => state

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+    __type(value, __u64); // ip_trace_id type
+    __uint(pinning, LIBBPF_PIN_BY_NAME);
+} ip_trace_id_map __section_maps_btf;
+...
+// remaining declarations 
+#endif
+```
+
+**Pros**
+- Non-competitive storage; since this is not saved directly to a predefined field in the skb struct, there aren’t competing uses for this storage place. It is dedicated directly to the ip_trace_id of the packet being processed.
```
bowei, line 139: (minor) suggest: "Packet tracing ID is stored outside of the skb struct which has limited room for custom attributes."

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+
+**Considerations**
+- Must ensure that data is cleared correctly.
+- This is universal storage and not specifically stored in the SKB struct, so it must be cleared when a new packet is processed.
+- At an ingress point, if there is no trace ID stored in a packet, the stored `ip_trace_id` is cleared from the array-map.
```
bowei, line 144: I don't think it's cleared -- don't you mean set to 0?

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,586 @@
...
+
+2. **Packet Throughput**
+   a. **Objective:** Assess the maximum throughput the system can handle without degradation in performance.
+   b. **Metric:** Determine the packet sending rate (packets per second, pps) when the rate is set (1000 pps).
+
+3. **Resource Usage**
+   a. **Objective:** Monitor resource usage (CPU, memory) of the control plane with and without the tracing feature under different loads.
+   b. **Metric:** CPU and memory usage.
+
+## Setup
```
bowei, line 223: This experimental section is great -- is there a link to the code?

bowei (commented):
minor comments -- proposal looks great

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: We mean to say that a user defines which IP option is being used to store their `ip_trace_id`. There isn't a dedicated `ip_trace_id` field. A user can configure which IP option type is most appropriate to store a series of bytes into as the `ip_trace_id`, which we parse from accordingly. Any well-defined IP Option field type can be used (https://www.iana.org/assignments/ip-parameters/ip-parameters.xhtml)

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: When setting up the GCP cluster for a demo, we saw that IP options are unsupported by [L4 ELBs](https://cloud.google.com/load-balancing/docs/tcp). But they **are** supported by [L4 ILBs](https://cloud.google.com/load-balancing/docs/tcp/internal-proxy). This is specific to GCP. 

Any provider who will productionize this feature will have to ensure compatibility, select compatible IP options types, and/or are responsible for listing limitations.

bowei: I think we should change what is written here not be specific to GCP:

Something like "load balancer support for tracing may vary depending on the specific implementation. Some load balancers may remove IP options when forwarding traffic."

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: When a user sets the feature-gate to enable this feature, they do so by setting the specific IP option where the packets `ip_trace_id`s are embedded. So, we want to pass this into the parsing function in order to know where to read these `ip_trace_id`s from. 

The implementation specifics can be viewed [here](https://github.com/cilium/cilium/pull/33957/files) in the `pkg/datapath/linux/config/config.go` file if you're interested in more context.

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: I just pushed it to a repo with the benchmark testing scripts. [Here it is](https://github.com/Bigdelle/packet-tracing-benchmarks/blob/main/benchmark_cpu_usage.sh).

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: This is a user selected value, yes. There is no value they must use, it can depend on what is appropriate for their setup and traffic generation. e.x.: if traffic in some cluster already defines data at option 136 (stream ID), they can choose any compatible [IPv4 option type](https://www.iana.org/assignments/ip-parameters/ip-parameters.xhtml)

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: Yes; fixed.

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: Fixed.

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: Fixed.

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: Fixed.

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: Fixed.

cilium/CFP-33716-distributed-packet-tracing.md
sypakine: Context that drove this decision:

Original development of this feature observed that the provider systems only supported a subset of IP option types in a packet. Stream ID was identified as a viable type for that provider, but we didn't want an opinionated approach for all providers.

The user guide for this feature would be a good place to discuss considerations when setting the value.

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: Agreed; has been changed and updated on CFP.

cilium/CFP-33716-distributed-packet-tracing.md
```diff
@@ -0,0 +1,598 @@
...
+**Begin Design Discussion:** Jun 12, 2024  
+**End Design Discussion:**  
+**Cilium Release:**  
+**Authors:** Ben Bigdelle <bigdelle@google.com>, Mark St. John <markstjohn@google.com> based on the work of Nathan Perkins <nathanperkins@google.com>  
+**Reviewers:**  
+
+## Summary
+We aim to enhance cross-cluster and multi-system observability for flows that contain an `ip_trace_id` saved into a defined IP Option (referred to as a tagged flow), thus supporting tracing across networks. These tagged flows will trigger adding the ID value to Flow events messages and client (Hubble / Cilium Monitor) representation. This allows correlating flow events across systems which integrate with the span observability.
+
+Additionally, within Cilium, these flows will generate datapath debugging events regardless of the global debug-verbose setting. This will improve debuggability of production systems.
```
pchaigno, line 15: How would you prevent someone from flooding the ring buffers of Cilium agents by sending packets with the right option?

pchaigno (commented):
Thanks for the feature proposal! It's interesting :smiley: 

Just one question below on the verbose debug case.

cilium/CFP-33716-distributed-packet-tracing.md
Bigdelle: This feature is intended to be enabled only for specific debugging scenarios, so its use is controlled by system administrators. When this is disabled, the event emission functions as it does right now, and there will be no way to abuse this system.

When this feature is active, standard rate limiting still applies to tagged flows, which prevents flooding the event ring buffers.

sypakine: This would necessitate a separate buffer for traced events, otherwise the potentially disproportionate number of debug events produced by traced packets may starve other events (which would reduce the effectiveness of downstream consumers).

If Paul concurs, that should be called out in the doc (including the associated memory cost).

Bigdelle: Updated to specifically call this out.

pchaigno: > This feature is intended to be enabled only for specific debugging scenarios, so its use is controlled by system administrators. When this is disabled, the event emission functions as it does right now.

OK, got it. That makes sense. I had not realized this would be transiently enabled for debugging.

(Considering that, you might want to make this an option that can be enabled at runtime without an agent restart, like the policy verdict filters or audit mode.)

pchaigno (approved):

cilium/CFP-33716-distributed-packet-tracing.md
sypakine: Thanks @pchaigno! I take it you mean via the Read-write configuration? i.e.`cilium config ...` command (using your example of audit mode: `cilium config PolicyAuditMode=Enabled`).

If my understanding is correct, that is a great suggestion.

cilium/CFP-33716-packet-tracing.md
```diff
@@ -0,0 +1,613 @@
...
+![User story](./images/CFP-33716-user-story.png)
+
+- As a production cluster administrator, particular flows from a client are experiencing connection timeouts (e.g. [issue #32472](https://github.com/cilium/cilium/issues/32472). Enabling datapath debugging for all flows leads to event loss and low SNR. To extract dataplane information to root cause the issue, generate traffic with trace ID IP option to emit datapath debug information specific to this flow.
+
+## Proposal
+
+### Overview
+Offer support for packets containing a user-defined IP option field. For tagged incoming flows, handle events for this flow specially; emit datapath debugging events for these packets, include the ID in the event message, and implement hubble filtering by the ID.
+
+### Section 1: Feature Enablement
```
sypakine, line 44: Let's split the enablement similar to https://docs.cilium.io/en/stable/security/policy-creation/#enable-policy-audit-mode-entire-daemon:

Enable IP Tracing via DaemonSet vs Specific Agent.

sypakine (approved):

cilium/CFP-33716-packet-tracing.md
```diff
@@ -0,0 +1,613 @@
...
+	__uint(value_size, sizeof(__u32));
+	__uint(max_entries, __NR_CPUS__);
+} EVENTS_MAP __section_maps_btf;
+```
+
+## Security Considerations
+
+### Section 1: Flow handling
+
+Packets with trace identifiers are processed in the exact same way as non-tagged traffic. The only difference is that the event message includes the ip_trace_id identifier for tagged traffic (which we load via a constant-time check). This approach ensures there is no additional threat of denial of service or system flooding with these tagged packets.
```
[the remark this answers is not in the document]
Bigdelle: Fixed

cilium/CFP-33716-packet-tracing.md
Bigdelle: Fixed.

cilium/CFP-33716-packet-tracing.md
```diff
@@ -0,0 +1,636 @@
...
+
+## Meta
+**SIG:** SIG-Hubble, SIG-Datapath  
+**Sharing:** Public  
+**Begin Design Discussion:** Jun 12, 2024  
+**End Design Discussion:**  
+**Cilium Release:**  
+**Authors:** Ben Bigdelle <bigdelle@google.com>, Mark St. John <markstjohn@google.com> based on the work of Nathan Perkins <nathanperkins@google.com>  
+**Reviewers:**  
+
```
joestringer, line 11: Could you add the status?

```suggestion
Status: Implementable
```

cilium/CFP-33716-packet-tracing.md
```diff
@@ -0,0 +1,636 @@
+# CFP-2024-06-12: Packet Tracing
+
+## Meta
+**SIG:** SIG-Hubble, SIG-Datapath  
+**Sharing:** Public  
+**Begin Design Discussion:** Jun 12, 2024  
+**End Design Discussion:**  
+**Cilium Release:**
```
joestringer, line 8: I would guess the target is during the next six months, so we can put that in.

```suggestion
**End Design Discussion:**  Aug 16, 2024
**Cilium Release:**  1.17
```

cilium/CFP-33716-packet-tracing.md
```diff
@@ -0,0 +1,636 @@
...
+- Default value: `<type>=0` (Feature is disabled).
+- When set at the DaemonSet level, all agents will embed the trace ID in the specified IP option type.
+
+Example DaemonSet configuration:
+```yaml
+args:
+  - --enable-ip-option-tracing=136
+```
+
+#### 1.2 Specific Agent-Level Enablement
```
joestringer, line 71: nit: 1.1 and 1.2 seem identical, and it's generally not meaningful to say whether config is in the daemonset or the agent. Configuration is specified in either a ConfigMap or in a CiliumNodeConfig CRD. I don't see the need to specify this twice.

```suggestion
#### 1.1 Configuration in cilium-agent
```

cilium/CFP-33716-packet-tracing.md
```diff
@@ -0,0 +1,636 @@
...
+- Non-competitive storage; packet tracing ID is stored outside of the skb struct which has limited room for custom attributes.
+
+**Considerations**
+- Must ensure that data is zero'ed correctly.
+- This is universal storage and not specifically stored in the SKB struct, so it must be zero'ed when a new packet is processed.
+- At an ingress point, if there is no trace ID stored in a packet, the stored `ip_trace_id` is zero'ed from the array-map.
+
+### Section 4: Event integration
+
+Include ID as event field for trace and drop structs and the flow message (for Hubble integration):
```
joestringer, line 144: Broadly this makes sense. We probably want to think about whether this uses a new trace_notify_v2 message type (and similarly new message type number + structure for drop). We've done this in the past at least, though it's probably worth digging into the reasoning why we did that to check whether it's still necessary.

cilium/CFP-33716-packet-tracing.md
```diff
@@ -0,0 +1,636 @@
...
+
+type IpTraceID struct {
+    TraceID      uint64 `protobuf:"varint,1,opt,name=trace_id,json=traceId,proto3" json:"trace_id,omitempty"`
+    IpOptionType uint8  `protobuf:"varint,2,opt,name=ip_option_type,json=ipOptionType,proto3" json:"ip_option_type,omitempty"`
+}
+```
+
+#### Tagged-flow ring buffer
+
+For tagged flows, there is a separate ring buffer to ensure that the processing of these events does not interfere with other events generated by Cilium. This separation helps maintain the performance of the system and mitigate concerns for tagged flows overflowing the event ring buffer, especially in high-traffic environments.
```
joestringer, line 183: What exactly is the purpose of the extra ring buffer? Is it just sharing the trace IDs to userspace?

A few reasons to ask this:
- Does userspace later need to correlate events to understand what the trace_ids mean? Does that add extra complication?
- If the class of information this generates is similar to the class for trace messages, then we could integrate the info into the existing events and require fewer userspace wakeups / event parsing+handling iterations. On the other hand if it is clearly a different priority/class of info vs. existing trace messages, then separating it can make sense. We do this today with some signals like when the CT/NAT maps become full, as those signals are more important; having a dedicated channel for those can make them more reliable.

cilium/CFP-33716-packet-tracing.md
```diff
@@ -0,0 +1,636 @@
...
+- Compatible with kernel versions post-v5.0-rc1.
+
+#### [Nonviable] Option 3.1: Use skb->mark or skb->CB field
+**Pros:**
+- The “mark” field is a generic packet mark for sk_buff and suitable for such purposes as adding a trace ID.
+
+**Considerations:**
+- Potential conflicts with other parts of the network that use this field.
+
+#### [Nonviable] Option 3.2: Store reference into skb->mark or skb->CB field
```
joestringer, line 629: What's the difference between Option 3.1 vs 3.2?

cilium/CFP-33716-packet-tracing.md
```diff
@@ -0,0 +1,636 @@
...
+      - L4 ILB
+      - Services
+      - Pod-to-pod (including across VPC)
+
+4. **IP Checksum Recalculation:**
+   - When IP options are added or modified, the IP checksum must be recalculated, adding additional processing overhead and complexity.
+
+Mitigation: This feature is a manually enabled debug option, giving users full control over performance overhead of IP option tagged packets. Users can manage overhead and performance concerns related to MTU, GRO compatibility, packet loss, and IP checksum recalculation. By making packet tagging the client's responsibility, Cilium remains unaffected by these concerns, allowing clients to incorporate these considerations into their use of the feature and determine the best way to utilize it according to their specific needs.
+
+## Benchmark Testing
```
joestringer, line 236: I appreciate the additional work over-and-beyond a CFP to put benchmarks in here :) Typically for a CFP this is not necessary, but since you've put it together I don't mind having this in here. Usually I'd expect this to go with the corresponding implementation PR rather than the design, but I know that the review for this design was slow this cycle and you've taken the time to develop the solution during that delayed design review.

joestringer (approved):
Thanks for working on this proposal, it sounds useful. I think this PR mainly just needs the metadata at the top to be updated to reflect that this is now considered "Implementable".

Reading through I did wonder about a couple of points, but I think that those points are probably crossing over from the design aspect and more into the implementation aspect, which I don't think needs to be signed off on in order to merge the PR considering it "Implementable". I figured it's just worth highlighting those below since those aspects could potentially come up during implementation / code review.

cilium/CFP-33716-packet-tracing.md
Bigdelle: This was a consideration very early in investigations about the resource usage of storing the value itself, or a reference to that value, so a memory address. Given that using the mark field wasn't viable, neither could be really considered. For purposes of documentation, I removed the 3.2 option as it doesn't add much value to this CFP.

cilium/CFP-33716-packet-tracing.md
Bigdelle: The extra ring buffer is a consideration pointed out by @sypakine and @pchaigno to ensure that events for non-tagged traffic don't get overrun by events for tagged traffic.

To your points:
1. The userspace will not need to correlate events to understand the trace_ids. It should be indistinguishable to a user that some events were pushed to the tagged flow ring buffer as opposed to the normal event ring buffer. This is to ensure that the current event generation is not overrun by tagged flows when this feature is enabled. 
2. As of the implementation I have right now, I do just that integration of these trace_ids into existing structs (https://github.com/cilium/cilium/pull/33957). These are implementation details that @sypakine will be taking over this work as I leave my internship.

cilium/CFP-33716-packet-tracing.md
Bigdelle: Fixed and pushed new changes.

cilium/CFP-33716-packet-tracing.md
Bigdelle: Fixed and pushed.

cilium/CFP-33716-packet-tracing.md
Bigdelle: Fixed.

cilium/CFP-33716-packet-tracing.md
joestringer: Hmm OK, I guess it would probably help to have a writeup of the exact patterns expected for usage of each ringbuffer and how the trace_ids flow through, depending on the traffic patterns.

For instance, let's say a packet with a trace_id arrives. Is only one trace_to event generated? And is that now generated to the new ringbuffer and not the old one? Or both?

Separately there's also the monitor_aggregation setting which avoids sending too many messages to userspace, so it'd be worth considering how that option intersects with this new option. I guess that's related to the thread with Paul about overloading; tagged packets could end up increasing userspace CPU usage for instance.

joestringer: I don't know if my feedback above is necessarily immediately actionable, and may be partially up to implementation details/discussion, but I would be happy to merge this maybe just with an extra line to highlight that this part is not fully resolved (unless we think we're likely to drive a clear conclusion from this discussion):

```suggestion
For tagged flows, there is a separate ring buffer to ensure that the processing of these events does not interfere with other events generated by Cilium. This separation helps maintain the performance of the system and mitigate concerns for tagged flows overflowing the event ring buffer, especially in high-traffic environments.

UNRESOLVED: How does the agent choose when to use this new ringbuffer or not? How does this integrate with Monitor Aggregation?
```

cilium/CFP-33716-packet-tracing.md
joestringer: ^ This doesn't need to be resolved before merging the CFP as "Implementable". We can figure out those details on the PR.

cilium/CFP-33716-packet-tracing.md
Bigdelle: Fixed and pushed this change.

cilium/CFP-33716-packet-tracing.md
Bigdelle: Sounds good, I updated the CFP with relevant changes, please let me know if there's anything else you need before merging. Thank you!

joestringer:
I think this CFP is in a solid state for implementing. There's a couple of details that could maybe be hashed out a bit more, but I don't think they ultimately detract that much from the overall proposal, and may be impacted by the experience of implementing it. The proposal has been out plenty long enough for @cilium/sig-hubble and @cilium/sig-datapath to provide feedback, and there's now committers endorsing the proposal, so I think this is in a good enough state to merge.

As and when the implementation matures and impacts the design, we can always iterate further on this CFP, but we don't have to hold it as a PR during that period. Let's get this merged :rocket:

Issue #45, opened 2024-07-17
Improve contribution guidelines


xtineskim:
From our Cilium community call, there can be some helpful follow up tasks to improve contributions to this repo
- [x] Update the main README.md to have a FAQ or similar on how to find which SIG to get in touch with
- [x] _Maybe_ look into having groups in the community sig groups be taggable in CFP PRs with Github?

joestringer:
> groups in the community sig groups be taggable in CFP PRs with Github

This should already be doable, just need to tag @cilium/<group>. This will trigger GitHub notifications to members of that group.

xtineskim:
a shortlist from looking and getting inspiration from other CNCF projects:
- Having a shared google drive for docs that are getting feedback (before submitting a PR), allows for better integration with google groups mailing lists.
- Having guidelines on expected review time (ex: design CFPs will be initially reviewed within 2 weeks, PRs will be reviewed within 2 days)

xmulligan:
Looks like we checked off the first one with #46 and the second one is possible. I think generally in the project we don't provide expectations on really anything for timelines. I think there is this [google drive](https://drive.google.com/drive/folders/0AOI5PGS6DjvyUk9PVA), but I'm not sure how much we use it anymore

PR #46, opened 2024-07-17
Add link to Cilium community teams repo, re-organize the order

xtineskim:
Related to https://github.com/cilium/design-cfps/issues/45, but adding in link to Cilium Community Teams so folks know who to ping on SIG related topics


joestringer:
For Cilium there is also CODEOWNERS which can help provide some more hints - for instance someone might not know which of the dozens of teams to contact, what the scope of those teams is, or which files they're responsible for. This could help to clarify some of these questions: https://github.com/cilium/cilium/blob/main/CODEOWNERS

README.md
```diff
@@ -22,7 +22,13 @@ accept.
 To create a CFP, it is recommended to use the `CFP-003-template.md`
 file as an outline. The structure of this template is meant to provide a starting
 point for people. Feel free to edit and modify your outline to best fit your
-needs when creating a proposal. When you are ready to submit your CFP please:
+needs when creating a proposal.
+
+Many design docs also begin their life as a Google doc or other shareable
+file for easy commenting and editing when still in the early stages of discussion.
+Once your proposal is done, submit it as a PR to the design-cfps folder.
```
xmulligan, line 29:
```suggestion
Once your proposal is ready, submit it as a PR to the design-cfps folder for the project it is for.
```

xtineskim: hmm, there aren't any designated folders, they are just all under the `cilium` folder

xmulligan: I was thinking that each project would have its own folder. It just hasn't happened yet since there haven't been CfPs for hubble or tetragon yet

xtineskim: ah gotcha - in that case, until the folders are made, i think we should keep it as is, make the change to the wording once they are present?

xmulligan: Maybe we can just add the folders in this PR too?

xmulligan (approved):

Issue #48, opened 2024-08-09
Should we merge PRs when they are not "Implementable"?


joestringer:
We've recently clarified status for CFPs (See #42).

While attempting to apply this standard to existing proposals, I noticed that there may be interest in merging CFPs when they are not considered "Implementable", but the current process largely assumes that this is the minimum bar to merge.

More specifically, there are a couple of cases I'm considering:

### Dormant Status

These may require subsequent discussion to identify and resolve key questions prior to marking the CFP as "Implementable". I've already merged #7 under this category, and I think #36 also applies under this category. The status text does not currently really describe one way or another whether such PRs should be merged, but it seems like a reasonable way to "close the loop" on an open PR. The alternative I see would be to just close dormant CFP proposals, and if we wish to adopt them in future then someone would have to resurrect the PR afresh.

### Draft Status

The current text prescribes that one of the properties of a draft is that it's not merged. It also describes "Implementable" PRs as being merged. This means that a specific idea needs to be significantly matured before it is merged into the repository. I noticed in #32 that contributors have been managing review discussions in their own public repositories in order to more easily iterate on the discussion. As a pattern, this can work, but it also draws the discussion away from the central official repositories, making them generally less visible to the overall community.

Furthermore when I reviewed the above PR, I had dozens of resulting comments. With so much activity, the GitHub UI tends to begin to inhibit healthy conversation by being slow, making it hard to find comments, and so on. This makes me wonder whether we should redefine the "Draft" status such that we _could_ merge proposals in draft status as long is the core idea has high level agreemeent. We could set the culture to expect "TODOs" to be documented in a draft CFP before merge. Then, before the proposal can graduate to to "Implementable" we would need to ensure that key questions have reasonable solutions and have no major flaws and no significant lack of consensus.

Overall with this proposal, if we had community members working closely together, I could imagine staging the development of a CFP where an initial version just describes the summary and goals to establish high level consensus on the core idea, then subsequent PRs could develop the idea, each of which could undergo some degree of review to merge, then as key ideas/questions/impacts are introduced, even those could be separately committed, and only once the CFP gets to a point where there is rough consensus and committers believe it has been sufficiently explored, then a subsequent PR could move it to "Implementable". While I don't know for sure whether contributors are looking for this degree of incremental progress on CFPs given that CFPs are often written proposed by a single individual, I would be open to exploring mechanisms like this for more incremental progress.

joestringer:
[...]

PR #54, opened 2024-08-21
CFP-30984 : Add CFP for DNS proxy HA - V2

hemanthmalla:
Add CFP for https://github.com/cilium/cilium/issues/30984 

Follow up of https://github.com/cilium/design-cfps/pull/32

### DIFF
--- a/cilium/CFP-30984-dns-proxy-ha-v2.md
+++ b/cilium/CFP-30984-dns-proxy-ha-v2.md
@@ -1,0 +1,109 @@
+# CFP-30984: toFQDN DNS proxy HA
+
+**SIG: SIG-POLICY**
+
+**Begin Design Discussion:** 2024-08-19
+
+**Cilium Release:** 1.17
+
+**Authors:** Hemanth Malla <hemanth.malla@datadoghq.com>, Vipul Singh <singhvipul@microsoft.com>
+
+## Summary
+
+Cilium agent uses a proxy to intercept all DNS queries and obtain necessary information to enforce FQDN network policies. However, the lifecycle of this proxy is coupled with the cilium agent. When an endpoint has a toFQDN network policy in place, cilium installs a redirect to capture all DNS traffic. So, when the agent is unavailable, all DNS requests time out, including when DNS name to IP address mappings are already in place for this name. DNS policy unload on shutdown can be enabled on the agent, but it works only when the L7 policy is set to * and the agent is shutdown gracefully.
+
+This CFP introduces a standalone DNS proxy that can run alongside the Cilium agent, which should eliminate hard dependency for names that already have policy map entries in place.
+
+## Motivation
+
+Users rely on toFQDN policies to enforce network policies against traffic to destinations outside the cluster, typically to blob storage / other services on the internet. Rolling out the Cilium agent should not result in packet drops. Introducing a high availability (HA) mode will allow for increased adoption of toFQDN network policies in critical environments.
+
+## Goals
+
+* Introduce a streaming gRPC API for exchanging FQDN policy related information.
+* Introduce a standalone DNS proxy (SDP) that binds on the same port as built-in proxy with SO_REUSEPORT.
+* Enforce L7 DNS policy via SDP.
+
+## Non-Goals
+
+* Updating new DNS <> IP mappings when the agent is down.
+
+## Proposal
+
+### Overview
+
+There are two parts to enforcing toFQDN network policy. L4 policy enforcement against IP addresses resolved from an FQDN and policy enforcement on DNS requests (L7 DNS policy). To enforce L4 policy, per endpoint policy bpf maps need to be updated. We'd like to avoid multiple processes writing entries to policy maps, so the standalone DNS proxy (SDP) needs a mechanism to notify agent of newly resolved FQDN <> IP address mappings. This CFP proposes exposing a new gRPC streaming API from the cilium agent. Since the connection is bi-directional, the cilium agent can reuse the same connection to notify the SDP of L7 DNS policy changes. 
+
+Additionally, SDP needs to translate the IP address to cilium identity to enforce the policy. Our proposal involves retrieving the identity mapping from the cilium_ipcache BPF map. Currently L7 proxy (envoy) relies on accessing ipcache directly as well. We aren't aware of any efforts to introduce an abstraction to avoid reading bpf maps owned by the cilium agent beyond the agent process. If / when such abstraction is introduced, SDP can also be updated to implement a similar mechanism. We brainstormed a few options on how the API might look like if we exchange IP to identity mappings via the API as well, but it brings in a lot of additional complexity to keep the mappings in sync as endpoints churn. This CFP will focus on the contract between SDP and Cilium agent to exchange minimum information for implementing the high availability mode.
+
+In addition to existing unix domain socket (UDS) opened by the agent to host HTTP APIs, we'll need a new UDS for the gRPC streaming service with similar permissions. 
+
+### RPC Methods
+
+Method : UpdateMappings (Invoked from SDP to agent)
+
+_rpc UpdatesMappings(steam FQDNMapping) returns (Result){}_
+Request :
+```
+message FQDNMapping {
+    string FQDN = 1;    // DNS Name of the request made by the client
+    repeated bytes IPS = 2; // Resolved IP addresses
+    uint32 TTL = 3;
+    uint64 source_identity = 4; // Identity of the client making the DNS request
+    int dns_response_code = 5;
+}
+```
+Response :
+```
+message Result {
+    bool success = 1;
+}
+```
+
+Method : UpdatesDNSRules ( Invoked from agent to SDP via bi-directional stream )
+
+_rpc UpdatesDNSRules(stream DNSPolicies) returns (Result){}_
+Request :
+```
+message DNSPolicy {
+  uint64 source_identity = 1;  // Identity of the workload this L7 DNS policy should apply to
+  repeated string dns_pattern = 2;  // Allowed DNS pattern this identity is allowed to resolve
+  uint64 dns_server_identity = 3;  // Identity of destination DNS server
+  uint16 dns_server_port = 4;   
+  uint8 dns_server_proto = 5;
+}
+
+
+message DNSPolicies {
+  repeated DNSPolicy l7_dns_policy = 1;
+}
+
+```
+
+Response :
+```
+message Result  {
+     bool success = 1;
+}
+```
+
+### Load balancing
+
+SDP and agent's DNS proxy will run on the same port using SO_REUSEPORT. By default, kernel will use round robin algorithm to distribute load evenly between all sockets in the reuseport group. If cilium agent's DNS proxy goes down, kernel will automatically switch all traffic to SDP and vice versa. In the future, we can consider using a custom bpf program to make SDP only act as a hot standby. See (PoC)[https://github.com/hemanthmalla/reuseport_ebpf/blob/main/bpf/reuseport_select.c] / eBPF summit 2023 talk for more details.
+
+
+### High Level Information Flow
+
+* Agent starts up with gRPC streaming service.
+* SDP starts up.
+* Connects to gRPC service, retrying periodically until success.
+* Agent sends current snapshot for L7 DNS Policy enforcement via UpdatesDNSRules to SDP.
+* On policy recomputation,  agent invokes UpdatesDNSRules.
+* On DNS request from the client, DNS request redirects to DNS proxy port.
+* Kernel round robin load balances between SDP and built in proxy.
+* Assuming SDP gets the request, SDP enforces L7 DNS policy.
+  * Lookup identity based on IP address via bpf map.
+  * Check against policy snapshot if this identity is allowed to resolve the current DNS name and is allowed to talk to DNS server target identity (also needs lookup).
+* Make upstream DNS request from SDP.
+* On response, SDP invokes UpdatesMappings() to notify agent of new mappings.
+* Release DNS response after success from  UpdatesMappings() / timeout.


cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,109 @@
...
+
+Additionally, SDP needs to translate the IP address to cilium identity to enforce the policy. Our proposal involves retrieving the identity mapping from the cilium_ipcache BPF map. Currently L7 proxy (envoy) relies on accessing ipcache directly as well. We aren't aware of any efforts to introduce an abstraction to avoid reading bpf maps owned by the cilium agent beyond the agent process. If / when such abstraction is introduced, SDP can also be updated to implement a similar mechanism. We brainstormed a few options on how the API might look like if we exchange IP to identity mappings via the API as well, but it brings in a lot of additional complexity to keep the mappings in sync as endpoints churn. This CFP will focus on the contract between SDP and Cilium agent to exchange minimum information for implementing the high availability mode.
+
+In addition to existing unix domain socket (UDS) opened by the agent to host HTTP APIs, we'll need a new UDS for the gRPC streaming service with similar permissions. 
+
+### RPC Methods
+
+Method : UpdateMappings (Invoked from SDP to agent)
+
+_rpc UpdatesMappings(steam FQDNMapping) returns (Result){}_
```
squeed, line 45: What happens if the agent is down in this period? are they cached?

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,109 @@
...
+
+* Agent starts up with gRPC streaming service.
+* SDP starts up.
+* Connects to gRPC service, retrying periodically until success.
+* Agent sends current snapshot for L7 DNS Policy enforcement via UpdatesDNSRules to SDP.
+* On policy recomputation,  agent invokes UpdatesDNSRules.
+* On DNS request from the client, DNS request redirects to DNS proxy port.
+* Kernel round robin load balances between SDP and built in proxy.
+* Assuming SDP gets the request, SDP enforces L7 DNS policy.
+  * Lookup identity based on IP address via bpf map.
```
squeed, line 105: I know that @joestringer had some thoughts about making the ipcache a public API. We might as well do it, but it may need some further explanation.

joestringer: I really don't think ipcache BPF format should be public API. Rather than committing the @cilium/sig-datapath folks to supporting a specific format forever, I would say there should be some mitigation in the SDP + testing to demonstrate when breaking changes occur, then we'll deal with the upgrade problem at that point.

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,109 @@
...
+
+Method : UpdatesDNSRules ( Invoked from agent to SDP via bi-directional stream )
+
+_rpc UpdatesDNSRules(stream DNSPolicies) returns (Result){}_
+Request :
+```
+message DNSPolicy {
+  uint64 source_identity = 1;  // Identity of the workload this L7 DNS policy should apply to
+  repeated string dns_pattern = 2;  // Allowed DNS pattern this identity is allowed to resolve
+  uint64 dns_server_identity = 3;  // Identity of destination DNS server
```
squeed, line 71: This should be repeated

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,109 @@
...
+In addition to existing unix domain socket (UDS) opened by the agent to host HTTP APIs, we'll need a new UDS for the gRPC streaming service with similar permissions. 
+
+### RPC Methods
+
+Method : UpdateMappings (Invoked from SDP to agent)
+
+_rpc UpdatesMappings(steam FQDNMapping) returns (Result){}_
+Request :
+```
+message FQDNMapping {
```
squeed, line 48: Consider adding source (endpoint) IP here.

cilium/CFP-30984-dns-proxy-ha-v2.md
joestringer: "mitigation in the SDP" is a bit terse, but if we can version it somehow or at least detect when the map properties like key/value sizes are different, then it would be obvious that the map has been made incompatible.

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+Method : UpdatesDNSRules ( Invoked from agent to SDP via bi-directional stream )
+
+_rpc UpdatesDNSRules(stream DNSPolicies) returns (Result){}_
+Request :
+```
+message DNSPolicy {
+  uint64 source_identity = 1;  // Identity of the workload this L7 DNS policy should apply to
+  repeated string dns_pattern = 2;  // Allowed DNS pattern this identity is allowed to resolve
```
joestringer, line 70: I mentioned during the sig-policy meeting discussion, as part of implementation we will want to define whether this is in the API format from CNP or something else. There's a chance that by the time it's been processed internally in cilium-agent this information is already converted into a regex.

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+### Load balancing
+
+SDP and agent's DNS proxy will run on the same port using SO_REUSEPORT. By default, kernel will use round robin algorithm to distribute load evenly between all sockets in the reuseport group. If cilium agent's DNS proxy goes down, kernel will automatically switch all traffic to SDP and vice versa. In the future, we can consider using a custom bpf program to make SDP only act as a hot standby. See (PoC)[https://github.com/hemanthmalla/reuseport_ebpf/blob/main/bpf/reuseport_select.c] / eBPF summit 2023 talk for more details.
+
+
+### High Level Information Flow
+
+* Agent starts up with gRPC streaming service.
```
joestringer, line 97: From sig-policy discussion: Agent starts opens this socket after syncing k8s resources and recreates+fully populates ipcache map.

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,109 @@
...
+}
+```
+
+Method : UpdatesDNSRules ( Invoked from agent to SDP via bi-directional stream )
+
+_rpc UpdatesDNSRules(stream DNSPolicies) returns (Result){}_
+Request :
+```
+message DNSPolicy {
+  uint64 source_identity = 1;  // Identity of the workload this L7 DNS policy should apply to
```
joestringer, line 69: Are these policies inherently egress policies from the Pod to the DNS server?

In netpol we often talk about "subjects" (of a policy) and "peers" (that the subject communicates with). This API is written with "source" and "DNS server" which strictly assumes an egress connection from a non-DNS Pod to a DNS server.

While that's probably fine for the target use case, at the implementation stage it could have some semantic implications if for instance the policy is ingress and the subject is actually the DNS server (such as kube-dns). This probably speaks a little to potential future work outside the scope of this CFP - for instance whether the component is intended to become a full-fledged L7 proxy for DNS traffic, or have a relatively limited scope for egress from local Pods only.

(I'm also fine to just pick this API for now, implement, get some experience, then decide whether we want to change it before graduating stability of the API).

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,109 @@
+# CFP-30984: toFQDN DNS proxy HA
+
+**SIG: SIG-POLICY**
+
+**Begin Design Discussion:** 2024-08-19
+
+**Cilium Release:** 1.17
+
+**Authors:** Hemanth Malla <hemanth.malla@datadoghq.com>, Vipul Singh <singhvipul@microsoft.com>
+
```
joestringer, line 10: Administrative point:

If we think we agree on the basics here, but would like more of the design to be fleshed out:

```suggestion

Status: Provisional

```
(If provisional, please highlight anything that still needs to be expanded with the text "UNRESOLVED: _description of unresolved design point_").

Or alternatively if we think the design is satisfactorily complete:


```suggestion

Status: Implementable

```

See status in the readme for more details. https://github.com/cilium/design-cfps?tab=readme-ov-file#status

joestringer: This was the point I made at the end of the discussion with @cilium/sig-policy today. Whatever we agree upon, we should be able to merge in. If there are points left to iterate on, we can highlight those points. If we've resolved the points of difference we can mark it implementable. Goal is to make iteration easier so we don't have to overthink an entire design that is subject to change during implementation, but rather we can more incrementally change the design in-tree as we discuss, debate, implement and learn.

cilium/CFP-30984-dns-proxy-ha-v2.md
hemanthmalla: Makes sense. I'll add a note in the CFP to make the failure explicit and loud (both in tests and implementation )

cilium/CFP-30984-dns-proxy-ha-v2.md
hemanthmalla: That's a good point. This CFP is intended to solve only for egress policies. So I'm updating `l7_dns_policy` field to `egress_l7_dns_policy` to make it a little more explicit.

cilium/CFP-30984-dns-proxy-ha-v2.md
hemanthmalla: Adding this to the CFP.

cilium/CFP-30984-dns-proxy-ha-v2.md
hemanthmalla: Added repeated to port and proto as well.

cilium/CFP-30984-dns-proxy-ha-v2.md
hemanthmalla: Thanks for raising this, will check this in our PoC and add more details. Do you anticipate any issues if we define this as regex instead of the notation used in CNPs ? Since SDP will need to evaluate this as regex anyway.

cilium/CFP-30984-dns-proxy-ha-v2.md
hemanthmalla: I'll add a status field once we close out the open discussion items.

hemanthmalla: Also looks like we haven't documented `Provisional ` in the readme yet ?

hemanthmalla: Also looks like we haven't documented `Provisional ` in the readme yet ?

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,109 @@
...
+## Non-Goals
+
+* Updating new DNS <> IP mappings when the agent is down.
+
+## Proposal
+
+### Overview
+
+There are two parts to enforcing toFQDN network policy. L4 policy enforcement against IP addresses resolved from an FQDN and policy enforcement on DNS requests (L7 DNS policy). To enforce L4 policy, per endpoint policy bpf maps need to be updated. We'd like to avoid multiple processes writing entries to policy maps, so the standalone DNS proxy (SDP) needs a mechanism to notify agent of newly resolved FQDN <> IP address mappings. This CFP proposes exposing a new gRPC streaming API from the cilium agent. Since the connection is bi-directional, the cilium agent can reuse the same connection to notify the SDP of L7 DNS policy changes.
```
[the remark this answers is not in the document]
joestringer: I don't see why we couldn't do L4 for actual traffic.

cilium/CFP-30984-dns-proxy-ha-v2.md
joestringer: Ah right, provisional is just under discussion in https://github.com/cilium/design-cfps/issues/48 but not documented yet.

cilium/CFP-30984-dns-proxy-ha-v2.md
joestringer: Other than that it should be specified that this is the Go regex dialect rather than other kinds of regex, I'm not immediately aware of any limitations. I have a feeling that regex encoding is difficult to get right when crossing a public API boundary which is part of the motivation to avoid full regex in the CNP API, but maybe as a somewhat "internal" API between Cilium components, we can get away with it.

cilium/CFP-30984-dns-proxy-ha-v2.md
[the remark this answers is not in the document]
hemanthmalla: Ah NVM, I mis-read the spec.
```
  egress:
  - toFQDNs:
    - matchPattern: "*.github.com"
    toPorts:
    - ports:
      - port: "443"
        protocol: TCP
```
Updating to L3/L4

cilium/CFP-30984-dns-proxy-ha-v2.md
gandro: For what it's worth, I think we should have a `message DNSServer { identity, port, proto }` that is repeated, rather than having all three fields repeated individually

hemanthmalla: Updating this.

cilium/CFP-30984-dns-proxy-ha-v2.md
hemanthmalla: We validated that we're sending the patterns defined originally in CNPs and then processing them in SDP. Added a note about this in the CFP.

cilium/CFP-30984-dns-proxy-ha-v2.md
hemanthmalla: Added a source_ip field.

tamilmani1989:
@squeed / @joestringer we addressed your comments and appreciate if you can another look. Haven't decided on status of CFP though. Do you think we can set to `implementable`?

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,123 @@
...
+
+Additionally, SDP needs to translate the IP address to cilium identity to enforce the policy. Our proposal involves retrieving the identity mapping from the cilium_ipcache BPF map. Currently L7 proxy (envoy) relies on accessing ipcache directly as well. We aren't aware of any efforts to introduce an abstraction to avoid reading bpf maps owned by the cilium agent beyond the agent process. If / when such abstraction is introduced, SDP can also be updated to implement a similar mechanism. We brainstormed a few options on how the API might look like if we exchange IP to identity mappings via the API as well, but it brings in a lot of additional complexity to keep the mappings in sync as endpoints churn. This CFP will focus on the contract between SDP and Cilium agent to exchange minimum information for implementing the high availability mode.
```
joestringer, line 37: nit: Some of the text here hedging is hedging over a technical decision which should probably be deferred to a dedicated section:

```suggestion
Additionally, SDP needs to translate the IP address to cilium identity to enforce the policy. This CFP proposes to retrieve the identity mapping from the cilium_ipcache BPF map. If / when other another abstraction is introduced for getting IP<->Identity mappings, this implementation can use that abstraction. This CFP will focus on the contract between SDP and Cilium agent to exchange minimum information for implementing the high availability mode.
```

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,123 @@
...
+* On response, SDP invokes UpdatesMappings() to notify agent of new mappings.
+* Release DNS response after success from  UpdatesMappings() / timeout.
+
+### Handling SDP <> Agent re-connections
+
+* When the agent is unavailable, SDP will periodically attempt to re-connect to the streaming service. Any FQDN<>IP mappings resolved when the agent is down will be cached in SDP and `UpdatesMappings` will be retried after establishing the connection.
+  * A new bpf map for ipcache is populated on agent startup, so SDP needs to re-open the ipcache bpf map when the connection is re-established. See https://github.com/cilium/cilium/pull/32864 for similar handling in envoy.
+  * On a new connection from SDP, the agent will invoke `UpdatesDNSRules` to notify SDP of all L7 DNS policy rules.
+
+* SDP will not listen on the DNS proxy port until a connection is established with cilium agent and initial L7 DNS policy rules are received. Meanwhile, built-in DNS proxy will continue to serve requests. SDP relies on cilium agent for initial bootstrap. In future, we could make SDP retrieve initial policy information from other sources, but this is not in scope for this CFP.
```
joestringer, line 123:
```suggestion
* SDP will not listen on the DNS proxy port until a connection is established with cilium agent and initial L7 DNS policy rules are received. Meanwhile, built-in DNS proxy will continue to serve requests. SDP relies on cilium agent for initial bootstrap. In future, we could make SDP retrieve initial policy information from other sources, but this is not in scope for this CFP.

### Handling Cilium version upgrades

UNRESOLVED: Document how we can support upgrades while still retaining flexibility to change datapath design like ipcache over time.
```

joestringer (commented):
I'd be fine with merging in "Provisional" status with these changes. Given how this feature interacts with availability during upgrade, I think we'd need to align on what upgrade should look like before really considering this as implementable.

The danger I see for not aligning around upgrade would be that this feature could either block initiatives from @cilium/sig-datapath / @cilium/sig-policy to optimize the dataplane, or alternatively such initiatives could end up breaking this feature which we don't really want either. 

Either this or we say that figuring out upgrade is part of moving this feature from alpha->beta or beta->stable. We haven't outlined criteria for alpha/beta/stable features in the project yet but I'd be open to discussions about how we set expectations around that.

tamilmani1989:
> Given how this feature interacts with availability during upgrade, I think we'd need to align on what upgrade should look like before really considering this as implementable

Agreed. we can create separate doc/CFP just to brainstorm about how upgrade works with SDP and in-agent dnsproxy. we can note `upgrade` as `unresolved` and merge this CFP with provisional status

> The danger I see for not aligning around upgrade would be that this feature could either block initiatives from @cilium/sig-datapath / @cilium/sig-policy to optimize the dataplane, or alternatively such initiatives could end up breaking this feature which we don't really want either.

That's true. For version compatibility, if required, we can follow same path as envoy by tying SDP version with cilium versions https://github.com/cilium/proxy#version-compatibility-matrix. The motive should be try to support upgrade without breaking but it may become necessary to break in some cases where we can come up with something like version compatibility.

joestringer:
> For version compatibility, if required, we can follow same path as envoy by tying SDP version with cilium versions 

AFAIK cilium-proxy doesn't try to solve upgrade, this is just stating which version of Cilium ships with which Envoy version.

cilium/CFP-30984-dns-proxy-ha-v2.md
hemanthmalla: @joestringer to elaborate on some of the challenges you have in mind, are you thinking about the window between agent upgrade (with a datapath change involving ipcache format) and SDP upgrade ?

I'm thinking since SDP is also in tree, whenever we introduce such a change, SDP can implement handling both versions and we upgrade SDP first. I'll add a note in the CFP to make sure we add a test during the implementation phase to catch such datapath changes.

Do you have other scenarios in mind ?

cilium/CFP-30984-dns-proxy-ha-v2.md
hemanthmalla: Added a note for this.

tamilmani1989:
> > For version compatibility, if required, we can follow same path as envoy by tying SDP version with cilium versions
> 
> AFAIK cilium-proxy doesn't try to solve upgrade, the link is just stating which version of Cilium ships with which Envoy version.

@joestringer hope you are talking about minor upgrades. In my opinion, upgrade should be backward compatible atleast with immediate previous minor. Old SDP <-> New cilium or New SDP <-> old cilium should work. It might require for us to support both old/new api/fields in new version for some time. Subsequently in later releases, we can remove deprecated api/fields after we make wide announcement.

cilium/CFP-30984-dns-proxy-ha-v2.md
joestringer: That's essentially what I'm thinking about, and yeah having some general sense like you describe about how such upgrades should be handled is worth noting down. Particularly since now that also expresses intent for a given SDP version to work with the previous version of Cilium.

joestringer: cc @tamilmani1989 this seems a little different from your views on upgrade below, as you were expressing intent for either order to work (SDP newer or older than cilium minor version).

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Handling Upgrades
+
+Other than the streaming API from the agent, this CFP introduces a dependency on the ipcache bpf map which isn't a stable API exposed to components beyond the agent. Sufficent tests will be added to catch such datapath changes impacting SDP. In order to support a safe upgrade path, SDP would need to support reading from the current and future formats of the map (including possibly reading from an entire new map).
```
joestringer, line 127: I think this is a reasonable state of intent / goals, but there is a lot of detail missing here. 

- What are "sufficient tests"?
- What upgrade path do you intend to support?
- What assumptions are you making about the formats of the map? How do you intend to detect if the format is different?

cilium/CFP-30984-dns-proxy-ha-v2.md
tamilmani1989: correct, that was my intent. We tend to keep SDP and cilium deployment separate and not part of same helm chart. so, we can't control the order of deployment. So either order should work.

cilium/CFP-30984-dns-proxy-ha-v2.md
tamilmani1989: @joestringer  reading the BPF map is a common issue for both Envoy and SDP. I hope this CFP doesn't introduce any new dependencies. I'm not pointing to Envoy as an excuse; instead, we should work together to solve this issue for both, which can be a separate discussion. One option we discussed was to abstract the BPF map internals with a Wrapper API that Envoy and SDP can use. This way, they wouldn't need to worry about underlying improvements or changes, and the Wrapper API would remain unaffected

joestringer: > > What assumptions are you making about the formats of the map? How do you intend to detect if the format is different?
> 
> Actually we already have a layer of indirection from the userspace cache and from SDP we're reusing the same cache structure. So any changes to the map, will come with accompanying userspace cache format changes, which SDP would also pick. So this would cover our needs from a format change perspective ?

Sorry, I don't quite follow. Are both SDP and cilium-agent making BPF syscalls to interpret BPF map formats? If yes, what expectations do you have if we want to change the ipcache format in future?

I think we've discussed earlier some potential expectations and solutions, but right now I'm finding it difficult to understand exactly what we're agreeing to. If we agree that this is a future problem, I'm fine with that, just know that means we do not have consensus about exactly how upgrades will behave, and that may lead to temporary packet loss when components are replaced. As long as we're on the same page and document this, I am fine. What I want to avoid is a situation where y'all have high standards for expectations of behavior during upgrade, but we never wrote down exactly what those expectations were and didn't get agreement from other developers it could affect.

cilium/CFP-30984-dns-proxy-ha-v2.md
joestringer: There's a section at the end of the CFP now for this, so we can resolve this thread.

Just noting for now that in contrast to @tamilmani1989 's comment, either order is *not* expected to work at this stage.

cilium/CFP-30984-dns-proxy-ha-v2.md
joestringer: > SDP would need to support reading from the current and future formats of the map

So I think that this does capture the essence of the proposed solution for upgrades.

@hemanthmalla , @tamilmani1989 and I discussed out-of-band how this could work:

> today there's only one ipcache format and no need for a wrapper. But let's say we want to change the ipcache format. We introduce this wrapper which understands two entirely different map formats and figures out the right format to read, based on some heuristic. Maybe key/value size or something. Then we carry that format change logic as a library for a release or two. If we also upgrade SDP first, it will migrate onto the wrapper and continue to work with the old ipcache ... if you then upgrade cilium, cilium will migrate to the newer ipcache format and the SDP will auto-detect / start reading the new format

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -23,10 +23,11 @@ Users rely on toFQDN policies to enforce network policies against traffic to des
 * Introduce a streaming gRPC API for exchanging FQDN policy related information.
 * Introduce a standalone DNS proxy (SDP) that binds on the same port as built-in proxy with SO_REUSEPORT.
 * Enforce L7 DNS policy via SDP.
+* When an endpoint's DNS traffic is selected by an L7 policy, DNS requests and responses will be forwarded to their destinations via SDP even if cilium-agent is not running. So, clients re-resolving DNS to establish new connections will not be blocked anymore if the IP addresses from the new resolution are unchanged. Note that the L3/L4 policy for the resolved names should have already been plumbed when the agent was running.
```
joestringer, line 26: This is not a goal, this is implementation detail. Should this be dropped or moved under the proposal section?

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Handling SDP <> Agent re-connections
+
+* When the agent is unavailable, SDP will periodically attempt to re-connect to the streaming service. Any FQDN<>IP mappings resolved when the agent is down will be cached in SDP and `UpdatesMappings` will be retried after establishing the connection.
+  * A new bpf map for ipcache is populated on agent startup, so SDP needs to re-open the ipcache bpf map when the connection is re-established. See https://github.com/cilium/cilium/pull/32864 for similar handling in envoy.
+  * On a new connection from SDP, the agent will invoke `UpdatesDNSRules` to notify SDP of all L7 DNS policy rules.
+
+* SDP will not listen on the DNS proxy port until a connection is established with cilium agent and initial L7 DNS policy rules are received. Meanwhile, built-in DNS proxy will continue to serve requests. SDP relies on cilium agent for initial bootstrap. In future, we could make SDP retrieve initial policy information from other sources, but this is not in scope for this CFP.
+
+### Handling Upgrades
```
joestringer, line 130: I still don't feel that this section adequately explains upgrades. The text is capturing some of the key discussions we have had though, which helps.

Overall I don't feel that continuing this path by debating back and forth in English is really helping drive the design forward. I think that at this point it makes sense to just accept that some aspects of this CFP are underspecified / not agreed upon, but the overall design is still "implementable" and ultimately subject to review of the actual implementation. As we learn more through implementation we can iron out any remaining differences.

joestringer (commented):
Minor suggestions:
1. Add the status
2. Move the new goal into the proposal section

Then let's move forward with this. I think that developing and testing the implementation will do more practical good than continuing discussion on the CFP.

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,134 @@
...
+}
+```
+Response :
+```
+message UpdatesMappingsResult {
+    bool success = 1;
+}
+```
+
+Method : SubscribeToDNSPolicies ( Invoked from agent to SDP via bi-directional stream )
```
bimmlerd, line 65: > Invoked from agent to SDP

Should probably be "invoked from SDP to agent" right?

Otherwise I think this doesn't match how the RPC is specified, and also would mean that the SDP exposes a gRPC API itself, which seems unnecessary.

cilium/CFP-30984-dns-proxy-ha-v2.md
```diff
@@ -0,0 +1,134 @@
...
+message DNSPolicies {
+  repeated DNSPolicy egress_l7_dns_policy = 1;
+  string request_id = 2; // Random UUID based identifier which will be referenced in ACKs
+}
+
+```
+
+*Note: `dns_pattern` follows the same convention used in CNPs. See https://docs.cilium.io/en/stable/security/policy/language/#dns-based for more details*
+
+*Note: `DNSPolicies` is a snapshot of the latest known policy information for all endpoints on the host. Sending a snapshot allows for dealing with deletions automatically*
```
bimmlerd, line 91: Do you intend to send a full snapshot each time? It seems possible to match k8s with a `ListAndWatch` approach to avoid sending a full snapshot.

bimmlerd (commented):
Apologies for the late comments. As Joe said, let's discuss more on an actual PR - please do include me on review.

cilium/CFP-30984-dns-proxy-ha-v2.md
tamilmani1989: that's correct. its invoked by SDP to agent and that opens up a stream which used by Cilium Agent to send updates

cilium/CFP-30984-dns-proxy-ha-v2.md
vipul-21: Yes, we intend to send the full snapshot of the policies applied. Not familiar with the `ListAndWatch` approach, just curious if that will require permissions for SDP to read the CRD ? Then SDP will map the policy rules (labels<>identity) too ?

joestringer:
From community meeting: @hemanthmalla to update the CFP based on the latest PR changes.

PR #57, opened 2024-09-03
CFP-32820: Improve Scalability of Health Probing

jshr-w:
Add CFP for [cilium issue #32820](https://github.com/cilium/cilium/issues/32820).
[Draft CFP](https://docs.google.com/document/d/102NC62FENIza27DzqMpFC33bXqR-h2nKmAi4XG2D6oQ/edit?usp=sharing)

### DIFF
--- a/cilium/CFP-32820-improve-scalability-of-health-probing.md
+++ b/cilium/CFP-32820-improve-scalability-of-health-probing.md
@@ -1,0 +1,168 @@
+# CFP-32820: Improve Scalability of Health Probing
+
+**SIG: SIG-Scalability**
+
+**Begin Design Discussion:** 2024-08-12
+
+**Cilium Release:** 1.17
+
+**Authors:** Shreya Jayaraman <shjayaraman@microsoft.com> based on the work of Marcel Zięba <marcel.zieba@isovalent.com> and feedback from Joe Stringer <joe@cilium.io>
+
+**Status:** Implementable
+
+## Summary
+
+Cilium currently recommends disabling ICMP and HTTP health probes for large-scale clusters due to their impacts on cpu/mem usage. We propose to dynamically set the probing interval and thereby provide a configurable number of queries per second issued by probers to reduce the effects of ping frequency at scale.
+
+## Motivation
+
+Currently, Cilium uses runs ICMP and HTTP probes at intervals of 60 seconds, which may be an issue in large-scale clusters and is why Cilium currently [recommends that users disable health checking](https://docs.cilium.io/en/stable/operations/performance/scalability/report/) in such clusters.
+
+There is a trade-off between resource consumption (for e.g., cpu/mem usage on the cluster and network bandwidth) and the frequency of updates regarding connectivity status. The method of health probing proposed here intends to allow Cilium users to determine which is more important to them.
+
+### User Journeys
+
+There are two conflicting user profiles to testing cluster connectivity which this CFP aims to address:
+
+1. As a cluster administrator, I have a large cluster and want to ensure that the network bandwidth consumed for validating basic connectivity across nodes is well-bounded, even if it means that Cilium can less quickly tell when there is a connectivity breakage on the underlying network.
+
+2. As a cluster administrator, I want to be able to quickly tell whether basic connectivity is OK across nodes, potentially even at the expense of network bandwidth usage within the environment.
+
+This proposal intends to provide users with a way to find a balance between probe frequency and resource use that reflects their needs.
+
+## Background
+
+In the current health probing system, there are ICMP and HTTP probes configured to probe nodes and health endpoints.
+
+### ICMP Probe
+
+The ICMP probe currently in-use is the [servak/go-fastping](https://github.com/servak/go-fastping) library. This probe sends a burst of pings to all the node IPs and health endpoint IPs with intervals of MaxRTT seconds, where MaxRTT is currently set to [60 seconds](https://github.com/cilium/cilium/blob/6186d579ed60f334c7a4daaf81060797b02cc6bd/cilium-health/launch/launcher.go#L37).
+
+### HTTP Probe
+
+The HTTP probe also runs at intervals of 60 seconds. In every loop, the probe will send an HTTP GET request to the /hello endpoint of every IP [known](https://github.com/cilium/cilium/blob/6186d579ed60f334c7a4daaf81060797b02cc6bd/pkg/health/server/prober.go#L260) to the probe.
+
+As every node in the cluster or clustermesh probes every other node and their health endpoints, probing is in the order of O(n^2). Further, the ICMP probe bursts pings to every node and the HTTP probe loops and probes a health endpoint on each node.
+
+## Proposal
+
+As an alternative, we propose the following updated design for health checking.
+
+### Configurable and Auto-Scaling Probing Interval
+
+First, the default probing interval will be adjusted so that it is no longer a constant 60 seconds and instead, a value that is dependent on cluster size. For example, for a cluster of size `n`, i.e., with `n` discovered nodes, the probing interval will be set to `N = ClusterSizeDependentInterval(n, k)`, which returns a value dependent on the cluster size according to the [function](https://github.com/cilium/cilium/blob/6186d579ed60f334c7a4daaf81060797b02cc6bd/pkg/backoff/backoff.go#L127) `k * ln(1+n)`.
+
+The value of `k` represents the base interval value and will be configurable, to a value between 10 and 110.
+
+For reference, the following table shows a back-of-the-napkin estimate for the probes per second and bandwidth consumed by ICMP pings, assuming a standard ICMP packet size of 74 bytes.
+
+| Cluster Size `n` | Base Interval `k` | Probe Interval `i=k*ln(1+n)` (s)| ICMP Events per Second `n^2 / i` | Estimate for Bandwidth for ICMP Pings (Kbps) |
+| ------------ | ------------- | ------------------ | ----------------- | -------------------------------------------- |
+| 100 | 10 | 46 | 217 | 128 |
+| 1000 | 10 | 69 | 14474 | 8568 |
+| 5000 | 10 | 85 | 293517 | 173762 |
+| 100 | 60 | 276 | 36 | 21 |
+| 1000 | 60 | 414 | 2412 | 1428 |
+| 5000 | 60 | 511 | 48920 | 28960 |
+| 100 | 110 | 508 | 20 | 12 |
+| 1000 | 110 | 759 | 1316 | 779 |
+| 5000 | 110 | 937 | 26683 | 15796 |
+
+The minimum value of `k=10` is similar to the current interval, and may be favored by users who require health probe metrics to be updated more frequently or have smaller clusters. On the other hand, setting `k=110` will cause very infrequent probing, which could benefit users who don't require such frequent metrics updates or who have larger clusters.
+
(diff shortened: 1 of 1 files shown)


cilium/CFP-32820-improve-scalability-of-health-probing.md
```diff
@@ -0,0 +1,149 @@
...
+
+**Cilium Release:** 1.17
+
+**Authors:** Shreya Jayaraman <shjayaraman@microsoft.com> based on the work of Marcel Zięba <marcel.zieba@isovalent.com>
+
+## Summary
+
+Cilium currently recommends disabling ICMP and HTTP health probes for large-scale clusters due to their impacts on cpu/mem usage. We propose to dynamically set the probing interval and thereby provide a configurable number of queries per second issued by probers to reduce the effects of ping frequency at scale.
+
+## Motivation
```
joestringer, line 15: I think the motivation section here seems more like background. Maybe that could be renamed.

However, for motivation, I think that the "user journeys" that I highlighted in the original google doc would provide pretty good motivation for the changes:

- "As a cluster administrator, I have a large cluster and want to ensure that the network bandwidth consumed for validating basic connectivity across nodes is well-bounded, even if it means that Cilium can less quickly tell when there is a connectivity breakage on the underlying network."
- "As a cluster administrator, I want to be able to quickly tell whether basic connectivity is OK across nodes, potentially even at the expense of network bandwidth usage within the environment".

joestringer: If we document both of these user journeys as motivations, then I think that helps to motivate why Option 1 (ratio) for configuration can help a wider range of users to tweak this feature to the demands of their environment.

joestringer (commented):
I have a couple of minor nits noted below with some suggested changes, then a few discussion points, none of which are blocking on the design. I think there's only one discussion point that has a particularly meaningful impact, which would be exactly how you intend for probe failures to impact the overall probing interval. Maybe you already have an idea in mind for that, but I didn't read the solution in the current text.

I'm fine with the ratio expression of the setting (Option 1). Unless @cilium/sig-scalability (@marseel?) have strong opinions otherwise, I think we can work under the assumption that this is the solution.

cilium/CFP-32820-improve-scalability-of-health-probing.md
```diff
@@ -0,0 +1,149 @@
...
+
+#### Metrics Update Period
+
+As a consequence of this proposal, health probe updates may take longer and be more unpredictably-timed for users. In order to keep users aware of how long it will take to receive accurate metrics for the known nodes in the cluster, we can update the output of `cilium-health status` to also report the current probe interval and/or the most recent timestamp at which metrics for all known nodes were updated.
+
+#### Probe Timeouts and Errors
+
+The HTTP probe is currently set to have a timeout of 30 seconds. The ICMP probe has MaxRTT set to 60 seconds, which acts as the timeout for the pings. Both timeouts seem larger than necessary. The HTTP probe timeout can be re-configured to 10 seconds, which is a reasonable timeout and accounts for the sizes of the maximum bucket configured for the histogram metric `node_health_connectivity_latency_seconds`.
```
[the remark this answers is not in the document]
joestringer: Maybe let me turn this feedback upside-down: I think it would be a non-goal for 5,000 nodes with no connectivity to take 50,000s (half a day) to ascertain that no other node is reachable. The probing should complete in around the target probing interval, +/- a few seconds or so.

I _suspect_ that there may be some optimizations to be made around exactly how socket and goroutine management work here, but I think that's also probably too much detail to lock down in the design here. We'll figure out what makes sense in that regard as you develop the implementation.

joestringer (changes requested):
Just to clarify the status from the above comments:
- Please apply the "suggestions" where they make sense, or adjust that wording to clarify the points I raised.
- It would be good to include the user journeys as motivating user-focused cases for the CFP.
- All my other feedback is informational-only for discussion purposes. If you want to update the text based on those discussions, feel free, but I don't consider them to be blocking.

cilium/CFP-32820-improve-scalability-of-health-probing.md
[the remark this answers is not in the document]
jshr-w: Ah, I was not envisioning taking 10s per node for sure 😅 The second option is more what I was thinking, with ~5.3 pings/second. The probing interval should not be significantly affected by failures -- it should at most be `probingInterval+maxRTT` in the case that some of the last probes sent out fail. I will keep in mind to think about how to best optimize the implementation here.

cilium/CFP-32820-improve-scalability-of-health-probing.md
```diff
@@ -0,0 +1,149 @@
...
+
+However, for the ICMP probe, by nature of the library `servak/go-fastping`, the probe interval value and max RTT cannot easily be separated. This is one reason to consider changing to the library `prometheus-community/pro-bing`, which will allow us to easily configure different Timeout and Interval values, allowing us to set the timeout to 10 seconds and the interval to the probe interval value.
+
+#### Synchronous Probe Behavior
+Currently, a synchronous probe is run in two situations:
+1. When the health server starts running
+2. When an HTTP GET request is made to the health server for /status/probe, usually by a call to `cilium-health status --probe`
```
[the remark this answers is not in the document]
jshr-w: This makes sense to me, thanks for sharing your perspective! I agree that it feels like it may defeat the purpose of this CFP a bit. :) I will adjust the document to intend to deprecate this synchronous probe for now.

cilium/CFP-32820-improve-scalability-of-health-probing.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Key Questions
+
+### Key Question:
+
+What is the preferable approach to configuring the QPS?
+
+### Option 1:
+
+Configure the flag `--health-probe-frequency-ratio` described in this proposal.
```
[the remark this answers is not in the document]
jshr-w: I see what you mean, makes sense to me

cilium/CFP-32820-improve-scalability-of-health-probing.md
jshr-w: This makes a lot of sense to me, thank you for the feedback!

cilium/CFP-32820-improve-scalability-of-health-probing.md
```diff
@@ -0,0 +1,149 @@
...
+
+The value of `k` will be indirectly configurable by the user, using the formula `k = 10 +  health-probe-frequency-ratio * 100`, where `--health-probe-frequency-ratio` is a cilium-agent flag that is a float in the interval `[0,1]`. 0 will give the most frequent probing and 1 will give the least frequent probing. Note that due to the function used, it is safe for the user to pass in a value of 0 or 1. We will set the default value of health-probe-frequency-ratio to be 0.5 to balance probing frequency with CPU and memory usage. (Note: during implementation, some testing can be done to ensure these initially proposed values are sensible for most supported cluster sizes.)
+
+In the current design, the health server discovers nodes in the cluster after the probe interval passes. This design should be updated so that every time the server gets the nodes in the cluster, the probe interval is also updated using the updated cluster size. Note that the updated probe interval should not affect the current sleep duration, but only the next one, i.e., the health server will:
+
+```
+server.runActiveServices:
+	Get the current node set, run synchronous probe, update metrics.
```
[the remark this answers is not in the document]
jshr-w: Good point, I agree that we can work this into the implementation.

cilium/CFP-32820-improve-scalability-of-health-probing.md
```diff
@@ -0,0 +1,158 @@
...
+
+We will plan to deprecate the capability to synchronously probe the health server via the `cilium-health` CLI command. This will reduce the negative effects of synchronous probing on resource usage at high-scale. On start-up, probing will rely on the probe interval determined by cluster size, as in the typical case.
+
+## Key Questions
+
+### Key Question:
+
+What is the preferable approach to configuring the QPS?
+
+### Option 1:
```
joestringer, line 122:
```suggestion
### Option 1 (Selected):
```

cilium/CFP-32820-improve-scalability-of-health-probing.md
```diff
@@ -0,0 +1,158 @@
...
+#### Pros
+
+* This flag abstracts away decision-making regarding the base interval from the user, and instead allows the user to make a decision regarding the trade-off between probe frequency and resource usage with a well-defined range of possible values and ease-of-understanding of a ratio between 0 and 1.
+
+#### Cons
+
+* The obfuscation may be more complex to interpret and document.
+* Removes some direct configuration power from user.
+
+### Option 2:
```
joestringer, line 135:
```suggestion
### Option 2 (Discarded):
```

cilium/CFP-32820-improve-scalability-of-health-probing.md
```diff
@@ -0,0 +1,158 @@
...
+
+#### Pros
+
+* Simpler to communicate to user.
+
+#### Cons
+
+* May be less intuitive for users to understand the impact of a particular base interval on the probing interval and cluster size.
+
+### Option 3:
```
joestringer, line 147:
```suggestion
### Option 3 (Discarded):
```

cilium/CFP-32820-improve-scalability-of-health-probing.md
joestringer: Maybe there's a better way to mark this but I think it makes sense to capture that these other options were considered but it seems like they're not the direction we want to go.

cilium/CFP-32820-improve-scalability-of-health-probing.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Motivation
+
+Currently, Cilium uses runs ICMP and HTTP probes at intervals of 60 seconds, which may be an issue in large-scale clusters and is why Cilium currently [recommends that users disable health checking](https://docs.cilium.io/en/stable/operations/performance/scalability/report/) in such clusters.
+
+There is a trade-off between resource consumption (for e.g., cpu/mem usage on the cluster and network bandwidth) and the frequency of updates regarding connectivity status. The method of health probing proposed here intends to allow Cilium users to determine which is more important to them. For instance, a cluster administrator for a large cluster who wants to ensure health checking does not have significant effects on bandwidth or resource use may find it reasonable to have the health probes run less frequently, even if the connectivity data is provided less often. On the other hand, a cluster administrator may want to have connectivity statuses updated more frequently, despite its negative effects. This proposal intends to provide users with a way to find a balance between probe frequency and resource use that reflects their needs.
+
```
joestringer, line 22: Does it make sense to add these here?

```suggestion

### User Journeys

There are two conflicting user profiles to testing cluster connectivity which this CFP aims to address:

1. As a cluster administrator, I have a large cluster and want to ensure that the network bandwidth consumed for validating basic connectivity across nodes is well-bounded, even if it means that Cilium can less quickly tell when there is a connectivity breakage on the underlying network.
1. As a cluster administrator, I want to be able to quickly tell whether basic connectivity is OK across nodes, potentially even at the expense of network bandwidth usage within the environment.
```

(Ideally we would hear these directly from users rather than me making them up, but I think that these opinions sound reasonable and they motivate 'why' to pick Option 1 as the solution).

joestringer (approved):
Apologies if this was waiting on me for a while. I think the CFP makes sense. I added some minor suggested changes below as well if you think those make sense.

cilium/CFP-32820-improve-scalability-of-health-probing.md
jshr-w: Makes sense to me :) I've pulled out the user journey statements from the paragraph to this more explicit form.

cilium/CFP-32820-improve-scalability-of-health-probing.md
jshr-w: Updated to clarify decisions made for each option

joestringer (approved):

PR #59, opened 2024-09-13
CFP-27752: Operator Manages Cilium Identities

dlapcevic:
CFP google doc: https://docs.google.com/document/d/1Hcc_2mB9OOUxrqQgZ-gSYDPnLYE_If_TCzVbUGDOdGM
The CFP google doc is outdated, and the latest state of the CFP is fully reflected only in this CFP PR.

More discussion can be found in: https://github.com/cilium/design-cfps/pull/19


dlapcevic:
cc @joestringer

dlapcevic:
**Status update**

The main code changes for operator ID management are merged.

Next steps:
* Testing - before we can expect users to adopt operator ID management in 1.17, we want to add more tests for it. We need to check if there are existing e2e tests that create a cluster and verify that pods are running and are able to communicate between each other while adhering to existing network policies. We would either want to use the existing test or create a new test, and enable operator ID management on it.
* Move CID GC from legacy (heartbeat) to the new CID controller.  https://github.com/cilium/cilium/issues/34740
* kvstore compatibility - https://github.com/cilium/cilium/issues/34865

dlapcevic:
Copy of @tamilmani1989's questions from Cilium Slack about operator ID management.

>  1. As operator starts creating cilium identities and then associate with CEP, does cilium agent watch only on CEP/CES to receive updates and not required to watch identity resource?

With current implementation, cilium-agent still needs to watch CID (CiliumIdentity). The way it works is that cilium-operator creates CIDs based on pods and namespaces, and then cilium-agent can assign security identity to CEP when pods are created.

Few notes:
- CID might not be needed when CESs are not used, because CEP objects already have security ID and labels. However, this is an inefficiency, which is solved in CESs, where core endpoints within CES are containing only an integer security ID and not labels (which can be very large).
- With the previous point in mind, we’re moving in a direction to remove CEP, and keep only CES, where cilium-operator is generating both CID and CES, and doesn’t require any additional input. It can use pod and namespace information.

>  2. Today identities are not created for pods that are running in unmanaged node (no cilium agent running). Is operator going to identify if pod scheduled on managed/unmanaged node and create identity only for managed nodes?

This is a very good question to raise. I created an issue for it. We want cilium-operator to create CIDs only for pods that are managed by Cilium.

New issue: https://github.com/cilium/cilium/issues/35402

>  3. I guess cilium agent watch on local pods on that node to track label changes. With this feature, is it required for agent to watch on pods?

Watching local pods is required for reasons more than just accessing pod labels. There is HostPort and other functionalities. It would be possible to redesign the network policy enforcement system to not require getting pod labels from the Pod object, but to get it from CES, but there won’t be a lot of benefits there. Local pod watchers are not very costly.

>  4. Since it gets rid of identity heartbeat, How does operator track identity has no endpoints and cleanup identity?

The initial implementation is just completed and it still uses the legacy CID GC. This is because the operator managing CIDs is a big architectural change and we wanted to have smaller steps. We’re planning to move CID GC to the new CID controller https://github.com/cilium/cilium/issues/34740.

Operator is keeping data about which pods use which CIDs, and how many use CIDs. At the same time it has access to how CIDs are used in CEPs and CESs. When operator notices that no CIDs are used in Pods, CEPs and CESs, then it will clean them up.

tamilmani1989:
> then cilium-agent can assign security identity to CEP when pods are created

1. In doc, It states operator updates CEP `status.Identity` field. should we update that? 
2. Is there any reason you switched to agent updating identity in CEP instead of operator?
3. how does agent associate CEP with CID when it receives update from operator. CID object just contains labels? 

Apologies to trouble with more questions. would be great if we update doc with this information.

> With the previous point in mind, we’re moving in a direction to remove CEP, and keep only CES, where cilium-operator is generating both CID and CES, and doesn’t require any additional input. It can use pod and namespace information.

Did you mean agent would not create CEP at all?

> Operator is keeping data about which pods use which CIDs, and how many use CIDs. At the same time it has access to how CIDs are used in CEPs and CESs. When operator notices that no CIDs are used in Pods, CEPs and CESs, then it will clean them up.

Why can't operator just track CID -> CEP mappings and if no CEP found for CID, then it can GC. Why does it need to track pod->cid mappings. anyway pod to CEP is 1:1

dlapcevic:
> Apologies to trouble with more questions. would be great if we update doc with this information.

The doc is outdated. I added to the CFP doc and this CFP PR that the doc is outdated and that this CFP PR contains up to date information.

> 2. Is there any reason you switched to agent updating identity in CEP instead of operator?

Yes, to avoid operator writing to CEP. It allowed us to go with the simpler implementation of operator CID management and also move into a direction to remove CEP. Operator is creating CIDs from pods and namespaces, and doesn't need CEP.

> 3. how does agent associate CEP with CID when it receives update from operator. CID object just contains labels?

cilium-agent without operator CID management gets or creates a CID depending if it exists, when a pod/CEP is created or updated.
With operator CID management it only tries to get CID from watcher, but doesn't try to create it. Take a look at the PR https://github.com/cilium/cilium/pull/34867.

> Did you mean agent would not create CEP at all? If so, what CES will contain then?

Yes. CES will be created from pods instead of CEPs. We have all needed information in pods and namespaces. This would not be possible without the new CID controller (added with operator CID management), because it contains information about which CID is assigned to which pod.

We already started working on creating CES from pod instead of CEP: https://github.com/cilium/cilium/pull/34784

> Why can't operator just track CID -> CEP mappings and if no CEP found for CID, then it can GC. Why does it need to track pod->cid mappings. anyway pod to CEP is 1:1

Operator creates CIDs from pod and namespace data and naturally tracks CID <-> Pod mapping. It doesn't need to depend on CEP/CES, but it can use it to determine if a CID needs to be deleted -- if CES is enabled then looks in CES, otherwise in CEP.

Note: operator still uses legacy CID GC.

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+    * Handling unplanned CID changes (create, update, delete) by other sources -- not Operator or Agent.
+
+2) Operator will modify the following resources:
+
+* CIDs - Create and update. If namespace or pod labels are changed, new CIDs will be created for all unique label sets that don’t match any existing identity. Unused CIDs will be cleaned up by the CID GC. In case of an update to a CID, the CID controller will update the CID back to the correct state, effectively reverting the initial change.
+
+3) CID GC:
+
+* In the initial release, CID controller will not delete CIDs. It will still rely on the legacy CID GC (the periodic GC) to deleted unused CIDs. The reason is to reduce the scope of the changes, and allow for simpler and smaller steps in the direction of CID controller fully managing CIDs.
+* In the next phase, CID GC will be moved to CID controller. The benefit is that CID controller will already have information about which CIDs are used and which not, and will be able to delete CIDs in real time, without periodically running GC.
```
joestringer, line 71: nit: I think that the discussion on real-time vs periodic is probably too much detail for this CFP. We may/may not even need a dedicated CFP for that aspect, though I think it would be an interesting discussion to have either way. When I read this, I see an opinionated design without the surrounding assumptions / arguments why this is how this aspect _should_ work. I think it'd be better deferred to a subsequent discussion.
```suggestion
* CID garbage collection continues to be the responsibility of the Operator. In the initial release, this will still rely on the legacy CID GC (the periodic GC) to deleted unused CIDs. The design of this can be evaluated after completing this CFP.
```

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Agent changes
+
+Agents will stop creating CIDs. Agents will only generate local security identities. Agents will no longer perform following actions:
+
+* Locally create and reserve security identities in the global range [256, 65536].
+* Create CIDs.
+* Remove heartbeat status annotation on CIDs (part of CID GC) - In the second phase when CID GC is moved to CID controller.
+* Agents will no longer need the permission to modify CIDs - In the second phase when CID GC is moved to CID controller.
```
joestringer, line 80:
```suggestion

Subsequent milestones will further remove the following from the Agents:
* Reporting heartbeat status annotation on CIDs.
* Permissions to modify CIDs.
```

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+Agents will wait for CIDs to be received through the watcher, and then allocated them to local endpoints (pods) by retrieving the CIDs from the watcher's store based on pod and namespace labels.
```
joestringer, line 82:
```suggestion
Agents will wait for CIDs to be received through the watcher, and then assign them to local endpoints (pods) by retrieving the CIDs from the watcher's store based on pod and namespace labels.
```

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Namespace watch**
+
+Agents will continue to watch namespaces in order to instantly update security identity labels in the local Endpoint manager’s cache. More info below in Label changes: Namespace.
```
joestringer, line 86: I'm curious, why is this necessary? The later discussion states that the Agents will wait until the Operator creates new CIDs anyway.

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Pod startup**
+
+CNI will be blocked if CID doesn't exist in the CID watcher's store. After retries, CNI will fail and pod creation will fail. This is useful to quickly identify issues with CID controller or Operator.
+
+Pod startup latency will be reduced when a CID for pod’s labels doesn’t exist yet. CID creation is time consuming mainly because of waiting for kube-apiserver to create the CID resource in Etcd. With operator managing CIDs, CID creation will happen as soon as Pod object is created, before it's even scheduled. Therefore, Agents will always have CID in the watcher store at the time of pod startup.
```
joestringer, line 92: I don't fully follow the argument of this paragraph, and it seems unnecessary for the design. It's interesting background but I'm unsure how accurate the content is. Should it be dropped?
```suggestion
```
In particular: 

> Agents will always have CID in the watcher store at the time of pod startup.

It's nice if this is true, but this sounds like it's prescribing that the kube-apiserver must wait until cilium-operator creates a CID resource before scheduling the Pod to a Node. Is that happening? Is that part of the design? Either way, the previous paragraph describes blocking CNI if CID doesn't exist, so the section here seems not self-consistent.

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Pod creation process
+
+CID assignment paths on pod creation:
```
joestringer, line 96:
```suggestion
This design does not prescribe that the Kubernetes control plane waits for Cilium operator to create a CID before scheduling Pods. Therefore the operations executed when a Pod is created can be performed in one of two different orders:
```

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+The only problem with simultaneous CID management (both Operator and Agents are managing CIDs) is when namespace labels are changed. The only bad thing that happens is that there is CID duplication. This is an existing problem already, that doesn't cause disruption.
+
+Simultaneous CID management is only expected to happen during upgrade from Agents managing CIDs to Operator managing CIDs.
+
+
+## CID deduplication
+
+Operator will not create duplicate CIDs, but it will allow for their existence. In case other sources, like Agents, are making duplicate CIDs, Operator will be compatible with tracking multiple CIDs. This is needed to make Operator compatible with Agents managing CIDs at the same time.
+
```
joestringer, line 139: Is there a plan for the operator to eventually de-duplicate the CIDs? (doesn't have to be part of this CFP, but it could be mentioned as a future milestone)

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Namespace
+
+When namespace labels change, Operator and Agents get the update at the same time. Agents wait until Operator processes the namespace label change update and creates all the new CIDs. Agents keep retrying to match CIDs for all their local endpoints until they are received through the CID watcher.
+
+
+## Simultaneous CID management compatibility
+
+The only problem with simultaneous CID management (both Operator and Agents are managing CIDs) is when namespace labels are changed. The only bad thing that happens is that there is CID duplication. This is an existing problem already, that doesn't cause disruption.
```
joestringer, line 131: nit: I found this paragraph confusing. I think the same information could be conveyed with simpler text?

```suggestion
Prior to this feature, all agents managed CIDs and operators did not manage CIDs. When upgrading Cilium to enable this feature, we expect users to enable both agents and operators to manage CIDs. As long as Agents are managing CIDs, the CID duplication issue may continue to occur. Once the upgrade is completed, users can disable managing CIDs from the agents.
```

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+Operator will not create duplicate CIDs, but it will allow for their existence. In case other sources, like Agents, are making duplicate CIDs, Operator will be compatible with tracking multiple CIDs. This is needed to make Operator compatible with Agents managing CIDs at the same time.
+
+
+## KVStore compatibility
+
+[KVStore](https://docs.cilium.io/en/stable/kvstore/) is a key-value store that is mostly used for sharing Cilium mappings between IPs, identities (IPCache) and labels, across the cluster ([what's inside Cilium’s KVStore](http://arthurchiao.art/blog/whats-inside-cilium-etcd/)).
+
+Operator managing CIDs is intended to work with both `identity-allocation-mode=kvstore` and `identity-allocation-mode=crd`. CRD being the default, the initial implementation will use the CRD allocation mode, and later expanded to other allocation modes.
+
+All features that rely on using kvstore will not work until CID controller is expanded to work with kvstore allocation mode. The work to enable kvstore compatibility is estimated to of medium size and it will be tracked in a separate GitHub issue (https://github.com/cilium/cilium/issues/34865).
```
joestringer, line 147: What exactly does this mean?

> All features that rely on using kvstore will not work until ...

We can't break kvstore mode for existing users. Is this intended just to mean that Users must not disable agents managing CIDs when using kvstore? Then maybe the text should say this.

Also for the implementation we should ensure that this incompatible configuration is reported to the user to ensure they don't end up breaking their environments.

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+# Enablement
+
+The feature will be guarded by a flag `operator-manages-identities`, that can be set in container arguments in Operator’s and Agent’s manifests. The default values for both will be `false`, meaning that Operator doesn’t manage CIDs, while Agents do.
+
+
+## Long term plan
+
+The long term plan is to make Operator managing CIDs the default, and deprecate the part of Agent that is currently used to manage CIDs.
+
```
joestringer, line 158: I like the intent behind this, but I think that this statement is out-of-scope for the CFP. I'm all for deprecating unneeded infrastructure and getting users onto a single, better, implementation. That said, we are going to need to be more methodical about this to set the criteria. I think it's better to drop this paragraph and maybe we can brainstorm on an issue or subsequent CFP about how to make this happen.

```suggestion
```

I expect we'll be need to:
- Gather feedback from users to evaluate the implementation, whether it sufficiently replaces the existing functionality
- Ensure that the central components don't become a bottleneck or break reliability
- Ensure the feature works with other incompatible features (example: kvstore)
- Ensure we have a clear path for upgrade

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+# Migration
+
+Any situation when neither Operator nor Agents are managing CIDs is effectively a downtime for pod creation. The control plane side of network policies and other features relying on CIDs will not carry out any required changes during this time, causing them to stop working.
```
joestringer, line 162: nit: I'm struggling a bit to understand what the second half of this paragraph is stating. Is there something more than "You can't successfully create Pods while the Operator is down"?

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+
+## Upgrade
+
+Safe upgrade, without downtime, requires simultaneous CID management compatibility.
```
joestringer, line 167: I would suggest breaking upgrade down into the exact upgrade steps you are analysing and describe the expectations. Specifically:

Is this upgrading from a version with Agent managing CID to Operator managing CID? What about upgrading from a cluster where operator manages CID to a newer version where operator manages CID?

The sentence here sounds like it's suggesting that users should re-enable simultaneous CID management during every subsequent upgrade. But I don't think that's what you mean.

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+If we upgrade the Operator from a version that is not managing CIDs to a version that is managing CIDs while Agents are also managing CIDs, there should be no negative impact on the system. Everything should work the same.
+
+
+## Downgrade
+
+Downgrade without downtime is possible only if Agent daemonset is downgraded first -- updated to an earlier version where Agent manages CIDs. Otherwise, if Operator is downgraded first, there will be a downtime lasting at least the amount of time needed for Agent daemonset to be downgraded after the Operator downgrade.
```
joestringer, line 174: I think that the text should be more rigorous about describing the exact steps for upgrades, including which components are managing CIDs, and the intended supported upgrade/downgrade paths.

This paragraph currently seems like it is talking about going from *only* operator managing CIDs to *only* agent, but it seems to me like you don't intend on supporting that path. At the same time, the "upgrade" section above talks about needing to run simultaneous CID management. Presumably if the user is running simultaneous CID management during upgrade, then they will still be running this after upgrade. Maybe then they should run that way for a while and then later on disable agent CID management, once they are sure they will no longer need to downgrade?

We could even facilitate this, for example by saying that v1.17 creates the option for CID management from operator; v1.18 enables it by default with simultaneous management, and v1.19 disables agent CID management by default. We don't have to do it this way, but this is an example.

joestringer: I should also reinforce: If a user cannot downgrade safely, then I generally consider that they can also not upgrade safely. A clean upgrade is reliant on the ability to roll back to a safe older setup if any issue is encountered.

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Operator acts on Pod instead of CEP changes
+
+Status: Included in the original proposal
+
+Performance optimization
+
+Operator doesn't watch for CEPs but watches for Pods instead. As soon as a pod is created (not started) Operator allocates a CID immediately. This way the new CID is getting created in parallel as the pod is getting scheduled.
+
+Downside: In case pods are unschedulable we will be allocating identities even though they might not be used at the end.
```
joestringer, line 201: This should be fairly rare. As long as the operator doesn't leak these but instead performs GC then this should be fine.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Pod creation is the main one, but there are others. I'm not sure if I should list them here, or mentioning that pod creation will fail is sufficient.
- namespace label changes
- changing identity relevant labels filter
- updating deployment labels (it might not be obvious that this in fact redeploy pods)

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: This is only for the case of namespace label changes. Agents anyway already watch namespaces and use the Namespace watcher cache (Store) to retrieve namespace labels. Agents need to know what are the namespace labels for each pod, to be able to match pod+namespace labels to a CID from CID watcher's cache.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Yes, but it's not so simple, unless we want to add this logic to Agents.

Duplicated CIDs are CIDs that have the same label set.

There is logic in the CID controller to pick the first CID that it finds in the CID watcher store that matches a new pod, and always associate all pods with that label set to that single CID. It's already implemented in the method `handleStoreCIDMatch()` - https://github.com/cilium/cilium/blob/main/operator/pkg/ciliumidentity/reconciler.go#L397

However, this doesn’t entirely solve the problem, because as long as multiple CIDs with the same label set exist, Agents will be able to assign any of them to new CEPs. Therefore, it can take a significant amount of time and pod churn to completely de-duplicate CIDs, because it will be random. This is especially true with the legacy CID GC, which is taking 30 minutes to clean up CIDs.

With CID GC enhancements, when it’s moved to CID controller, the delete delay will be much smaller (e.g. 30 seconds), which will make de-duplication more rapid.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Yes. I added that it's a very minor downside.

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,214 @@
...
+
+CIDs are presenting issues with reliability and scalability, and are causing network policies to be in a broken state. Most notable cases:
+
+* Reaching maximum number (65k) of CIDs in the cluster.
+* eBPF policy maps overflowing with over 16k entries.
+* CID garbage collection malfunction.
+* Network policy incorrectly dropping connections on KCP upgrade, when cilium-agent restarts. Related to CID duplication
+* Unhealthy cluster state that is difficult (or impossible) to recover from - related to numbers of IDs (and other Cilium custom resources) and CID duplication.
```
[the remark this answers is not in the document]
dlapcevic: I agree. Done. Thanks.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: I completely agree with you. I think we shouldn't make operator CID management available to the users until we make it compatible with kvstore.

I updated the doc (at the top of the KVStore compatibility chapter) to state that this is a blocker for releasing the feature.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Thanks for the suggestion. It's more clear now. Done.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Applied

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Applied.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Applied.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Applied.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Applied

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Applied

cilium/CFP-27752-operator-manages-cilium-identities.md
joestringer: OK. Reading the current text, it sounded like the motivation is to "instantly update", which implies that if this namespace watcher was removed then the update would just be delayed. However, this response from you implies that this step is required for correct operations:

> Agents need to know what are the namespace labels for each pod, to be able to match pod+namespace labels to a CID from CID watcher's cache.

I'd suggest merging this motivation into this section so future readers can understand that this is a required aspect of the design, not optional.

cilium/CFP-27752-operator-manages-cilium-identities.md
joestringer: I see yeah. I was thinking of somehow having some ordering metric so that agents could listen for CIDs and detect duplicates, then all independently pick one "original" instance of the duplicate CID to switch endpoints over to. Once the agent picks this "original" CID, it would actively change local endpoints over to use that one instead of the duplicate. If all agents independently do this, they could all align on using the same CID in a distributed way, then they would each remove references to the other duplicate CIDs. This would free the duplicates up for release to the pool (without even requiring Pod churn).

Still though this problem is independent and has a bunch of nuance to it so maybe we can tackle that separately from this CFP if you think it's worth pursuing.

cilium/CFP-27752-operator-manages-cilium-identities.md
joestringer: I see yeah, it gets pretty nuanced. I would rather highlight these specific aspects. It'd also be better to avoid wording like `causing them to stop working`, as this implies a failure of some kind but is not very specific about the failure condition. 

Maybe something like,
```suggestion
Any situation when neither Operator nor Agents are managing CIDs is effectively a downtime for assigning new Security Identities. Either of these components must be available and healthy in order to:
- Deploy workloads with a new set of identity-relevant labels
- Update existing workloads with different identity-relevant labels
- Modify the identity-relevant labels filter
- Modify namespace labels and enforce policies for the newly changed labels
```

joestringer: Something else I had in mind at the same time was that if a new instance of an existing deployment is created or a new Pod is created where the Security Identity is already allocated, then it should actually be possible to deploy that Pod without these components running.

cilium/CFP-27752-operator-manages-cilium-identities.md
joestringer: I'm still debating back and forth myself about whether it's worth having an explicit "simultaneous management" stage or not. When thinking about upgrades in general, can be useful to break the upgrade into multiple stages. Each stage can be more easily tested since the target configuration can be explicitly configured for each state rather than relying on when the configuration is applied and which operators/agents were restarted to apply the new configuration. I also think that this would fit more easily with the `rolloutPods` option which automatically restarts Agents when configuration changes.

On the other hand I think that if you are managing your upgrade more carefully in order to upgrade Operator first, it should be totally viable to upgrade directly from CIDs being managed only by the agents to only managed by the operator.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Yes, you're right. It would be possible to deploy new pods that have labels that match an existing CID.

The blast radius is much clearer the way you phrased it.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Added.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: I think better tackle it separately from this CFP.

I added to the paragraph the following:

Completely deduplicating CIDs could be done with additional changes in Agent. One option is to have Agents always pick CIDs with the lowest integer value in the name, that matches the label set. This way all Agents would always pick the same CID for the same label set. However, this is only a suggestion, and a proper solution will be covered separately, outside this CFP.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: It’s worth putting more thought and time into converging on a solution.

To me it makes sense to do the following:
- Have two configuration flags, one for Operator and one for Agent, to allow more control.
- Treat simultaneous CID management as a transitional state, because that’s not where we want to end up. Although it will be a fully valid state, that doesn’t degrade user experience (any of the points mentioned in your [comment above](https://github.com/cilium/design-cfps/pull/59#discussion_r1812908975)).
- Define an upgrade path, where the Operator’s configuration flag should be always enabled first, before the Agent’s configuration flag is disabled. It will allow users to stay in the transitional state.
- Define a downgrade path, where the Agent’s configuration flag should be always enabled again, before the Operator’s configuration flag is disabled.

We could also consider making Agents manage CIDs whenever Agent’s flag is enabled OR Operator’s flag is disabled, to ensure that there is always at least one of them managing CIDs.

joestringer: SGTM for most of the above.

> We could also consider making Agents manage CIDs whenever Agent’s flag is enabled OR Operator’s flag is disabled, to ensure that there is always at least one of them managing CIDs.

I see where you're coming from, but I think this will be very counter-intuitive and confusing for users. We can complain if both flags are disabled, but I wouldn't enable functionality based on another flag being disabled.

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,220 @@
...
+[KVStore](https://docs.cilium.io/en/stable/kvstore/) is a key-value store that is mostly used for sharing Cilium mappings between IPs, identities (IPCache) and labels, across the cluster ([what's inside Cilium’s KVStore](http://arthurchiao.art/blog/whats-inside-cilium-etcd/)).
+
+Operator managing CIDs is intended to work with both `identity-allocation-mode=kvstore` and `identity-allocation-mode=crd`. CRD being the default, the initial implementation will use the CRD allocation mode, and later expanded to other allocation modes.
+
+All features that rely on using kvstore will not work until CID controller is expanded to work with kvstore allocation mode. The work to enable kvstore compatibility is estimated to of medium size and it will be tracked in a separate GitHub issue (https://github.com/cilium/cilium/issues/34865).
+
+
+# Enablement
+
+The feature will be guarded by a flag `operator-manages-identities`, that can be set in container arguments in Operator’s and Agent’s manifests. The default values for both will be `false`, meaning that Operator doesn’t manage CIDs, while Agents do.
```
joestringer, line 151: For the record both the agent and operator read the cilium-config config map for configuration. Typically for options shared by the two components we set a single flag value and rely on both components reading the same flag.

If we intend for this to be managed differently for agent vs. operator then we probably want to come up with two separate flags, or alternatively a single flag that represents all possible configurations. For instance:

- `--identity-management-mode=agent`
- `--identity-management-mode=operator`
- `--identity-management-mode=both`

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,220 @@
...
+- Update existing workloads with different identity-relevant labels
+- Modify the identity-relevant labels filter
+- Modify namespace labels and enforce policies for the newly changed labels
+
+
+## Upgrade
+
+**From Agents managing CIDs to Operator managing CIDs”**
+
+Safe upgrade, without downtime, requires simultaneous CID management compatibility. The CID controller in Operator is designed to support simultaneous CID management and won’t interfere with Agents.
```
joestringer, line 168: Let's update this based on the latest discussion on upgrade. For instance, "safe upgrade, without downtime" is vague and doesn't directly reference what kind of downtime / impact the upgrade may have.

joestringer (changes requested):
I agree with the results of the discussions above. Let's just apply that into the text, then I think this should be good to go (?) . Please ping on a thread with specific questions if there's something outstanding that I missed.

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,220 @@
...
+
+Performance optimization
+
+Operator doesn't watch for CEPs but watches for Pods instead. As soon as a pod is created (not started) Operator allocates a CID immediately. This way the new CID is getting created in parallel as the pod is getting scheduled.
+
+Downside: This is a very minor downside and isn't expected to cause any problems. In case pods are unschedulable we will be allocating identities even though they might not be used at the end.
+
+## Deprecate CEP
+
+This topic entails the possibility of removing CEPs altogether, because we can just have CIDs and CESs. Agents don’t create CEPs anymore, just watch for CESs and CIDs. Operator creates both CIDs and CESs with all info gathered from pods.
```
tamilmani1989, line 211: is this planned for initial release or for future?

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,220 @@
...
+
+Agents will wait for CIDs to be received through the watcher, and then assign them to local endpoints (pods) by retrieving the CIDs from the watcher's store based on pod and namespace labels.
+
+**Namespace watch**
+
+Agents need to know what are the namespace labels for each pod, to be able to match pod+namespace labels to a CID from CID watcher's cache. Agents will continue to watch namespaces in order to instantly update security identity labels in the local Endpoint manager’s cache, and start attempting to get CID from CID watcher store, to update the corresponding endpoint. More info below in Label changes: Namespace.
+
+**Pod startup**
+
+CNI will be blocked if CID doesn't exist in the CID watcher's store. After retries, CNI will fail and pod creation will fail. This is useful to quickly identify issues with CID controller or Operator.
```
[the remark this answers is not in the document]
joestringer: If this is a strong pain point, I think we can explore improvements in this area as a subsequent milestone. Early versions of the CFP had a lot of complexity to try to solve this problem and we can't even quantify it yet because the overall architecture is not in place.

I would also challenge the idea that Security Identities are constantly changing. If Security Identities are relatively static or preallocated then pod startup time should not be heavily affected.

cilium/CFP-27752-operator-manages-cilium-identities.md
joestringer: This milestone is under a section called "Future enhancements". I consider this to be subject to future discussion. None of the items under this section are considered "Implementable" per the intended status of the CFP, these are ideas that can help guide this CFP and may be designed/implemented in future.

cilium/CFP-27752-operator-manages-cilium-identities.md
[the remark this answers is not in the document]
tamilmani1989: pod startup time is critical metric for us and wanted to see if it has any real impact.

> we can't even quantify it yet because the overall architecture is not in place.

Agreed

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,220 @@
...
+Operator will start managing CIDs, and will become the only system pod that issues mutating requests for CIDs to the kube-apiserver. A new controller (CID controller) will run inside Operator that will be responsible for managing (creating) CIDs.
+
+1) Operator requires watching the following resources to allocate CIDs:
+
+* Pods (Already watches) - To manage CIDs based on pod labels.
+* Namespaces - to manage CIDs based on namespace labels changes.
+* CIDs - to keep the desired state of CIDs. Required for:
+    * Getting the initial state on start-up.
+    * Being compatible with Agents managing CIDs simultaneously.
+    * Handling unplanned CID changes (create, update, delete) by other sources -- not Operator or Agent.
```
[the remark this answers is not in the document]
dlapcevic: Operator will immediately reconcile the CID to match the desired state:
for modify: updating it back to the correct API object state.
for delete: recreating it

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,220 @@
...
+Agents will stop creating CIDs. Agents will only generate local security identities. Agents will no longer perform following actions:
+
+* Locally create and reserve security identities in the global range [256, 65536].
+* Create CIDs.
+
+Subsequent milestones will further remove the following from the Agents:
+* Reporting heartbeat status annotation on CIDs.
+* Permissions to modify CIDs.
+
+Agents will wait for CIDs to be received through the watcher, and then assign them to local endpoints (pods) by retrieving the CIDs from the watcher's store based on pod and namespace labels.
```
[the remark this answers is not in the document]
dlapcevic: No, it's not new. The difference is that Agents currently don't wait for CID to be received through the watcher (if it doesn't already exist there), but attempts to create a new one.

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,220 @@
...
+
+**Namespace watch**
+
+Agents need to know what are the namespace labels for each pod, to be able to match pod+namespace labels to a CID from CID watcher's cache. Agents will continue to watch namespaces in order to instantly update security identity labels in the local Endpoint manager’s cache, and start attempting to get CID from CID watcher store, to update the corresponding endpoint. More info below in Label changes: Namespace.
```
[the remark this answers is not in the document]
dlapcevic: No, it doesn't, because when pod labels change, the pods are recreated, and when namespace labels change, the pods are not recreated.

When pods are recreated, they follow the pod creation path -- CNI, Endpoint Create, AllocateIdentity

cilium/CFP-27752-operator-manages-cilium-identities.md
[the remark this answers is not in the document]
dlapcevic: Pod startup latency can be increased, but only if there is an issue with Operator.
Otherwise, pod scheduling depends on kube-apiserver being healthy, and if kube-apiserver is healthy, Operator will be able to create CID before the pod even gets scheduled (Operator creates CIDs on Pod object creation).

This means that pod startup latency will be decreased, because Agents will no longer need to create CID (request to kube-apiserver), but just read it locally from the watcher's store.

Keep in mind, the difference is only when a CID doesn't exist for the label set for a new pod.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Thanks for the suggestion. I added it to the CFP.

Related changes and discussion about enablement is here: https://github.com/cilium/cilium/pull/36411

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: I updated the CFP based on our discussion. I described the type of disruptions and which ones can occur in which situations. Both for upgrade and downgrade.

cilium/CFP-27752-operator-manages-cilium-identities.md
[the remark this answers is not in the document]
tamilmani1989: > Pod startup latency can be increased, but only if there is an issue with Operator.

Thanks @dlapcevic. I think it depends on how much operator can scale. Operator can become single point of failure which may potentially delay pod creation in every node. I have one follow-up question. Does this model allow multiple cilium operator working together (like active-active) or its more of primary/secondary model (like replicas used only as backups)?

cilium/CFP-27752-operator-manages-cilium-identities.md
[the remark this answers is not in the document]
tamilmani1989: how will it detect if cid not updated by agent and by some other source?

cilium/CFP-27752-operator-manages-cilium-identities.md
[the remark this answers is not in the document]
tamilmani1989: i guess former otherwise it would again endup in duplicate identity creation. this is a trade-off for going with centralized management

cilium/CFP-27752-operator-manages-cilium-identities.md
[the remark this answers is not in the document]
dlapcevic: Agent doesn't update CIDs. If labels change, it creates a new CID. It's required to keep it secure.

cilium/CFP-27752-operator-manages-cilium-identities.md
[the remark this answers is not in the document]
dlapcevic: Right. Operator is running as a deployment (multi-pods), but has only one leader that actually does all the work. Multiple pods are just for HA.

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,225 @@
...
+CIDs are presenting issues with reliability and scalability, and are causing network policies to be in a broken state. Most notable cases:
+
+* Reaching maximum number (65k) of CIDs in the cluster.
+* eBPF policy maps overflowing with over 16k entries.
+
+# Design
+
+## Operator changes
+
+Operator will start managing CIDs, and will become the only system pod that issues mutating requests for CIDs to the kube-apiserver. A new controller (CID controller) will run inside Operator that will be responsible for managing (creating) CIDs.
```
joestringer, line 48:
```suggestion
Operator will start managing CIDs, and will become the only system pod that issues creation or deletion requests for CIDs to the kube-apiserver. A new controller (CID controller) will run inside Operator that will be responsible for managing (creating) CIDs.
```

I mean to clarify here that these resources are intended to be immutable and never to be updated; mutation in this context is ambiguous.

joestringer: (I see that later text reflects this immutable understanding and even implements safeguards to undo externally-sourced mutations)

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,225 @@
...
+
+Subsequent milestones will further remove the following from the Agents:
+* Reporting heartbeat status annotation on CIDs.
+* Permissions to modify CIDs.
+
+Agents will wait for CIDs to be received through the watcher, and then assign them to local endpoints (pods) by retrieving the CIDs from the watcher's store based on pod and namespace labels.
+
+**Namespace watch**
+
+Agents need to know what are the namespace labels for each pod, to be able to match pod+namespace labels to a CID from CID watcher's cache. Agents will continue to watch namespaces in order to instantly update security identity labels in the local Endpoint manager’s cache, and start attempting to get CID from CID watcher store, to update the corresponding endpoint. More info below in Label changes: Namespace.
```
joestringer, line 82: > Agents will continue to watch namespaces in order to instantly update security identity labels in the local Endpoint manager’s cache, and start attempting to get CID from CID watcher store, to update the corresponding endpoint

Why is this necessary? Isn't the agent already listening for CIDs? And won't the new CID be a different CID anyway? How would the agent know which identity number to look for?

joestringer: I guess the answer is that the agent receives the namespace change signal to identify which endpoints must be updated, then logically it can put those endpoints into a "waiting-for-identity" mode which then needs to be correlated with a subsequent CID creation event to actually assign the new identity to the endpoint and close the loop. :+1:

joestringer: > start attempting to get CID from CID watcher store

This part seems like a no-op to me. The agent should already be listening for CIDs, and it wouldn't know which identity number to look for. There just needs to be a new callback on CID update events when there are endpoints looking for an updated identity, then it needs to match the updated identity with the new desired labels for the endpoint and only update the endpoint/CEP if there is a new identity that matches the pod+namespace labels.

cilium/CFP-27752-operator-manages-cilium-identities.md
```diff
@@ -0,0 +1,225 @@
+# CFP-27752: Operator Manages Cilium Identities
+
+**Authors:** dordel@google.com
+
+**Area:** SIG: SIG-POLICY, SIG-AGENT, SIG-K8S
+
+**Target Cilium Release**: 1.17
```
joestringer, line 7:
```suggestion
**Target Cilium Release**: 1.18
```

joestringer: (Though this may still be optimistic with the kvstore compatibility)

joestringer (approved):
Minor nits below but I agree this is implementable.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: This paragraph just states how it will work, not necessarily changing the namespace watcher or how labels are updated in local Endpoints. Basically, when namespace labels change, everything will be the same, except at the end where it needs to create a CID and assign it, instead it will wait for matching CID based on labels to be present in the watcher's cache. This is part of the change to identity allocation for new pods.

cilium/CFP-27752-operator-manages-cilium-identities.md
dlapcevic: Applied change

PR #62, opened 2024-12-03
Neighbor Entry Refresh For Non-Cluster IPs

jrife:
## Cilium Feature Proposal

**Is your proposed feature related to a problem?**

Neighbor entries for non-cluster IPs can remain stale forever breaking connectivity in some cases (e.g. node port traffic). Cilium should provide *some* mechanism to ensure that neighbor entries on cluster nodes for non-cluster IPs remain up-to-date when using BPF host routing.

Context: https://github.com/cilium/cilium/issues/34503


joestringer:
cc @cilium/sig-datapath

pchaigno:
cc @borkmann

cilium/CFP-36372-neigh-entry-refresh-non-cluster.md
```diff
@@ -0,0 +1,136 @@
...
+* Requires no changes to the kernel.
+* Easy to implement
+
+#### Cons
+* Not very flexible and harder to manage. The feedback from our team internally
+  was that in their situation it is hard to know ahead of time which IPs will be
+  used.
+
+### Option 2: Always use `bpf_redirect_neigh()`
+Always use `bpf_redirect_neigh()` even if `bpf_fib_lookup()` returns
```
tommyp1ckles, line 94: Is there a particular set of conditions that make using the stale entry provided by fib lookup more likely?  Maybe this could be an opt-int option rather than making it the default (i.e. to balance performance vs correctness)?

jrife: > Is there a particular set of conditions that make using the stale entry provided by fib lookup more likely?

Assuming this behavior is opt-in, this would apply to any environments where a user has some host(s) talking to cluster nodes in the same L2 domain and where the IP address of those external hosts may migrate to new hosts.

I believe our specific scenario involved a virtual load balancer appliance talking to a node port service where both the appliance and cluster nodes are VMs. In some scenarios (environment upgrades) these VMs are replaced and IPs reassigned to their replacements breaking connectivity.

> Maybe this could be an opt-int option rather than making it the default (i.e. to balance performance vs correctness)?

This is certainly an option, although I worry about adding yet another flag or option for users to dig through to understand if it applies to their scenario. Hopefully we can just have performance *and* correctness. Perhaps some benchmarking is warranted here to determine what the effects would be of always using `bpf_redirect_neigh()` if we don't already have some numbers available. My hope would be that the performance difference is negligible.

joestringer:
From the community meeting: Seems like this may be a less controversial bugfix to propose - Discussing on a PR instead is a good step forward.

jrife:
Going to move this to a PR.

PR #63, opened 2025-01-16
Egress Connection Timeouts Design

rari459:
As mentioned in the discussion with Cilium Community on 1/8/2025, we propose an additional timeouts field to CiliumEgressGatewayPolicy to allow users to control egress connection timeouts values at the CEGP level.  This lays out approaches for designing this feature.

Issue: https://github.com/cilium/cilium/discussions/37454


cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,319 @@
...
+We propose enabling users to adjust connection timeout settings for specific egress gateways via the addition of a new field in Cilium Egress Gateway Policy (CEGP).
+
+
+# Motivation
+
+Currently, Cilium relies heavily on default operating system connection timeout settings and offers control over connection timeouts at cluster or node level. It is not optimal for all workloads to be bound to these node level timeouts, especially with respect to egress gateways where prolonged idle connections can contribute to port exhaustion on the NAT gateway.
+
+Modifying CiliumEgressPolicy to include an optional timeout field would allow us to ingest custom timeouts and give users additional control over egress connections. 
+
+
```
julianwiedmann, line 25: Some High-level comment
1. do you have feedback from potential users on whether they're comfortable fiddling with such low-level values? Making this a per-CEGP configuration feels like a reasonable granularity, but I'm unsure whether admins will know what timeouts to use. And how to tune them.
2. I haven't grasped whether you're proposing to use the custom timeouts only on the GW node, or also for the CT entry on the worker nodes?
3. As your motivation is to avoid port exhaustion on the gateway - what interaction do you see with the CT GC engine? Getting reliable results from configuring low CT lifetimes would require an appropriate GC timer, no?

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,319 @@
...
+## Overview
+
+**Steps to take are as follows:**
+
+
+
+*   Add an optional timeout field to CEGP.
+*   Modify CEGP ingestion logic to add new timeout fields to EGRESS_POLICY_MAP
```
julianwiedmann, line 106: I believe extending the map value will require a map migration. At that point, let's consider what additional values we should place into the map value - I've wanted to store the ifindex of the egress interface for a long time :).

julianwiedmann (commented):
Thank you, this sort of low-level datapath optimization is an interesting idea! I left some initial thought inline to better understand the proposal.

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,319 @@
...
+*   Use key of saddr and daddr and make unique entry for each src endpoint
+
+Include a boolean field `custom_timeouts_specified `to specify whether custom connection timeouts were specified in the CEGP. This will prevent additional map lookups in the default case.
+
+
+```
+struct egress_gw_policy_entry {
+	__u32 egress_ip;
+	__u32 gateway_ip;
+	__u8 custom_timeouts_specified;
```
tommyp1ckles, line 272: Migrating entry code could present some complexity with regard to upgrading Cilium. Possible alternative approach could be to enable the lookup code only if the feature is enabled.

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,319 @@
...
+   </td>
+   <td style="background-color: #efefef">Option 2. Timeout at Node Level
+   </td>
+   <td style="background-color: #efefef">Option 3. Timeout at Socket Level
+   </td>
+  </tr>
+  <tr>
+   <td style="background-color: null">Engineering Investment
+   </td>
+   <td style="background-color: #ffe599">Modify SNAT Datapath and conntrack creation to ingest and write custom timeouts to egress nat conntrack entries.
```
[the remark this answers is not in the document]
tommyp1ckles: To broaden a bit, I think we'd need to detail how the various mechanisms for enforcing timeouts would play out in practice - especially with regards to intended use cases. 

For example, are we looking for timeout functionality from/in-parity with kernel sysctl settings (ex. ipv4 timeout) such that the connection is immediately terminated following the expiry?

That might help with making a decision.

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,319 @@
...
+   </td>
+  </tr>
+  <tr>
+   <td style="background-color: null">Engineering Investment
+   </td>
+   <td style="background-color: #ffe599">Modify SNAT Datapath and conntrack creation to ingest and write custom timeouts to egress nat conntrack entries. 
+   </td>
+   <td style="background-color: #b6d7a8">Watch CEGP and modify CiliumNodeConfig for the respective gateway node
+   </td>
+   <td style="background-color: #ea9999"> New bpf program to hook into sockopt syscall event
```
tommyp1ckles, line 64: From what I know, this would be a novel approach in Cilium and would require building out new bpf infrastructure that we don't currently.  We may want to consider that for this approach.

cilium/egress-connection-timeouts.md
rari459: 1. I will try to collect some data on potential users via the Slack Channel and update.
2. Since Egress Policy Map is available at both nodes, I think it makes the most sense to update timeouts for ct entries for both src and gw nodes. 
3. Right now we do not make any changes to the GC engine. The report interval for GC is set to 5s by default and is configurable. I think this config can be left for the user and out of the scope for this feature, if the user sets low CT lifetimes they should configure the CT report interval to be lower.

cilium/egress-connection-timeouts.md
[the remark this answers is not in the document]
rari459: We want to keep the behavior consistent with current cilium implementation. Today ct timeouts are configurable at the cluster/node level and these timeouts are not passed up to the socket. This feature just looks to improve the granularity of the existing functionality.

CT timeouts are especially useful for the EGW node because they can free up ports from stale connections and prevent exhaustion. This is the primary motivator for the feature; so passing these timeouts up to socket level is not a goal of this feature.

cilium/egress-connection-timeouts.md
rari459: Makes sense, the preferred approach in which we modify the existing datapath should bypass this issue.

cilium/egress-connection-timeouts.md
rari459: Makes sense, we can introduce a new cilium config flag to enable this feature.

cilium/egress-connection-timeouts.md
rari459: Sounds good, where can I get a list of proposed additional values to add for this map?

julianwiedmann: I don't think such a list exists, it's something we can raise with @cilium/egress-gateway once work on an implementation has started.

cilium/egress-connection-timeouts.md
rari459: Ah yes, I confused GC Interval with Ct Report Interval.
CT GC interval is set dynamically and proportionally by taking into account the percent of entries deleted from the ct_table in each cycle. 

I believe this would be sufficient for the initial implementation of this feature as it keeps the footprint of this change small. If a large number of ct_entries are being deleted in each interval on the GW Node due to low CT timeouts, then GC interval should happen more frequently.

I'm proposing this simpler approach for now to minimize the initial impact.  However, I agree that a more comprehensive review of the GC interval mechanism could be beneficial down the road.  Perhaps we can explore that as a follow-up enhancement to this feature?

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,319 @@
...
+**SIG: SIG-DATAPATH, EGRESS-GATEWAY**
+
+**Begin Design Discussion:** 2024-01-16
+
+**Cilium Release:** X.XX
+
+**Authors:** Murari Bhimavarapu <murarib@google.com>
+
+**Contributors:** Carlos Abad <carlosab@google.com>
+
```
xmulligan, line 12:
```suggestion

**Status:** Implementable

```

xmulligan: see https://github.com/cilium/design-cfps?tab=readme-ov-file#status

nathanjsweet:
Ongoing discussion is taking place here: https://github.com/cilium/cilium/discussions/37454

tommyp1ckles:
@rari459 Just a few more comments, seems like we have a couple things we need to update on from the discussion over on https://github.com/cilium/cilium/discussions/37454.

Overall, from my perspective this proposal makes sense, with the ux changes regarding the EgressGatewayPolicy I think this provides a good way to expose this somewhat low level configuration that some users may be interested in.

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,317 @@
...
+
+```
+apiVersion: cilium.io/v2
+kind: CiliumEgressGatewayPolicy
+metadata:
+  name: egress-sample
+spec:
+  [other fields] ...
+  connection-timeouts: |
```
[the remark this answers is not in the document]
rari459: Since the expected users of this field are power users, I wanted to see if we could simplify the field to only accept custom timeouts. The advantage to keeping the profile and custom fields is that we could specify some preset connection timeout settings but this leads to the issue of actually having a hard-coded set of connection timeout configurations. 

To avoid this, I merged the custom field into the connection timeouts field but I am open to keeping the profile and custom fields if we feel that having preset fields is needed. In that case there still is the question of what those preset fields should be.

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,317 @@
...
+
+It is crucial to understand that utilizing this custom field signifies an acceptance of risk. Users who delve into these implementation details are expected to (a) fully acknowledge and bear the responsibility for any instability or unexpected behavior resulting from their custom configurations, and (b) proactively monitor upstream Cilium development for changes in these settings. Furthermore, these power users are encouraged to actively participate in the Cilium community, providing feedback on their experiences and helping to guide the evolution of these advanced options. While we will strive to release-note any modifications to the interpretation of this field, there are fewer guarantees regarding the stability and compatibility of these settings across different Cilium versions. Strong and consistent feedback from multiple users regarding the utility and effectiveness of specific custom settings may lead to their consideration for formal inclusion in future API revisions.
```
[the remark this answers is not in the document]
rari459: Sounds good, I'll move this out.

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,313 @@
...
+
+
+# Non-Goals
+
+
+
+*   Modify connection timeouts for non-egress traffic.
+*   Modify connection timeouts at a more granular level than CiliumEgressGatewayPolicy, i.e. per egress connection.
+*   Modify egress connection timeouts for IPv6 traffic.
+
```
joestringer, line 43:
```suggestion
* Commit to stable API for timeout configuration in CEGP
```

joestringer (commented):
Seems like this latest version takes into account the concerns I raised in https://github.com/cilium/cilium/discussions/37454 regarding API commitments. Minor suggestion to reinforce below on that.

Other than that I defer to @cilium/egress-gateway reviewers for feedback.

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,313 @@
+# CFP-TBD: Egress Connection Timeouts
+
+**SIG: SIG-DATAPATH, EGRESS-GATEWAY**
+
+**Begin Design Discussion:** 2024-01-16
+
+**Cilium Release:** X.XX
```
joestringer, line 7: Small administrative item now that the PR is approved: v1.18 development closes soon, so v1.19 would be the next ideal target.
```suggestion
**Cilium Release:** 1.19
```

cilium/egress-connection-timeouts.md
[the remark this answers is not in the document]
tommyp1ckles: > We want to keep the behavior consistent with current cilium implementation. 

Makes sense, just wanted to make that this was clear regarding the intended functionality of this.  Probably worth outlining this in documentation,

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,313 @@
...
+
+
+## Option 1 Impacts:
+
+
+### Performance of  SNAT Datapath 
+
+**Latency:**
+
+We will be performing one lookup in the EGRESS_POLICY_MAP for all egress nat cases.
```
tommyp1ckles, line 191: Just for the new ct egress case though right.

tommyp1ckles (approved):
Proposal makes sense to me, couple last thoughts:

* @pchaigno raises a good point on how we can optimize the codepath for checking if we have ct timeouts set.
* The migration will probably require a upgrade procedure, because we're just adding new fields I don't think we need to worry about downgrades.  In the past we've done this by creating a new map called something like ex: `cilium_egress_policy_v2` and realizing the new configuration to this map upon startup.  @julianwiedmann mentioned extending by adding ifindex etc but imo we can probably brainstorm that and then follow up with another migration down the line.

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,275 @@
...
+## Add Timeouts Field to Existing BPF EGRESS_POLICY_MAP
+
+
+```
+struct egress_gw_policy_entry {
+	__u32 egress_ip;
+	__u32 gateway_ip;
+	__u8 custom_timeouts_set:1,
+	     reserved:7;
+	struct connection_timeouts;
```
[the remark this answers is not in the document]
rari459: ACK

rari459:
Thanks for all the feedback! Addressed all comments and should be good for merge.

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,275 @@
+# CFP-TBD: Egress Connection Timeouts
```
joestringer, line 1:
```suggestion
# CFP-37454: Egress Connection Timeouts
```

cilium/egress-connection-timeouts.md
```diff
@@ -0,0 +1,275 @@
...
+
+**SIG: SIG-DATAPATH, EGRESS-GATEWAY**
+
+**Begin Design Discussion:** 2024-01-16
```
joestringer, line 5:
```suggestion
**Begin Design Discussion:** 2025-01-16
```

joestringer (approved):

cilium/egress-connection-timeouts.md
joestringer, line 1: Please rename the file to `cilium/CFP-37454-egress-connection-timeouts.md`.

PR #65, opened 2025-02-06
Add CFP for flow log aggregation

anubhabMajumdar:
Adding a proposal for `flowlog` aggregation in Hubble.


xmulligan:
cc: @cilium/sig-hubble

hubble/CFP-37472-flowlog-aggregation.md
```diff
@@ -0,0 +1,147 @@
...
+
+Reduces ingestion and storage cost of logs, while providing insight into cluster networking. Querying the data is also simplified because of already aggregated data.
+
+### Will this impact existing clusters?
+
+No, this is not a breaking change. By default, the new option won't aggregate any logs. Adding fields to `fieldAggregate` will still write the same `Flow` json structure in the same location - only the number of entries written will differ. Any tooling built to ingest these logs will keep working as expected.
+
+### Will this incur more resource usage?
+
+Depending on implementation, the aggregation mechanism would incur additional CPU and memory cost in the agent.
```
rolinh, line 128: I think that aggregation may actually result in decreased CPU resource usage. We know that the main cause of CPU usage at export time is JSON serialization given that this step uses reflection, which is rather costly. Thus, exporting less flow results in less events to serialize which potentially result in similar or lower CPU resource usage for flow export overall.

rolinh (approved):
I think the proposal makes a lot of sense and the design is sane. The approach of using `fieldAggregate` is both intuitive (for users who are used to `fieldMask` at least) and flexible.

The only caveat I can think of at the moment is potential pitfalls when combining `fieldMask` and `fieldAggregate` (e.g. exclude a field via `fieldMask` but attempt to use it in `fieldAggregate`). However, I think this is a problem that can be solved with specific checks.

hubble/CFP-37472-flowlog-aggregation.md
```diff
@@ -0,0 +1,147 @@
+# CFP-37472: Support aggregation to manage flow log volume on disk
+
+**SIG: SIG-NAME** SIG-HUBBLE-API/Observability
+
+**Begin Design Discussion:** 2025-02-06
+
+**Cilium Release:** N/A
+
+**Authors:** Anubhab Majumdar <anmajumdar@microsoft.com> , Neha Aggarwal <neaggarw@microsoft.com>
```
xmulligan, line 9:
```suggestion
**Authors:** Anubhab Majumdar <anmajumdar@microsoft.com> , Neha Aggarwal <neaggarw@microsoft.com>

**Status:** Implementable
```

xmulligan: see https://github.com/cilium/design-cfps?tab=readme-ov-file#status

hubble/CFP-37472-flowlog-aggregation.md
```diff
@@ -0,0 +1,147 @@
...
+
+This configuration can cut down log volume by significant amount. This helps store more logs using same disk space on host. Also, less I/O writes on disk.
+
+### Impact: Cost
+
+Reduces ingestion and storage cost of logs, while providing insight into cluster networking. Querying the data is also simplified because of already aggregated data.
+
+### Will this impact existing clusters?
+
+No, this is not a breaking change. By default, the new option won't aggregate any logs. Adding fields to `fieldAggregate` will still write the same `Flow` json structure in the same location - only the number of entries written will differ. Any tooling built to ingest these logs will keep working as expected.
```
kaworu, line 124: From the Motivation section:
```
| source_pod | source_port | destination_pod | destination_port | protocol | ingress_flow_count | egress_flow_count |
|------------|-------------|-----------------|------------------|----------|--------------------|-------------------|
| client-1   | 12345       | server          | 80               | TCP      | 4                 | 6                 |
| client-2   | 23456       | server          | 80               | TCP      | 4                 | 6                 |
```
How are the `ingress_flow_count` and `egress_flow_count` computed with the current JSON flow structure?

anubhabMajumdar: I was thinking of extending Flow using the [extensions](https://github.com/cilium/cilium/blob/81c42ff218655c50c11e5cb36f7b488a42d04f87/api/v1/flow/flow.proto#L129) field.

hubble/CFP-37472-flowlog-aggregation.md
anubhabMajumdar: Thanks for pointing this out, I will edit this line to include this information.

anubhabMajumdar:
> I think the proposal makes a lot of sense and the design is sane. The approach of using `fieldAggregate` is both intuitive (for users who are used to `fieldMask` at least) and flexible.
> 
> The only caveat I can think of at the moment is potential pitfalls when combining `fieldMask` and `fieldAggregate` (e.g. exclude a field via `fieldMask` but attempt to use it in `fieldAggregate`). However, I think this is a problem that can be solved with specific checks.

@rolinh  I was thinking the `fieldMask` and `fieldAggregate` can work independently. We use `fieldAggregate` first to group the flows we see from dataplane, then use `fieldMask` when writing to file. I can test this out during implementation to make sure there's no error.

anubhabMajumdar:
@rolinh @xmulligan @kaworu Thanks for the reviews! 
Addressed your comments and updated the proposal. Let me know if this can be merged.

neaggarwMS:
> > I think the proposal makes a lot of sense and the design is sane. The approach of using `fieldAggregate` is both intuitive (for users who are used to `fieldMask` at least) and flexible.
> > The only caveat I can think of at the moment is potential pitfalls when combining `fieldMask` and `fieldAggregate` (e.g. exclude a field via `fieldMask` but attempt to use it in `fieldAggregate`). However, I think this is a problem that can be solved with specific checks.
> 
> @rolinh I was thinking the `fieldMask` and `fieldAggregate` can work independently. We use `fieldAggregate` first to group the flows we see from dataplane, then use `fieldMask` when writing to file. I can test this out during implementation to make sure there's no error.

Do we need checks? if fieldMask is empty, shouldnt we add all fields, no?

rolinh:
> Let me know if this can be merged.

Let's way a couple more days to give time to other Hubble maintainers to have a look at the proposal.

> Do we need checks?

I think we just need to ensure the relation from `fieldMask` to `fieldAggregate` is intuitive.

> if fieldMask is empty, shouldnt we add all fields, no?

Yes

hubble/CFP-37472-flowlog-aggregation.md
```diff
@@ -0,0 +1,150 @@
...
+| client-2   | 23456       | server          | 80               | TCP      | FIN     |
+| server     | 80          | client-2        | 23456            | TCP      | FIN-ACK |
+| client-2   | 23456       | server          | 80               | TCP      | ACK     |
+
+However, we can gleam the same level of information if the logs were aggregated as follows:
+
+| source_pod | source_port | destination_pod | destination_port | protocol | ingress_flow_count | egress_flow_count |
+|------------|-------------|-----------------|------------------|----------|--------------------|-------------------|
+| client-1   | 12345       | server          | 80               | TCP      | 4                 | 6                 |
+| client-2   | 23456       | server          | 80               | TCP      | 4                 | 6                 |
```
glibsm, line 52: Does this assume that we take the traffic direction into account, or just ignore the reply flows?

hubble/CFP-37472-flowlog-aggregation.md
```diff
@@ -0,0 +1,150 @@
...
+          - source_pod: ["default/"]
+            event_type:
+            - type: 1
+          - destination_pod: ["frontend/webserver-975996d4c-7hhgt"]
+          excludeFilters: []
+          end: "2023-10-09T23:59:59-07:00"
+        - name: "test003"
+          filePath: "/var/run/cilium/hubble/test003.log"
+          fieldMask: ["source", "destination","verdict"]
+          fieldAggregate: ["source.pod_name", "destination.pod_name", "l4.TCP.destination_port"]
```
glibsm, line 106: To expand a little bit on the https://github.com/cilium/design-cfps/pull/65/files#r1946717206 question:

If we use SQL group_by as something that has similar functionality to aggregation, what happens to all the other fields _not_ aggregated? 

When a group_by is added to the sql query, all the other fields not included in the group_by need an aggregation function, so integers can be summed, values concatenated, etc.

Using your provided table as an example, lets say we don't include the destination port into the aggregation and just use `source_pod, source_port`, how are destination_pod, destination_port, protocol and flags are going to be handled?

```
| source_pod | source_port | destination_pod | destination_port | protocol | flags   |
|------------|-------------|-----------------|------------------|----------|---------|
| client-1   | 12345       | server          | 80               | TCP      | SYN     |
| server     | 80          | client-1        | 12345            | TCP      | SYN-ACK |
| client-1   | 12345       | server          | 80               | TCP      | ACK     |
| client-1   | 12345       | server          | 80               | TCP      | PSH     |
| server     | 80          | client-1        | 12345            | TCP      | ACK     |
| server     | 80          | client-1        | 12345            | TCP      | PSH     |
| client-1   | 12345       | server          | 80               | TCP      | ACK     |
| client-1   | 12345       | server          | 80               | TCP      | FIN     |
| server     | 80          | client-1        | 12345            | TCP      | FIN-ACK |
| client-1   | 12345       | server          | 80               | TCP      | ACK     |
| client-2   | 23456       | server          | 80               | TCP      | SYN     |
| server     | 80          | client-2        | 23456            | TCP      | SYN-ACK |
| client-2   | 23456       | server          | 80               | TCP      | ACK     |
| client-2   | 23456       | server          | 80               | TCP      | PSH     |
| server     | 80          | client-2        | 23456            | TCP      | ACK     |
| server     | 80          | client-2        | 23456            | TCP      | PSH     |
| client-2   | 23456       | server          | 80               | TCP      | ACK     |
| client-2   | 23456       | server          | 80               | TCP      | FIN     |
| server     | 80          | client-2        | 23456            | TCP      | FIN-ACK |
| client-2   | 23456       | server          | 80               | TCP      | ACK     |
```

hubble/CFP-37472-flowlog-aggregation.md
anubhabMajumdar: I am thinking we should take into account `is_reply` and provide the ingress and egress count of "flows" observed.  
`egress_flow_count` : `client-1` ---> `server`
`ingress_flow_count` : `client-1` <--- `server`

hubble/CFP-37472-flowlog-aggregation.md
anubhabMajumdar: I want to ignore all the other fields and aggregate on the fields user provides. Ideally, the `fields.mask` should be a subset to `fields.aggregate`. Maybe we can document this? If `fields.aggregate` is non-empty, then it would override the `field.mask` field.

glibsm: Thank you for clarifying. 

So effectively `fieldAggregate: ["one", "two", "three"]` will become `select one, two, three, count(*) from ... group by one, two, three` (with additional accountability for traffic direction)

glibsm (approved):

anubhabMajumdar:
> Please squash your commits if you want to preserve commit messages, or we will squash on merge and the commit message will be more automated.

Squashed commits into one. Ready to merge! Thanks!

PR #66, opened 2025-02-07
Add status to the template

xmulligan:
Adding the four statuses Implementable, Released, Dormant, Declined to the CfP template


joestringer (approved):
I think it could be useful to also have a draft or provisional status but until we resolve https://github.com/cilium/design-cfps/issues/48 we can just go with this.

PR #68, opened 2025-02-24
cfp/37315 BGP - Auto Peer Discovery

naveenachyuta:
This PR adds a CFP for Auto Peer Discovery for Cilium BGP https://github.com/cilium/cilium/issues/37315

### DIFF
--- a/cilium/CFP-37315-bgp-auto-discovery.md
+++ b/cilium/CFP-37315-bgp-auto-discovery.md
@@ -1,0 +1,188 @@
+# CFP-37315
+
+**SIG: SIG-BGP**
+
+**Begin Design Discussion:** 2024-02-24
+
+**Cilium Release:** X.XX
+
+**Authors:** Naveen Achyuta <naveen.achyuta22@gmail.com>
+
+**Status:** Implementable
+
+**Issue:** https://github.com/cilium/cilium/issues/37315
+
+## Summary
+
+Cilium’s current BGP implementation requires specifying the peer IP address in the BGP cluster configuration. In large-scale environments with thousands of Kubernetes nodes, managing distinct BGP configuration files (one per rack) becomes impractical.
+
+This proposal offers a simpler alternative: rely on the default gateway of ToR switches to automatically create BGP sessions when no peer IP is provided in the BGP cluster configuration.
+
+## Motivation
+
+Large networks with thousands of Kubernetes nodes cannot feasibly manage the peer IP addresses of all ToR switches within numerous BGP cluster configurations. Allowing Cilium to automatically discover and use the  default gateway as the peer simplifies BGP session creation, reducing operational overhead and configuration complexity.
+
+## Goals
+
+### Auto-Discovery of Peer IPs
+Allow Cilium to establish BGP sessions with the default gateway on the node.
+### Single Config File for All Nodes
+Enable large networks to use a generic BGP configuration file without needing to specify unique peer IP addresses on a per-rack basis.
+
+## Non-Goals
+
+- Allow discovering the peer from arbitrary prefix ([ref](https://github.com/cilium/design-cfps/pull/68#issuecomment-2722443710))
+
+## Proposal
+
+### Overview
+
+At a high level, this proposal modifies how Cilium handles BGP peer IP configuration. If a user omits the peer IP for a BGP process, the Cilium agent attempts to retrieve the node’s default gateway from statedb (which stores local routing information). Cilium can then use this default gateway as the peer IP to establish the BGP session. If the default gateway is not found, the agent skips the neighbor and logs an error for it.
+
+### BGP Cluster Config
+
+If the user enables auto-discovery for a peer, cilium agent should use the discovery mode to get the peer ip. If default gateway mode is being used,
+it should rely on statedb to get default gateway. If not found, it should skip the neighbor and log an error.
+
+Every peer will have to specify the address family under autoDiscovery config block so that we fetch the default gateway for that address family only.
+
+If there are multiple default gateways, select the default gateway with the minimum metric. If it couldn't break the tie in this way, the reconciliation fails.
+
+config proposal:
+```
+bgpInstances:
+  - name: "65001"
+    localASN: 65001
+    peers:
+    - name: "peer1 - 65000"
+      peerASN: 65000
+      peerConfigRef:
+        name: "cilium-peer"
+      autoDiscovery:
+        mode: "default-gateway"
+        defaultGateway:
+          addressFamily: ipv4 // or ipv6
+          interface: eth0 (optional)
+      
+```
+In multi-homed environments (i.e., multiple default gateways leading to different ASNs), the above proposal does not address which gateway belongs to which peer ASN.
+
+example multi-homed config:
+```
+bgpInstances:
+  - name: "65001"
+    localASN: 65001
+    peers:
+    - name: "65000"
+      peerASN: 65000
+      peerConfigRef:
+        name: "cilium-peer"
+    - name: "65011"
+      peerASN: 65011
+      peerConfigRef:
+        name: "cilium-peer"
+```
+If multiple default gateways exist, we cannot reliably match them to different peer ASNs. One workaround is to use same local ASN on both the ToRs for bgp sessions with the k8s node, and bgp cluster config will have one peer configured.
+
+Another solution can be to add an optional "interface" configuration parameter which can be used to determine the default gateway to use for peering with ToR.
+An important consideration for the interface parameter is ensuring consistent interface naming across Kubernetes deployments. Since servers may use PCI slots, different NIC models, SKUs, and vendors may have varying interface naming conventions.
+
+proposed multi-homed config:
+
+```
+bgpInstances:
+  - name: "65001"
+    localASN: 65001
+    peers:
+    - name: "65000"
+      peerASN: 65000
+      peerConfigRef:
+        name: "cilium-peer"
+      autoDiscovery:
+        mode: "default-gateway"
+        defaultGateway:
+          addressFamily: ipv4 // or ipv6
+          interface: eth0 (optional)
+```
+
+proposed config in the device (arista):
+```
+bgp listen range 10.2.0.0/24 peer-group SERVERS remote-as 65001
+router bgp 65100
+   router-id 10.1.1.1
+   timers bgp 1 3
+   neighbor SERVERS local-as 65000 no-prepend replace-as
+```
+
+### Routes from Statedb
+
+cilium agent should rely on statedb to get the default gateway to create the bgp session. we can inject statedb into neighbor reconcillation struct and when we loop through neighbors, we can populate the peer ip by looking for default gateway of 0.0.0.0/0 or ::0/0 route in routes table of statedb.
+
+In multi-homing environments, when the BGP session with the primary default gateway (lowest metric) fails, statedb detects the routing table change and we can trigger the establishment of a new BGP session with the alternate default gateway.
+
+## Impacts / Key Questions
+
+### Impact 1: Multi-Homing Scenario
+
+In multi-homed environments, we will be breaking the design of one peer per bgp session because we will create two bgp sessions with one peer defined in the config. One of the workarounds could be to define two peers with same config and a unique name.
+
+### Key Question
+
+Is there a different way to accomplish this?
+
+
+### Impact 2: Multiple BGP Instances
+
+The proposal does not take into account multiple bgp instances so auto-discovery will work in that case.
+
+### Key Question
+
+What are use-cases of having multiple bgp instances? (bgp instances is a list in config)
+
+### Impact 3: config knob for auto-discovery
+
+config knob to enable auto-discovery will be crucial because that enables the users to understand the dependencies before making the changes. We can have documentation for auto-discover and provide the pre-requisites for enabling it.
+Without the knob, the user may use auto-discovery without understanding the requirements for it.
+
+### Key Question
+
+are we okay to add a knob for auto-discovery or should we enable auto-discovery by default if peerAddress is not specified?
+
+
+### Option 1:
+
+If the user wants to create bgp sessions without specifying peerAddress, the user will remove peerAddress field, enable auto-discovery in the config and add ipv4 and/or ipv6 peers explicitly with "afi" field
+
+#### Pros
+
+Adds flexibility to create ipv4 and/or ipv6 sessions.
+auto-discovery field will help users to understand the requirements provided in the documentation before using it. (User can obviously enable it without knowing the requirements, but, we can error out with clear error messages)
+
+#### Cons
+
+For multi-homed environments, it deviates from the current "one bgp session per configured peer" design if we decide to use only one bgp peer for both sessions.
+
+### Option 2:
+
+do nothing
+
+#### Pros
+
+
+
+#### Cons
+
+User will have to manage bgp cluster files for all the k8s nodes and updates to these files can be cumbersome
+
+## Future Milestones
+
+### Deferred Milestone 1
+
+add optional interface configuration knob to auto discovery to identify the default
+gateway to peer with
+https://github.com/cilium/design-cfps/pull/68/files#r1993328948
+
+### Deferred Milestone 2
+
+add BGP unnumbered for auto-discovery using ipv6 link local addresses and interface
+https://github.com/cilium/design-cfps/pull/68#pullrequestreview-2691943125


cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,127 @@
...
+#### Cons
+
+Deviates from the current "one bgp session per configured peer" design.
+Still does not address multi-homed or multiple-peer scenarios.
+
+## Future Milestones
+
+_List things that this CFP will enable but that are out of scope for now. This can help understand the greater impact of a proposal without requiring to extend the scope of a CFP unnecessarily._
+
+### Deferred Milestone 1
```
naveenachyuta, line 123: should i put auto-discovery of multiple peers as a deferred milestone?

xmulligan:
cc: @cilium/sig-bgp

cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,165 @@
...
+proposed multi-homed config:
+bgpInstances:
+  - name: "65001"
+    localASN: 65001
+    peers:
+    - name: "65000"
+      peerASN: 65000
+      afi: ipv4
+      peerConfigRef:
+        name: "cilium-peer"
```
YutaroHayakawa, line 88: I think as long as the selection is consistent, it is ok to select only one of them and use it. For example, we can use metric to compare them. In my understanding, Linux doesn't allow us to create two default gateway with the same metric. So, it should be possible prefer the route with the lowest metric for example.

Since we are anyways subscribing to the netlink, we can also do failover when one of them disappears.

cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,165 @@
...
+What are use-cases of having multiple bgp instances? (bgp instances is a list in config)
+
+### Impact 3: config knob for auto-discovery
+
+config knob to enable auto-discovery will be crucial because that enables the users to understand the dependencies before making the changes. We can have documentation for auto-discover and provide the pre-requisites for enabling it.
+Without the knob, the user may use auto-discovery without understanding the requirements for it.
+
+### Key Question
+
+are we okay to add a knob for auto-discovery or should we enable auto-discovery by default if peerAddress is not specified?
```
YutaroHayakawa, line 129: I'd prefer an explicit knob. Also, that implicit knob only allows single type of the discovery mechanism. We may want to add different kind of discovery mechanism in the future.

YutaroHayakawa (approved):
Hi, sorry for the late reply. I made an initial review.

YutaroHayakawa (changes requested):
Sorry I meant to request change.

cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,165 @@
...
+    - name: "peer1 - 65000"
+      peerASN: 65000
+      afi: ipv4
+      peerConfigRef:
+        name: "cilium-peer"
+    - name: "peer2 - 65000"
+      peerASN: 65000
+      afi: ipv6
+      peerConfigRef:
+        name: "cilium-peer"
```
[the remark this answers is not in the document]
naveenachyuta: fixed.

cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,165 @@
...
+
+If the user skips the peer ip for a bgp process and enables auto discovery, cilium agent should not return back an error when its creating new bgp sessions. It should first rely on statedb to get default gateway. If not found, it should skip the neighbor and log an error.
+Every peer will have to specify the address family so that we fetch the default gateway for that address family
+
+config proposal:
+
+bgpInstances:
+  - name: "65001"
+    localASN: 65001
+    autoDiscovery: true
```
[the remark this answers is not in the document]
naveenachyuta: ya i agree. it will provide flexibility per peer. but i'm curious about the use-case. What is the use-case to use different discovery mechanisms for each peer OR auto-discovery for one peer and static ip for another?

cilium/CFP-37315-bgp-session-with-default-gateway.md
naveenachyuta: perfect! 100% agreed!

cilium/CFP-37315-bgp-session-with-default-gateway.md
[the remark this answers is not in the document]
YutaroHayakawa: For example, using default gateway to peer with ToR and use static ip to peer with route reflector. Whether or not this is realistic, I'd prefer the option giving the maximum leverage. The instance-level knob enforces users to use default gateway for all peers.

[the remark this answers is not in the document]
naveenachyuta: okay i was wondering if there is an existing use-case. 
but i agree with you. it provides flexibility for sure.

cilium/CFP-37315-bgp-session-with-default-gateway.md
YutaroHayakawa: > are you saying that in multi-homing, we will create a bgp session with default gateway of lowest metric?

Yes, right. Then, when the lowest one disappears, we'll be notified through statedb (statedb watches netlink, so we don't need to deal with it directly). Then now we can select the one left on the routing table (the higher one). As a result, we failed over.

naveenachyuta: okay got it. If you can send sample code or parts of cilium code that i can update to make this work, that would be helpful.

cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,170 @@
...
+        name: "cilium-peer"
+    - name: "peer2 - 65000"
+      peerASN: 65000
+      afi: ipv6
+      peerConfigRef:
+        name: "cilium-peer"
+```
+In multi-homed environments (i.e., multiple default gateways leading to different ASNs), this proposal does not address which gateway belongs to which peer ASN.
+
+example multi-homed config:
```
rastislavs, line 63: Maybe we could add an optional config knob `interface` that could help in multi-homing scenario? E.g.:

```yaml
peers:
- name: peer0
  peerASN: 65000
  peerConfigRef:
    name: peer-config
  autoDiscovery:
    mode: default-gateway
    addressFamily: ipv4
    interface: eth0
- name: peer1
  peerASN: 65001
  peerConfigRef:
    name: peer-config
  autoDiscovery:
    mode: default-gateway
    addressFamily: ipv4
    interface: eth1
```

In this case, for `peer0` we would peer with the gateway based on the default route via `eth0`, for `peer1` we would peer with the gateway based on the default route via `eth1`.

cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,170 @@
...
+In multi-homed environments, we will be breaking the design of one peer per bgp session because we will create two bgp sessions with one peer defined in the config. One of the workarounds could be to define two peers with same config and a unique name.
+
+### Key Question
+
+Is there a different way to accomplish this?
+
+
+### Impact 2: Multiple BGP Instances
+
+The proposal does not take into account multiple bgp instances so auto-discovery will work in that case.
```
rastislavs, line 121: I think what is proposed here may actually work in multi-instance setup as well, instance1 and instance2 could even peer with the same GW IP, with a different peer port in the peer config. Or instance1 can use auto-peering and instance2 static peering.

cilium/CFP-37315-bgp-session-with-default-gateway.md
harsimran-pabla: Practically speaking, this might not be a big problem. Common scenario is server peering with two ToRs. So, it will be via different physical interfaces. I am okay to introduce optional `interface` field under `autoDiscovery` stanza.

YutaroHayakawa: The idea of the interface field sounds nice to me. With the BGP-based first hop redundancy scenario, this could help. Namely, the host is connected to two ToRs and have two default routes advertised from them with different metrics. In that case, on the FIB, there are two default routes with different interfaces.

YutaroHayakawa:
BTW, what do you think about generalizing this approach a bit and allow arbitrary prefix instead of default route? It might be useful for some multi-homing scenario.

We can do like:

```
# Default Gateway
autoDiscovery:
  mode: prefix
  prefix: 0.0.0.0/0

# Private subnet
autoDiscovery:
  mode: prefix
  prefix: 10.0.0.0/8
```

Cilium search for the route entry with exact matching (not prefix matching). Address family is not required anymore since it's obvious from the specified prefix.

harsimran-pabla:
> BTW, what do you think about generalizing this approach a bit and allow arbitrary prefix instead of default route?

Are you suggesting something like [dynamic peering](https://github.com/osrg/gobgp/blob/master/docs/sources/dynamic-neighbor.md)?

YutaroHayakawa:
Well, close, but while the dynamic peering is a feature to accept the BGP session initiation from the certain IP range, this is still about finding the peer.

cilium/CFP-37315-bgp-session-with-default-gateway.md
dswaffordcw: I see where `interface` would be helpful, but I'd like to +1 the point about it being an optional setting.  Within the same cluster, and the same server SKU, our primary network interface shows up in Linux with a variety of names such as `enp193s0np0`.  The PCI slot used appears to be part of the name.

cilium/CFP-37315-bgp-session-with-default-gateway.md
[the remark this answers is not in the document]
naveenachyuta: updated the config and explanation

cilium/CFP-37315-bgp-session-with-default-gateway.md
naveenachyuta: yes i was about to mention that. Our initial proposal had interface as a way to solve multi-homing but i removed it because servers with pci slots can have different interface naming convention across different per nic model, per sku or per vendor. If a kubernetes cluster has different types of nic types or models, this will not work.

cilium/CFP-37315-bgp-session-with-default-gateway.md
naveenachyuta: got it. just curious can i know the use-case of having multiple instances of bgp?

cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,170 @@
...
+
+**Begin Design Discussion:** 2024-02-24
+
+**Cilium Release:** X.XX
+
+**Authors:** Naveen Achyuta <naveen.achyuta22@gmail.com>
+
+## Summary
+
+Cilium’s current BGP implementation requires specifying the peer IP address in the BGP cluster configuration. In large-scale environments with thousands of Kubernetes nodes, managing distinct BGP configuration files (one per node) becomes impractical.
```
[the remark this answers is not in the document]
naveenachyuta: i will fix that. it should be rack

cilium/CFP-37315-bgp-session-with-default-gateway.md
naveenachyuta: i'm okay to add an optional interface config so that the user can decide on their own if they want to use it or not. 
but it still wont solve the problem for users that have different server vendors, thats why, we proposed our approach of using local-asn in the ToRs

YutaroHayakawa: Ah, indeed. "Predictable" interface name feature is always tricky. Users can still deal with that by adjusting [udev device renaming scheme](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html/networking_guide/sec-understanding_the_device_renaming_procedure#sec-Understanding_the_Device_Renaming_Procedure), but it also has another problem.

Yep, I think we can make it optional and even don't have to implement it from the beginning. It is still possible to add it without breaking the compatibility later. We can just take a note.

YutaroHayakawa:
Yep, sounds good to me.

cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,167 @@
...
+## Goals
+
+### Auto-Discovery of Peer IPs
+Allow Cilium to establish BGP sessions by discovering the default gateway on the node, when no peer IP is explicitly provided in the BGP cluster configuration.
+### Single Config File for All Nodes
+Enable large networks to use a generic BGP configuration file without needing to specify unique peer IP addresses on a per-node basis.
+
+## Non-Goals
+
+* _List aspects which are specifically out of context for this CFP._
```
YutaroHayakawa, line 30:
```suggestion
- Allow discovering the peer from arbitrary prefix ([ref](https://github.com/cilium/design-cfps/pull/68#issuecomment-2722443710))
```

cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,167 @@
...
+This proposal offers a simpler alternative: rely on the default gateway of ToR switches to automatically create BGP sessions when no peer IP is provided in the BGP cluster configuration.
+
+## Motivation
+
+Large networks with thousands of Kubernetes nodes cannot feasibly manage the peer IP addresses of all ToR switches within numerous BGP cluster configurations. Allowing Cilium to automatically discover and use the  default gateway as the peer simplifies BGP session creation, reducing operational overhead and configuration complexity.
+
+## Goals
+
+### Auto-Discovery of Peer IPs
+Allow Cilium to establish BGP sessions by discovering the default gateway on the node, when no peer IP is explicitly provided in the BGP cluster configuration.
```
YutaroHayakawa, line 24:
```suggestion
Allow Cilium to establish BGP sessions with the nexthop IP address of the default gateway on the node.
```

cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,167 @@
...
+### Single Config File for All Nodes
+Enable large networks to use a generic BGP configuration file without needing to specify unique peer IP addresses on a per-node basis.
```
YutaroHayakawa, line 26:
```suggestion
Enable large networks to use a generic BGP configuration file without needing to specify unique peer IP addresses on a per-rack basis.
```

cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Overview
+
+At a high level, this proposal modifies how Cilium handles BGP peer IP configuration. If a user omits the peer IP for a BGP process, the Cilium agent attempts to retrieve the node’s default gateway from statedb (which stores local routing information). Cilium can then use this default gateway as the peer IP to establish the BGP session. If the default gateway is not found, the agent skips the neighbor and logs an error for it.
+
+### BGP Cluster Config
+
+If the user enables auto-discovery for a peer, cilium agent should use the discovery mode to get the peer ip. If default gateway mode is being used,
+tt should rely on statedb to get default gateway. If not found, it should skip the neighbor and log an error.
+Every peer will have to specify the address family under autoDiscovery config block so that we fetch the default gateway for that address family only.
```
YutaroHayakawa, line 42:
```suggestion
If the user enables auto-discovery for a peer, cilium agent should use the discovery mode to get the peer ip. If default gateway mode is being used,
it should rely on statedb to get default gateway. If not found, it should skip the neighbor and log an error.

Every peer will have to specify the address family under autoDiscovery config block so that we fetch the default gateway for that address family only.

If there are multiple default gateways, select the default gateway with the minimum metric. If it couldn't break the tie in this way, the reconciliation fails.
```

cilium/CFP-37315-bgp-session-with-default-gateway.md
```diff
@@ -0,0 +1,167 @@
...
+
+
+
+#### Cons
+
+User will have to manage bgp cluster files for all the k8s nodes and updates to these files can be cumbersome
+
+## Future Milestones
+
+_List things that this CFP will enable but that are out of scope for now. This can help understand the greater impact of a proposal without requiring to extend the scope of a CFP unnecessarily._
```
YutaroHayakawa, line 161: We can potentially mention about the `interface` field here if we decided to not implement it in the first iteration.

https://github.com/cilium/design-cfps/pull/68/files#r1993328948

naveenachyuta: ya. I think we should add it to future milestones. It will be better to keep it simple and iterate over it when its needed. also, adding interface field should be easy to add in the future.

naveenachyuta:
@YutaroHayakawa @harsimran-pabla @dswaffordcw @rastislavs I made all the changes needed to reflect the final solution we discussed except interface optional config. we need to have consensus on whether it should be a future milestone or it should be part of the initial iteration.

Please read through the doc one more time and let me know if i need to change anything. Thanks!

harsimran-pabla:
>  interface optional config. we need to have consensus on whether it should be a future milestone or it should be part of the initial iteration.

I think you should keep the interface option in the CFP. It does not need to be part of initial implementation. We can iteratively add it as needed. It also makes for a good first issue for someone to contribute.

dswaffordcw:
@naveenachyuta Could you update the title for this PR to `Auto Peer Discovery for Cilium BGP`?  That'll make it more clear to others browsing the overall PR history.

naveenachyuta:
@dswaffordcw thats a good point. updated the PR description and the file name as well.

@harsimran-pabla okay i added it to future milestone and its also part of the multi-homing proposed config with some explanation

cilium/CFP-37315-bgp-auto-discovery.md
```diff
@@ -0,0 +1,176 @@
...
+  - name: "65001"
+    localASN: 65001
+    peers:
+    - name: "peer1 - 65000"
+      peerASN: 65000
+      peerConfigRef:
+        name: "cilium-peer"
+      autoDiscovery:
+        mode: default-gateway
+        addressFamily: ipv4 // or ipv6
```
YutaroHayakawa, line 59: Sorry, one last suggestion. Probably, we should isolate `addressFamily` field to the isolated stanza. Otherwise, it will be confusing when we introduce another auto discovery mode and the address family doesn't make sense for it.

```suggestion
      defaultGateway:
        addressFamily: ipv4 // or ipv6
```

naveenachyuta: hmm thats a good point. agreed!
if we decide to use bgp unnumbered some day, then that config will not be useful.

xmulligan:
Can you also add the status when it is ready? https://github.com/cilium/design-cfps?tab=readme-ov-file#status

YutaroHayakawa (approved):
Now looks good to me. Thanks!

naveenachyuta:
@xmulligan i updated the status to Implementable because this PR is approved. When do i merge the PR?
Do i need to create a PR in cilium with the code changes, get that approved and then i can merge this?

dswaffordcw:
> @xmulligan i updated the status to Implementable because this PR is approved. When do i merge the PR? Do i need to create a PR in cilium with the code changes, get that approved and then i can merge this?

@naveenachyuta I would put `Implementable`.  This should get merged before your source code changes merge.  You likely don't have merge permissions, so typically one of the other maintainers will hit that button for you (same w/ the source code changes to Cilium).

@xmulligan re: the `Released` status, is the desire to publish a follow-up PR to the original CFP to capture that?

harsimran-pabla (changes requested):
Hi 👋 Before we move forward with this, I have few more concerns and alternative solution to this problem - aka BGP unnumbered.

There are many limitations with current default route peering approach.
- It simplifies peering in very specific context - single peering, single default route. This overall concerns me that we are implementing feature for specific setup. This can give impression that Cilium recommends (maybe that is too strong word) such topologies.
- Route-based peer discovery combined with multiple interfaces can be harmful as our config starts relying on a transient network state (where default route is advertised from).

This problem is solved in practice using BGP unnumbered, which works with IPv6 link local addresses.
- Configuration wise, you specify which interface peering should be setup.
- IPv6 RAs should be sent/received on that interface.
- Peering will be setup using IPv6 LL addresses on that interface, which can exchange v4/v6 AFI/SAFIs.

This is more documented and practiced solution, see ref 
https://github.com/osrg/gobgp/blob/master/docs/sources/unnumbered-bgp.md

xmulligan:
@naveenachyuta 

@dswaffordcw answer is correct. Implementable is the right status. A maintainer (like @YutaroHayakawa) will merge is once there is agreement on the proposal. Looks like there are still some questions from @harsimran-pabla before we can merge it.

@dswaffordcw correct, ideally once the code is merged, the CFP will be updated to reflect how it was coded and the status updated to Released

harsimran-pabla:
If the primary concern is interface name in BGP unnumbered, it could be a regex match instead of exact match in the CRD - with some restriction that it can match only single interface on the node.

harsimran-pabla:
Overall, I'd prefer that we implement single solution for a given problem, instead of having multiple ways to solve it. It reduces code and maintenance burden. BGP unnumbered aligns with general industry practice on how to solve this problem.

harsimran-pabla:
Had another discussion offline with @YutaroHayakawa  and @rastislavs, overall code duplication should not be large, so I am okay with going ahead with both implementations. Couple of requests
- Please document guidelines under which default gateway peering should be used. 
- BGP unnumbered should be added as a future milestone in this CFP.

naveenachyuta:
awesome! I will work on creating a PR for it.

@xmulligan i think we have consensus for this proposal. you can merge this.

xmulligan:
I'll let @YutaroHayakawa merge it since he is the committer with the most relevant experience in this area

joestringer:
The DCO check also failed, probably because the git commit author email is different from the sign off. You should be able to fix this up either with `git commit -s --amend` or by resetting the author to match your git configuration. Might be a good time to double-check which email you have configured in git to check whether that's the one you intend to use for contributions. https://github.com/cilium/design-cfps/pull/68/checks?check_run_id=39009998322 .

cilium/CFP-37315-bgp-auto-discovery.md
```diff
@@ -0,0 +1,186 @@
...
+```
+bgpInstances:
+  - name: "65001"
+    localASN: 65001
+    peers:
+    - name: "peer1 - 65000"
+      peerASN: 65000
+      peerConfigRef:
+        name: "cilium-peer"
+      autoDiscovery:
```
YutaroHayakawa, line 61: Please revive the `mode` field. I know that this `defaultGateway` stanza is already enough to express that we're using default-gateway auto discovery, but introducing `mode` field makes an implementation a bit simpler when there's one or more discovery mode.

Without mode

```golang
enabledAutodiscovery := []string{}
if d.DefaultGateway != nil {
    append(enabledAutodiscovery, "default-gateway")
}
if d.BGPUnnumbered != nil {
    append(enabledAutodiscovery, "bgp-unnumbered")
}
if len(enabledAutodiscovery) > 1 {
    return fmt.Errorf("only one auto discovery method should be enabled: %v", enabledAutodiscovery)
}
switch enabledAutodiscovery[0] {
case "default-gateway":
    configureDefaultGateway(d.DefaultGateway)
case "bgp-unnumbered":
    configureBGPUnnumbered(d.BGPUnnumbered)
}
```

With mode

```golang
switch d.Mode {
case "default-gateway":
    configureDefaultGateway(d.DefaultGateway)
case "bgp-unnumbered":
    configureBGPUnnumbered(d.BGPUnnumbered)
}
```

YutaroHayakawa (changes requested):

cilium/CFP-37315-bgp-auto-discovery.md
naveenachyuta: makes sense. fixed

naveenachyuta:
> The DCO check also failed, probably because the git commit author email is different from the sign off. You should be able to fix this up either with `git commit -s --amend` or by resetting the author to match your git configuration. Might be a good time to double-check which email you have configured in git to check whether that's the one you intend to use for contributions. https://github.com/cilium/design-cfps/pull/68/checks?check_run_id=39009998322 .

fixed. DCO check passed.

YutaroHayakawa (approved):

YutaroHayakawa:
Ok, let me merge this now. Please go ahead and implement it. Thanks for your patience!

PR #73, opened 2025-06-17
CFP-32810: Add hybrid routing mode in Cilium

anubhabMajumdar:
Adding a CFP to add hybrid routing mode in Cilium.
Ref: https://github.com/cilium/cilium/issues/32810

### DIFF
--- a/cilium/CFP-32810-hybrid-routing-mode.md
+++ b/cilium/CFP-32810-hybrid-routing-mode.md
@@ -1,0 +1,163 @@
+# CFP-32810: Add hybrid routing mode in Cilium
+
+**SIG: SIG-NAME** SIG-Datapath
+
+**Begin Design Discussion:** 2025-06-16
+
+**Cilium Release:** 1.19
+
+**Authors:** Anubhab Majumdar <anmajumdar@microsoft.com>, Vamsi Kalapala <vakr@microsoft.com>, Krunal Jain <krunaljain@microsoft.com>
+
+**Status:**
+
+Implementable
+
+## Summary
+
+Currently, Cilium supports two modes routing - [encapsulation](https://docs.cilium.io/en/stable/network/concepts/routing/#encapsulation) and [native-routing](https://docs.cilium.io/en/stable/network/concepts/routing/#native-routing). When workloads are on cross subnets with no direct connection, Cilium needs to run in encapsulation mode. This is not efficient, as every packet needs to be encapsulated, even when they are on the same subnet, thus paying the encapsulation overhead and reducing network throughput. This CFP proposes adding a third option to allow Cilium dataplane to make the decision at runtime on whether to route the packet natively or encapsulate.
+
+## Motivation
+
+Consider any of the following scenarios:
+
+1. Two Kubernetes cluster connected through cluster mesh. The nodes are backed by different subnets in two different VNETs, which are connected through VNET peering. The pods are assigned IP addresses from overlay network.
+2. In a large Kubernetes cluster, pod addresses are assigned from two different subnets with no direct connection.
+
+In either case, Cilium needs to run in encapsulation routing mode. However, when routing packets in above clusters, in many situations, we can get by native routing, and not pay the encapsulation overhead that comes with encapsulation. The reason for this inflexibility comes from the fact that Cilium datapath is not aware of the underlying network topology, and hence cannot make a decision between the two routing modes during runtime.
+
+With higher adoption in cluster mesh and large scale clusters being common to support AI workloads, it would be beneficial to have a more performant dataplane that can optimize the throughput, while providing support for both encapsulation and native-routing.
+
+## Goals
+
+* Retain the behavior of current routing modes as-is
+* Allow users to set a new routing mode called `hybrid`
+* Allow users to pass in network topologies expressed as CIDR ranges through configmap
+* Support for both IPV4 and IPV6
+* Allow dynamic configuration of CIDRs (add/remove) without agent restart (update w/o agent restart)
+* Support for multiple subnets that are not not all mutually routable
+
+## Proposal
+
+### Overview
+
+Add a new option for routing-mode config as follows `routing-mode: hybrid`. Also, add a new field in config to pass information about the CIDRs and connectivity. The CIDR information will persist in pinned eBPF maps so that routing decision can be made in runtime by eBPF programs. The concept is similar to [IP Masquerading](https://docs.cilium.io/en/stable/network/concepts/masquerading/#ebpf-based) and the eBPF table it uses.
+
+### Examples
+
+```
+subnet-topology: 10.0.0.1/24
+```
+
+If source and destination IP belongs to above CIDR, route natively. Else, preserve current behavior (encapsulate if being routed to known CIDRs/pass to Linux stack).
+
+```
+subnet-topology: 10.0.0.1/24,10.10.0.1/24
+```
+
+The above two CIDR ranges have direct connectivity. If source and destination IP belongs to either CIDR, route natively. Else, preserve current behavior (encapsulate if being routed to know CIDRs/pass to Linux stack).
+```
+subnet-topology: 10.0.0.1/24,10.10.0.1/24;10.20.0.1/24
+```
+
+* `10.0.0.1/24` and `10.10.0.1/24` are connected and traffic can be natively routed
+* If source and destination IP belongs to `10.20.0.1/24`, route natively
+* Else, preserve current behavior (encapsulate if being routed to know CIDRs/pass to Linux stack).
+
+### Control plane implementation
+
+Similar to dynamic flowlogs or Hubble dynamic exporter, allow users to configure the CIDRs on the fly through a configmap. The configmap will be mounted on each agent. On create/update, update `LocalNodeConfig` and trigger the eBPF map reconcile.
+
+Here's how the different topologies would be represented in-memory cache for the following config: `10.0.0.1/24,10.10.0.1/24;10.20.0.1/24;2001:0db8:85a3::/64`
+
+| CIDR | Subnet ID |
+|------|-----------|
+| 10.0.0.1/24 | 1  |
+| 10.10.0.1/24 | 1 |
+| 10.20.0.1/24 | 2 |
+| 2001:0db8:85a3::/64 | 3 |
+
+#### Notes:
+
+- Subnet ID allocation starts from 1
+- If config changes, replace the entire cache with new Subnet ID calculation instead of trying to reconcile the diff. Predicating it on the assumption that subnet segmentation would be limited (like not 100s of CIDR blocks for a cluster).
+
+### Datapath implementation
+
+This section provides an overview of the datapath implementation. Consider the current implementation - as long as nodes have IP connectivity, Cilium can route traffic, both intra-cluster and inter-cluster (in case of cluster mesh). 
+
+#### eBPF Map
+
+The above in-memory cache should be represented in a pinned eBPF map. Should look something like (representative, not actual implementation)
+
+```
+struct lpm_trie_key {
+    __u32 prefixlen;     // 32 for IPv4, 128 for IPv6
+    __u8  addr[16];      // Supports both IPv4 (first 4 bytes) and IPv6 (full 16 bytes)
+};
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LPM_TRIE);
+    __uint(max_entries, 1024);
+    __type(key, struct lpm_trie_key);
+    __type(value, __u32); // or a custom struct
+    __uint(map_flags, BPF_F_NO_PREALLOC);
+} cidr_subnet_map SEC(".maps");
+```
+
+#### Function
+
+Datapath should implement a new function to extract Subnet ID for given IP address.
+
+```
+/*
+lookup_subnet_id returns subnet ID for given IPV4 or IPV6 address
+- return 0 if no matching CIDR found
+- return ID (>=1) otherwise
+*/
+lookup_subnet_id(addr)
+```
+
+#### Routing decision
+
+Retain datapath implementation for routing as is, except use the above function to check for subnet IDs before making tunneling decision ([example](https://github.com/cilium/cilium/blob/a84cacd7d639ba16d90f707c91c2cbfd4010bf6c/bpf/bpf_lxc.c#L718)).
+
+`skip_tunnel_for_same_subnet_id = lookup_subnet_id(src) == lookup_subnet_id(dest)`
+
+#### Implication
+
+If `hybrid` mode is enabled without configuring any `subnet-topology`, it will behave the same as `tunnel` mode. Pod IPs can be from the overlay or from a subnet, traffic will be encapsulated either way. So datapath implementation would be exactly as it is today for the `tunnel` routing mode.
+
+If `subnet-topology` is configured, the CIDR ranges can be configured as part of node configs, similar to [IPv4PodSubnets](https://github.com/cilium/cilium/blob/f78aca7b5dff52e6d07723f6577dd0d3f913fea6/pkg/datapath/types/node.go#L177).
+
+Similar to how masquerade uses `NativeRoutingCIDRIPv4` to check if [SNAT is needed](https://github.com/cilium/cilium/blob/f78aca7b5dff52e6d07723f6577dd0d3f913fea6/bpf/lib/nat.h#L713), this implementation would add a similar function that can check source and destination address against multiple CIDRs. However, in this case, since we will be storing multiple CIDRs, we will use an eBPF map of type `BPF_MAP_TYPE_LPM_TRIE` to efficiently search the map. 
+
+Next, while making routing decision today, in both `lxc` and `host` eBPF programs, a check is made to see if tunneling should be done. This implementation would add an extra check to tunneling decision. The check would use above function to figure out if native routing can be done. 
+- If yes, we skip tunneling and the packet is instead sent to Linux stack for handling, where the underlying networking would route the packet to destination. 
+- If no, tunneling happens exactly as it works today 
+
+Above behavior simply optimizes current encapsulation behavior, without introducing major change to how encapsulation routing works/behaves. 
+
+## Impacts
+
+### Impact: Network throughput
+
+Given that today Cilium is encapsulating packets that could be natively routed, removing those scenarios would yield higher throughput. This is especially true in clusters without jumbo frames.
+
+### Impact: Datapath latency
+
+In a highly fragmented network topography, lookup of IP to make routing decision may not be negligible.
+
+## Key Questions
+
+### Connectivity disruption
+
+Things to consider during implementation to minimize traffic disruption:
+- how to ensure atomic/revision-based update to the subnet eBPF map, and minimize traffic disruption because of the update
+- since subnets can be updated on the fly, it is important to retain connectivity for existing connections between connected subnets that remain connected both before and after the subnet configuration change
+- when restarting or upgrading Cilium, datapath should likewise retain the connectivity
+
+Resolving these questions is deferred to the implementation phase when we expect to learn more about the solution space.
+
+## Future Milestones
+
+This initiative can be extended through a more formal CRD in future.


cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,83 @@
+# CFP-32810: Add hybrid routing mode in Cilium
+
+**SIG: SIG-NAME** SIG-Clustermesh
+
+**Begin Design Discussion:** 2025-06-16
+
+**Cilium Release:** N/A
```
xmulligan, line 7: is there a release you want to target?

anubhabMajumdar: Preferably 1.18, but I guess it would depend on feedback on this CFP and what the final scope of the work would be.

anubhabMajumdar:
cc: @cilium/sig-datapath 

I think a lot of the changes in this CFP would involve `datapath` changes, so asking their feedback as well.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Non-Goals
+
+* Allow dynamic update to CIDR ranges (update without restart)
+
+## Proposal
+
+### Overview
+
+Add a new option for routing-mode config as follows `routing-mode: hybrid`. Also, add a new field in config to pass information about the CIDRs and connectivity. The CIDR information will persist in pinned eBPF maps so that routing decision can be made in runtime by eBPF programs.
```
joestringer, line 40: Currently we have `routing-mode: native` + native routing CIDR that determines the directly reachable subnets (which implies that Cilium then masquerades to other destinations). I wonder if there's some conceptual overlap.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Examples
+
+```
+subnet-topology: 10.0.0.1/24
+```
+
+If source and destination IP belongs to above CIDR, route natively. Else, encapsulate the packet.
```
joestringer, line 48: The "else encapsulate" seems to hint at a default route for tunneling traffic, is that intended? Typically today in tunnel mode, only traffic to known CIDRs (typically PodCIDRs) is encapsulated, then the remainder of the traffic is passed up to the stack to follow whatever default routing that Linux has configured.

joestringer: I guess there's two notable points here: One is that "else encapsulate" doesn't currently make sense and would require some new notion of how to determine the remote tunnel endpoint for fallthrough traffic. The second point is that it is useful in many cases to be able to defer to the native Linux routing stack to determine forwarding behavior for traffic that Cilium isn't explicitly configured to handle.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,83 @@
+# CFP-32810: Add hybrid routing mode in Cilium
+
+**SIG: SIG-NAME** SIG-Clustermesh
```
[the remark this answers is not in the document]
anubhabMajumdar: Agreed. I have cc'd datapath SIG for feedback.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,83 @@
...
+
+**Begin Design Discussion:** 2025-06-16
+
+**Cilium Release:** N/A
+
+**Authors:** Anubhab Majumdar <anmajumdar@microsoft.com>, Vamsi Kalapala <vakr@microsoft.com>, Krunal Jain <krunaljain@microsoft.com>
+
+## Summary
+
+Currently, Cilium supports two modes routing - [encapsulation](https://docs.cilium.io/en/stable/network/concepts/routing/#encapsulation) and [native-routing](https://docs.cilium.io/en/stable/network/concepts/routing/#native-routing). When workloads are on cross subnets with no direct connection, Cilium needs to run in encapsulation mode. This is not efficient, as every packet needs to be encapsulated, even when they are on same subnet, thus paying the MTU overhead and reducing network throughput. This CFP proposes adding a third option to allow Cilium dataplane to make the decision at runtime on whether to route the packet natively or encapsulate.
```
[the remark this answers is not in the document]
anubhabMajumdar: My mistake; I meant encapsulation overhead. Fixed.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,83 @@
...
+## Proposal
+
+### Overview
+
+Add a new option for routing-mode config as follows `routing-mode: hybrid`. Also, add a new field in config to pass information about the CIDRs and connectivity. The CIDR information will persist in pinned eBPF maps so that routing decision can be made in runtime by eBPF programs.
+
+### Examples
+
+```
+subnet-topology: 10.0.0.1/24
```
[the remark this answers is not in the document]
anubhabMajumdar: Agreed. Updating to reflect this.

cilium/CFP-32810-hybrid-routing-mode.md
anubhabMajumdar: Right, the idea is kind-of similar. In hybrid mode, we would allow native routing to ranges configured via option. Otherwise encapsulate/pass to Linux stack.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,83 @@
...
+
+Add a new option for routing-mode config as follows `routing-mode: hybrid`. Also, add a new field in config to pass information about the CIDRs and connectivity. The CIDR information will persist in pinned eBPF maps so that routing decision can be made in runtime by eBPF programs. The concept is similar to [IP Masquerading](https://docs.cilium.io/en/stable/network/concepts/masquerading/#ebpf-based) and the eBPF table it uses.
+
+### Examples
+
+```
+subnet-topology: 10.0.0.1/24
+```
+
+If source and destination IP belongs to above CIDR, route natively. Else, preserve current behavior (encapsulate if being routed to know CIDRs/pass to Linux stack).
```
giorio94, line 49: Same for all occurrences.

```suggestion
If source and destination IP belongs to above CIDR, route natively. Else, preserve current behavior (encapsulate if being routed to known CIDRs/pass to Linux stack).
```

giorio94 (commented):
Thanks! Looks reasonable to me from a high-level perspective; I'm not part of sig-datapath though, and I will defer to them the actual detailed review. A couple of typo fixes inline.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Motivation
+
+Consider any of the following scenarios:
+
+1. Two Kubernetes cluster connected through cluster mesh. The nodes are backed by different subnets in two different VNETs, which are connected through VNET peering. The pods are assigned IP address from overlay network.
+2. In a large Kubernetes cluster, pod addresses are assigned from two different subnets with no direct connection.
+
+In either case, Cilium needs to run in encapsulation routing mode. However, when routing packets in above clusters, in many situations, we can get by native routing, and not pay the MTU overhead that comes with encapsulation. The reason for this inflexibility comes from the fact that Cilium datapath is not aware of the underlying network topology, and hence cannot make a decision between the two routing mode during runtime.
```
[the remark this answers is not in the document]
anubhabMajumdar: Fixed

pchaigno (changes requested):
Thanks! Looks like a useful feature that makes a lot of sense. Per below comments, I think we'll need to discuss a bit the datapath implementation to understand the potential tradeoffs here.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+The above two CIDR ranges have direct connectivity. If source and destination IP belongs to either CIDR, route natively. Else, preserve current behavior (encapsulate if being routed to know CIDRs/pass to Linux stack).
+```
+subnet-topology: 10.0.0.1/24,10.10.0.1/24;10.20.0.1/24
+```
+
+* `10.0.0.1/24` and `10.10.0.1/24` are connected and traffic can be natively routed
+* If source and destination IP belongs to `10.20.0.1/24`, route natively
+* Else, preserve current behavior (encapsulate if being routed to know CIDRs/pass to Linux stack).
```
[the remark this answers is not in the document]
anubhabMajumdar: Added a section to outline a datapath implementation.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Future Milestones
+
+### Deferred Milestone 1
+
+Allow dynamic configuration of CIDRs (add/remove) without agent restart.
+
+### Deferred Milestone 2
+
+Configure the subnets though a custom resource.
```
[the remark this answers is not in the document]
anubhabMajumdar: Moved the dynamic update as a goal, I think it's best if user can add new CIDR block without restart.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,95 @@
...
+
+Similar to dynamic flowlogs or Hubble dynamic exporter, allow users to configure the CIDRs on the fly through a configmap. The configmap will be mounted on each agent. On create/update, update `LocalNodeConfig` and trigger dataplane reconcile.
+
+### Datapath implementation
+
+This section provides an overview of the datapath implementation. Consider the current implementation - as long as nodes have IP connectivity, Cilium can route traffic, both intra-cluster and inter-cluster (in case of cluster mesh). 
+
+#### Case 1
+
+If `hybrid` mode is enabled without configuring any `subnet-topology`, routing mode will behave exactly same as encapsulation mode. Pod IPs can be from overlay or from a subnet, it will be encapsulated either way. So datapath implementation would be exactly as it is today for encapsulation routing mode.
```
pchaigno, line 71: Just on wording for clarity:
```suggestion
If `hybrid` mode is enabled without configuring any `subnet-topology`, it will behave the same as `tunnel` mode. Pod IPs can be from the overlay or from a subnet, traffic will be encapsulated either way. So datapath implementation would be exactly as it is today for the `tunnel` routing mode.
```

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,95 @@
...
+
+If `subnet-topology` is configured, the CIDR ranges can be configured as part of node configs, similar to [IPv4PodSubnets](https://github.com/cilium/cilium/blob/f78aca7b5dff52e6d07723f6577dd0d3f913fea6/pkg/datapath/types/node.go#L177).
+
+Similar to how masquerade uses `NativeRoutingCIDRIPv4` to check if [SNAT is needed](https://github.com/cilium/cilium/blob/f78aca7b5dff52e6d07723f6577dd0d3f913fea6/bpf/lib/nat.h#L713), this implementation would add a similar function that can check source and destination address against multiple CIDRs. However, in this case, since we will be storing multiple CIDRs, we will use an eBPF map of type `BPF_MAP_TYPE_LPM_TRIE` to efficiently search the map. 
+
+Next, while making routing decision today, in both `lxc` and `host` eBPF programs, a check is made to see if tunneling should be done. This implementation would add an extra check to tunneling decision. The check would use above function to figure out if native routing can be done. 
+- If yes, we skip tunneling and the packet is instead sent to Linux stack for handling, where the underlying networking would route the packet to destination. 
+- If no, tunneling happens exactly as it works today
```
pchaigno, line 81: Could you describe how this check would be implemented exactly? In particular to handle the case `subnet-topology: 10.0.0.1/24,10.10.0.1/24;10.20.0.1/24` described above. I don't think it's as simple as skipping tunneling if both the src and dst IP addresses are found in the LPM map.

pchaigno (commented):
I think we still need more details on the datapath implementation. The current description doesn't seem to address the more advanced `subnet-topology` behaviors you listed.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -61,6 +58,29 @@ subnet-topology: 10.0.0.1/24,10.10.0.1/24;10.20.0.1/24
 * If source and destination IP belongs to `10.20.0.1/24`, route natively
 * Else, preserve current behavior (encapsulate if being routed to know CIDRs/pass to Linux stack).
 
+### Control plane implementation
+
+Similar to dynamic flowlogs or Hubble dynamic exporter, allow users to configure the CIDRs on the fly through a configmap. The configmap will be mounted on each agent. On create/update, update `LocalNodeConfig` and trigger dataplane reconcile.
```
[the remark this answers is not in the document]
pchaigno: Having read the datapath implementation section, I don't understand why a full datapath reconcile would be needed. Can't we simply update the LPM map?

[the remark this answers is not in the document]
anubhabMajumdar: I should have been explicit about this. I meant reconcile the config to the eBPF map. This shouldn't need full datapath restart that disrupts networking. I will rewrite this section to clarify my point.

pchaigno (approved):
A small concern on the current strategy to update the map, but LGTM otherwise.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,145 @@
...
+|------|-----------|
+| 10.0.0.1/24 | 1  |
+| 10.10.0.1/24 | 1 |
+| 10.20.0.1/24 | 2 |
+| 2001:0db8:85a3::/64 | 3 |
+
+#### Notes:
+
+- Subnet ID allocation starts from 1
+- If config changes, replace the entire cache with new Subnet ID calculation instead of trying to reconcile the diff. Predicating it on the assumption that subnet segmentation would be limited (like not 100s of CIDR blocks for a cluster).
```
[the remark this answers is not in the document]
anubhabMajumdar: Yeah, I thought about this as well. My intention is to make the map updates atomic (maybe spin locks), but I don't want to set it in stone in the CFP itself. I think PR would be a good place to discuss the specific implementation (lock or revision number)? Based on test/performance implication (if any).

cilium/CFP-32810-hybrid-routing-mode.md
joestringer: Similar or same?

I understand the idea that if we add another flag with its own parameters, that makes it easiest to avoid breaking some other existing functionality. At the same time, if the native routing CIDR determines the directly reachable subnet (and hence excludes traffic from SNAT/masquerade) and then we force the user to also separately configure the `subnet-topology`, then it seems like we're just asking the user to provide the same information twice in two different flags just because of some internal implementation details rather than because of the logical concept that's being configured.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,145 @@
...
+Similar to dynamic flowlogs or Hubble dynamic exporter, allow users to configure the CIDRs on the fly through a configmap. The configmap will be mounted on each agent. On create/update, update `LocalNodeConfig` and trigger the eBPF map reconcile.
+
+Here's how the different topologies would be represented in-memory cache for the following config: `10.0.0.1/24,10.10.0.1/24;10.20.0.1/24;2001:0db8:85a3::/64`
+
+| CIDR | Subnet ID |
+|------|-----------|
+| 10.0.0.1/24 | 1  |
+| 10.10.0.1/24 | 1 |
+| 10.20.0.1/24 | 2 |
+| 2001:0db8:85a3::/64 | 3 |
```
joestringer, line 72: Is there an assumption here that not all subnets are mutually routable? That is, a Cilium-managed node may be connected to multiple subnets, but other hosts may not be connected to the same subnets?

joestringer: ^ I would suggest putting that in the goals of the CFP to state the assumption up front.

cilium/CFP-32810-hybrid-routing-mode.md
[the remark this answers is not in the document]
joestringer: I think this makes sense. Probably worth calling this concern out somewhere in the CFP just to make sure we can come back and review it. I'm fine with not settling on a solution in the CFP before you've had a chance to dig into the options and figure out what works best.

There's two impacts I can think of that relate to this with the subnet ID allocation:
1. If Cilium is running and the subnets are updated, Cilium should retain connectivity for existing connections between connected subnets that remain connected both before and after the subnet configuration change
2. When restarting or upgrading Cilium, Cilium should likewise retain the connectivity.

Worth noting as well that this interface design specifically avoids delegating the subnet ID allocation problem to the user.

cilium/CFP-32810-hybrid-routing-mode.md
joestringer, line 1: I think that at some point we may want to consider a more formal CRD for this and other related configurations, but... well, I've tried to draft such a CRD in the past and there is a *lot* of possible interactions to consider so I don't really want to open that can of the worms too early. Therefore I think that specifying the subnets as config flags is probably the best option we have today.

cilium/CFP-32810-hybrid-routing-mode.md
joestringer: I think this will be answered in other threads below from my latest review, so I'll resolve this thread.

joestringer (commented):
All my comments are relatively minor, but I pulled out what I think are probably a couple of undeclared assumptions that the design is based on. I bring them up because if they are assumptions, it's good to include those assumptions in the text. If they're not assumptions, maybe there are other assumptions or maybe we could explore some simplifications to the design. I think the latter is probably not that likely.

Overall looks good, I just had a couple of points that might make sense to also highlight in the CFP:
1. Any assumptions as mentioned above; see threads below.
2. Note the "unresolved question" of how to handle subnet updates
3. Drop the "key question" or rephrase it to highlight what the question is and any options we have to resolve that question.

anubhabMajumdar:
@joestringer  @pchaigno @xmulligan 
Let me know if there's any additional concern around this plan. If not, can we check this in? Thanks again for all the reviews :-)

cilium/CFP-32810-hybrid-routing-mode.md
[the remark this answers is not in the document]
joestringer: Does it make sense to add a "key question" for this like "How can we minimize traffic impact when changing subnets or upgrading/downgrading Cilium?"

Even if we just state that this question requires investigation before we can agree on an approach, that would be OK. We don't have to resolve this key question during the CFP stage, I just want to make sure it's clearly noted so we don't forget about it :)

joestringer:
Looks fine to me, just one more minor question about the traffic disruption thread [here](https://github.com/cilium/design-cfps/pull/73#discussion_r2223297038).

cilium/CFP-32810-hybrid-routing-mode.md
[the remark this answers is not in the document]
anubhabMajumdar: Great idea. I have added our discussions under `Key Questions` that should be evaluated during implementation.

anubhabMajumdar:
> Looks fine to me, just one more minor question about the traffic disruption thread [here](https://github.com/cilium/design-cfps/pull/73#discussion_r2223297038).

Added `key questions` section. Thanks!

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,157 @@
+# CFP-32810: Add hybrid routing mode in Cilium
+
+**SIG: SIG-NAME** SIG-Datapath
+
+**Begin Design Discussion:** 2025-06-16
+
+**Cilium Release:** 1.19
+
+**Authors:** Anubhab Majumdar <anmajumdar@microsoft.com>, Vamsi Kalapala <vakr@microsoft.com>, Krunal Jain <krunaljain@microsoft.com>
```
xmulligan, line 9:
```suggestion
**Authors:** Anubhab Majumdar <anmajumdar@microsoft.com>, Vamsi Kalapala <vakr@microsoft.com>, Krunal Jain <krunaljain@microsoft.com>

**Status:**
```

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -8,6 +8,8 @@
 
 **Authors:** Anubhab Majumdar <anmajumdar@microsoft.com>, Vamsi Kalapala <vakr@microsoft.com>, Krunal Jain <krunaljain@microsoft.com>
 
+**Status:**
```
xmulligan, line 11: Please add a status here

https://github.com/cilium/design-cfps?tab=readme-ov-file#status

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Key Questions
+
+### Connectivity disruption
+
+Things to consider during implementation to minimize traffic disruption:
+- how to ensure atomic/revision-based update to the subnet eBPF map, and minimize traffic disruption because of the update
+- since subnets can be updated on the fly, it is important to retain connectivity for existing connections between connected subnets that remain connected both before and after the subnet configuration change
+- when restarting or upgrading Cilium, datapath should likewise retain the connectivity
+
```
joestringer, line 158: Assuming reviewers are good with this:

```suggestion

Resolving this question is deferred to the implementation phase when we expect to learn more about the solution space.

```

cilium/CFP-32810-hybrid-routing-mode.md
anubhabMajumdar: Adding `Implementable` here.

cilium/CFP-32810-hybrid-routing-mode.md
anubhabMajumdar: Added to the doc.

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,163 @@
...
+If source and destination IP belongs to above CIDR, route natively. Else, preserve current behavior (encapsulate if being routed to known CIDRs/pass to Linux stack).
+
+```
+subnet-topology: 10.0.0.1/24,10.10.0.1/24
+```
+
+The above two CIDR ranges have direct connectivity. If source and destination IP belongs to either CIDR, route natively. Else, preserve current behavior (encapsulate if being routed to know CIDRs/pass to Linux stack).
+```
+subnet-topology: 10.0.0.1/24,10.10.0.1/24;10.20.0.1/24
```
tamilmani1989, line 59: what's the usecase for this case `10.0.0.1/24,10.10.0.1/24;10.20.0.1/24` ? I couldn't understand this

tamilmani1989: does this represent in a cluster, pods belong to`10.0.0.1/24` or `10.10.0.1/24` can't natively route to pods belong to  `10.20.0.1/24` but pods within the range can be natively routed. So this is for hybrid routing within a cluster?

anubhabMajumdar: Yes. Comma separated subnets are natively routable, and `;` represents subnets which needs tunneling. This notation will support hybrid routing in a cluster.

tamilmani1989: good to clarify this in CFP

cilium/CFP-32810-hybrid-routing-mode.md
```diff
@@ -0,0 +1,163 @@
...
+
+* `10.0.0.1/24` and `10.10.0.1/24` are connected and traffic can be natively routed
+* If source and destination IP belongs to `10.20.0.1/24`, route natively
+* Else, preserve current behavior (encapsulate if being routed to know CIDRs/pass to Linux stack).
+
+### Control plane implementation
+
+Similar to dynamic flowlogs or Hubble dynamic exporter, allow users to configure the CIDRs on the fly through a configmap. The configmap will be mounted on each agent. On create/update, update `LocalNodeConfig` and trigger the eBPF map reconcile.
+
+Here's how the different topologies would be represented in-memory cache for the following config: `10.0.0.1/24,10.10.0.1/24;10.20.0.1/24;2001:0db8:85a3::/64`
```
tamilmani1989, line 70: why configmap and not CRD? Isn't this easy to represent in CRD than configmap?

anubhabMajumdar: We discussed this here - https://github.com/cilium/design-cfps/pull/73#discussion_r2199013817 .
Possibility of adding a CRD is a future milestone.

joestringer (approved):

joestringer:
It looks like there's been plenty of opportunity for discussion, and no new threads are coming up. I'll merge this in. Thanks!

anubhabMajumdar:
Thanks everyone for reviews and suggestions! Cheers!

pchaigno:
Looking forward to the PR(s)!

smagnani96:
Hey folks, while reviewing Part 4 https://github.com/cilium/cilium/pull/45001, concerns regarding UI/UX with config flags vs CRD have been expressed https://github.com/cilium/cilium/pull/45001#issuecomment-4516494464, and I'd love inputs from original CFP reviewers, as I see the topic was already touched and postponed.

I don't have prior expertise in Cilium with CRDs or migrating config flags into apposite CRDs (I can imagine the pain), but given this feature is new not yet in-use (hidden agent flags), it can make sense to go with the CRD option.

In case, regarding the proposed CRD, Here are an enriched example of what I understood, hope it's aligned and makes sense @networkop. As cons, I see headaches in case of v1 <-> v2 upgrade/downgrades. As pros, the CRD option is more expressive and less hacky. Adding CIDRs overlaps or catch-all rules might not be simple/intuitive with just config flags. Destinations with node selectors can be optional/unimplemented in v1 and left for future multi-pool IPAM support, with also apposite exclusions.

**NB:** adding per-group `mode: overlay` (which means "even if src and dst are in the same group, still tunnel") requires a BPF datapath change. For instance, right now two subnet with same ID automatically skip tunnel ([ex](https://github.com/cilium/cilium/blob/27afa2a15f8fd422bfe04daf1050f4fd625ab84b/bpf/bpf_host.c#L384)).

```yaml
 apiVersion: cilium.io/v2alpha1
  kind: CiliumSubnetTopology
  metadata:
    name: zone-a
  spec:
    # To which nodes the topology applies to.
    nodeSelector:
      matchLabels:
        topology.kubernetes.io/zone: us-east-1a

    # Each entry describes a set of remote destinations and how to reach them.
    destinations:
      # 1: Remote nodes in the same zone; derive CIDRs from IPAM.
      # (future: nodeSelector triggers dynamic CIDR resolution from CiliumNode)
      - name: "same-zone-nodes"
        nodeSelector:
          matchLabels:
            topology.kubernetes.io/zone: us-east-1a
        mode: native
        # Optional per-CIDR overrides within this group.
        cidrOverrides:
          - cidr: "10.0.99.0/24"
            mode: overlay  # force tunnel for a specific prefix even though group is native

      # 2: Explicit CIDRs for a cross-zone subnet
      - name: "cross-zone-x"
        mode: native
        cidrs:
          - "10.10.0.0/24"
          - "10.10.1.0/24"

      # 3: Force all IPv6 through overlay
      - name: "ipv6-overlay"
        mode: overlay
        cidrs:
          - "::/0"

      # 4: Explicit tunnel
      - name: "overlay-zone"
        mode: overlay
        cidrs:
          - "172.16.0.0/12"
```

While reasoning on this and reading the code, I had the following unrelated question: are we explicitly stating in the doc what is the behavior with respect to other features? For example, Transparent Encryption should not be impacted, being able to encrypt both native and tunneled traffic regardless. Similarly, Egress Gateway should not be affected, and clustermesh too should always ignore this and tunnel I think.

PR #78, opened 2025-10-28
CFP-42453: Oracle Cloud Infrastructure (OCI) Cloud Provider Design

trungng92:
[Cilium Issue Link](https://github.com/cilium/cilium/issues/42453)

As discussed during the Cilium Weekly Community meeting, this is a CFP that starts a discussion on various possible integrations with Cilium and will ultimately help determine which solution works best for OCI.

### DIFF
--- a/cilium/CFP-42453-oci-cloud-provider-design.md
+++ b/cilium/CFP-42453-oci-cloud-provider-design.md
@@ -144,30 +144,38 @@ _List crucial impacts and key questions. They likely require discussion and are 
 
 ### Impact: Integration with OCI
 
-### Impact: Increased Maintenanc
-
-Depending on the chosen solution, there may be additional maintenance and coordination required. 
-
-For in-tree solutions, the controller will need to be updated.
-
-For out-of-tree solutions, there may be additional integration testing.
+The Kubernetes IPAM solution requires the least effort to integrate with Cilium while also providing a Kubernetes Agnostic solution. All other solutions require additional maintenance or coordination to implement.
 
 ### Key Question: Cilium Testing
 
-How does Cilium perform CNI testing? What solution will require the least amount of maintenance/coordination?
+> How does Cilium perform CNI testing? What solution will require the least amount of maintenance/coordination?
+
+Cloud providers are responsible for their own solutions.
 
 ### Key Question: Cloud Provider Support Model
 
-What does the support model look like for Cloud Providers for integrated pieces, like the Cilium Operator?
+> What does the support model look like for Cloud Providers for integrated pieces, like the Cilium Operator?
+
+The team suggests not implementing in-tree solutions into the Cilium Operator because of the required coordination.
 
 ### Key Question: Multi-VNIC Solution
 
-Is there any negative impact from implementing a Multi-VNIC solution?
+> Is there any negative impact from implementing a Multi-VNIC solution?
+
+Cilium does not have strong support for Multi-VNIC solutions right now.
 
 ## Future Milestones
 
+## Single-VNIC Solution
+
+As a first pass, OCI will implement the Kubernetes IPAM solution.
+
+Ultimately, the simplicity of the solution along with the CNI agnostic behavior made this the most attractive short term solution.
+
 ### Multi-VNIC Solution
 
-Long term, OKE will provide a multi-NIC solution, where users can route pods out of VNICs.
+Long term, OKE will reevalute multi-NIC solutions, where users can route pods out of VNICs, when Dynamic Resource Allocation (DRA) provides better support.
 
-TODO: Provide more details on this
+[DRANet](https://github.com/google/dranet) is a possible future solution (it would need to incorporate [consumable capacity](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#consumable-capacity)).
+
+[Node Resource Interface](https://github.com/containerd/nri) is another possible alternative.
--- a/cilium/CFP-42453-oci-cloud-provider-design.md
+++ b/cilium/CFP-42453-oci-cloud-provider-design.md
@@ -1,0 +1,173 @@
+# CFP-42453: OCI Cloud Provider Design
+
+**SIG: SIG-COMMUNITY**
+
+**Begin Design Discussion:** 2025-10-15
+
+**Cilium Release:** X.XX
+
+**Authors:** Trung Nguyen <trung.tn.nguyen@oracle.com>
+
+**Status:** Draft
+
+## Summary
+
+This document provides details for integration with OCI Oracle Kubernetes Engine to implement a Cilium Direct Routing solution.
+
+## Motivation
+
+Many customers are requesting support for Cilium on OCI. This feature proposal will provide guidance on various implementation options.
+
+## Goals
+
+* Discuss possible solutions for integrating OCI with Cilium IPAM solutions for Direct Routing
+* Determine short term solutions for providing an integration with Cilium
+* Determine long term solutions for providing an integration with Cilium where users can specify which VNIC a pod can route out of (see examples below for details)
+
+## Non-Goals
+
+* Timelines of integrations
+
+## Proposal
+
+### Background
+
+OCI Oracle Kubernetes Engine (OKE) provides controls for the user to determine how many VNICs should be attached to a node and which VNIC a pod should route out of.
+
+*Note: The existing OCI OKE solution uses a pre-attach model, where the Cloud Controller Manager preallocates all VNICs/IPs and does not perform any detaches (detaches happen upon node termination), and then the CNI only sets up OS level network routing to the pods*
+
+#### Example 1
+
+A node has 1 VNIC with 256 IPs attached to it. Host and Pod traffic will route out of this VNIC.
+
+```
++--------------------------------------+
+|         Kubernetes Node              |
+|                                      |
+|   +--------+                         |
+|   | VNIC 1 |                         |
+|   +--------+                         |
+|       |                              |
+|   +--------+                         |
+|   |  Pod A |                         |
+|   +--------+                         |
+|                                      |
+|  Pod A routes traffic via VNIC1      |
++--------------------------------------+
+```
+
+#### Example 2
+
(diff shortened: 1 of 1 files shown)


cilium/CFP-42453-oci-cloud-provider-design.md
```diff
@@ -0,0 +1,173 @@
...
+
+For out-of-tree solutions, there may be additional integration testing.
+
+### Key Question: Cilium Testing
+
+How does Cilium perform CNI testing? What solution will require the least amount of maintenance/coordination?
+
+### Key Question: Cloud Provider Support Model
+
+What does the support model look like for Cloud Providers for integrated pieces, like the Cilium Operator?
```
antonipp: I will let Cilium maintainers correct me, but from what I've seen running Cilium on multiple Cloud Providers over the years is the following:
- For the Cilium AWS Operator, AWS itself doesn't seem to be involved (I believe they prefer to invest into https://github.com/aws/amazon-vpc-cni-k8s). So the burden of maintaining the AWS Operator falls on Isovalent + the users. We (Datadog) have been very active in maintaining the Operator because we heavily rely on it. This seems to be the case for other big users as well, such as Palantir.
- For Azure, I believe Azure used to maintain the Operator but then decided to move to the Delegated IPAM model so that they can maintain their code out of tree: https://github.com/Azure/azure-container-networking. We still use the Azure Operator and are not planning to switch to the Delegated IPAM model, so we actively maintain it in Cilium. I don't really know if there are any other big users.
- For GCP, they have implemented the first option you listed in this doc ("Kubernetes IPAM solution"), so they maintain their code in CCM: https://github.com/kubernetes/cloud-provider-gcp/tree/master/cmd/cloud-controller-manager

There's also the Alibaba Cloud integration but I don't really have context around it

antonipp (commented):
Thank you for the write-up! 

Just to give some perspective, as a Cilium user running self-managed Kubernetes on 3 Cloud Providers in parallel, I'd really prefer a solution consistent with the UX we have in other Cloud Providers, meaning Option 1 ("Kubernetes IPAM solution"), like in GCP, or Option 2 ("In-tree Extending the Cilium Operator"), like in AWS and Azure.

I think Option 3 is just more complex Option 1, I am not sure it's worth it? And Option 4 ("Delegated IPAM") makes sense from the code maintenance perspective for Isovalent (and I know they've been pushing it, like here: https://github.com/cilium/cilium/issues/34604#issuecomment-2325949413). 

However, as a user, it really makes life much harder: the delegated IPAM model adds way more complexity. The setup is different from what we are running in other Cloud Providers, and there are more moving parts. It will be harder to debug and to operate, because we'd need to run this additional IPAM management system alongside the regular Kube and Cilium components. The failure modes will be different as well: for example, all Cloud Provider API interactions will now be made in a distributed fashion instead of being centralized in the Operator.

Moreover, the current Operator Cloud Provider-specific IPAM code is mostly relying on core IPAM logic which has been tested for years, has many edge cases already figured out and which is still maintained. The only things which are less maintained are the Cloud Provider interface implementations. However, the code there is not too complex (it’s just implementations of `InstancesManager` and `Node` interfaces). And I think relying on support from large enterprise users, is not such a big issue.

joestringer:
I think @antonipp summarized the state pretty well :+1: The one thing I would add around option (2) is just that it has been a struggle to find people willing to help maintain IPAM code. The cloud SDKs are large dependencies that we have minimal understanding about, and I see that as a maintenance burden and risk for the project as we try to keep them up to date. I'm not excited about the idea of adding yet another cloud SDK.

cilium/CFP-42453-oci-cloud-provider-design.md
joestringer: I don't recall any Azure engineers being involved in the Azure IPAM implementation in Cilium. I believe the story was similar for both AWS and Azure with a combination of Isovalent and community user participants in developing the implementation.

---

On the topic of the highlighted text from this thread, what do you mean by "support model"?

trungng92:
> On the topic of the highlighted text from this thread, what do you mean by "support model"?

Thanks for the replies. What @antonipp discussed is what I was looking after. Essentially, in a solution where there's an in-tree OCI integration with Cilium, who becomes responsible for the OCI integration? Cilium? (Of course I wouldn't want to put additional burden on your team 🙂) OCI? The community?

> I think @antonipp summarized the state pretty well 👍 The one thing I would add around option (2) is just that it has been a struggle to find people willing to help maintain IPAM code. The cloud SDKs are large dependencies that we have minimal understanding about, and I see that as a maintenance burden and risk for the project as we try to keep them up to date. I'm not excited about the idea of adding yet another cloud SDK.

In a typical integration between two services (Service A and Service B), ideally:

- If Service A adds a new feature, they validate that the feature works with Service B
- And if Service B adds a new feature, they validate that the feature works with Service A

But I know this doesn't always happen in the real world (as @antonipp gave good examples for). Also, it's not a 1:1 model, it's a 1:many model. I can understand that testing every cloud provider for every feature would be difficult for Isovalent which could lead to features getting stale/left behind for certain cloud providers.

I am strongly leaning towards attaching a CIDR block and updating the `v1.Node.spec.podCIDR` field (option 1) in CCM, as it is the simplest solution to maintain (from OCI's point of view, as long as Isovalent maintains the contract of processing the `podCIDR` field) and will work for a majority of our customers.

The main drawback is that it isn't compatible with a multi-NIC solution. At the same time though, there doesn't seem to be _any_ "easy" solutions for multi-NIC use cases, and any multi-NIC solution will require a deeper in-tree or out-of-tree integration. Aside from cloud providers implementing multi-NIC, does Cilium have any expectations around NIC usage? Is the default behavior/expectation that Cilium will just use the default ip route as specified by the OS?

xmulligan:
Cc: @rbtr

cilium/CFP-42453-oci-cloud-provider-design.md
rbtr: I don't think Azure contributed to the Operator, our integrations started with delegated IPAM (I have worked on this from the Azure side since it was just an idea we had).

@antonipp have you had problems with delegated IPAM? Can you elaborate on your experience with delegated IPAM wrt this:
> However, as a user, it really makes life much harder: the delegated IPAM model adds way more complexity.

Azure delegated IPAM is only available in AKS where we manage everything. I expect OCI OKE would offer a similar managed experience. It's not intended for a self-managed cluster, but only because it requires special access to the networking controlplane only available in AKS so that we can offer advanced fabric integration (eg for Azure Overlay networking). 

However, I built lots of the delegated IPAM implementation for AKS and I don't think it's more complex than operating Cilium already is. We run one additional daemonset that drops the azure-ipam binary and the CNI conflist that plugs it in. It would be trivial for you to install, operate, and debug, if you can do these for Cilium already. 

For me, there's a clear separation of responsibilities - Cilium does the node-local CNI things it's good at, and the cloud provider owns the tight native integration. Doing this through standard interfaces is the only way that makes sense - Cilium talking to the cloud provider(s) directly is never going to be maintained in a way that makes everyone happy.

rbtr:
> The main drawback is that it isn't compatible with a multi-NIC solution. At the same time though, there doesn't seem to be _any_ "easy" solutions for multi-NIC use cases, and any multi-NIC solution will require a deeper in-tree or out-of-tree integration. Aside from cloud providers implementing multi-NIC, does Cilium have any expectations around NIC usage? Is the default behavior/expectation that Cilium will just use the default ip route as specified by the OS?

We've also been exploring multi-NIC and so far are only offering it via AzCNI because we didn't want to dump a bunch of impl specific code in Cilium and there was no standard contract for it.

But isn't this the promise of DRA(NET) and NRI?

cilium/CFP-42453-oci-cloud-provider-design.md
antonipp: For context, we have not tried running delegated IPAM in our infra because as you mentioned it looks like it wasn't designed for self-managed clusters but only for AKS. I thought about trying it out because at one point there were talks about deprecating the Azure Operator IPAM model which we were using. The only reason why delegated IPAM looked more complex to me is that it indeed requires more moving parts, i.e. the deamonset + the binary and it makes things way different from what we already have in AWS and Azure, where it's exactly the same model with the Cilium Operator managing everything.

trungng92:
I don't have experience with NRI, but we have looked into DRA a bit.

For anyone who needs a background on Dynamic Resource Allocation, the idea is that you can add "attributes" to devices (e.g. NICs). And then in your pod, you can request devices that meet specified attribute requirements.

It does seem useful for multi-NIC cases, although it might require the consumable capacity feature to be usable (multiple pods connecting to the same VNIC):

https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#consumable-capacity

Perhaps the answer to generic multi-NIC support for Cilium should be to wait for this feature to be available.

cilium/CFP-42453-oci-cloud-provider-design.md
trungng92: > Cilium does the node-local CNI things it's good at, and the cloud provider owns the tight native integration.

This was my original thought as well, and is kind of how the initial (non multi-vnic) solution will work, where OCI OKE will attach the IPs, give them to Cilium, and let Cilium do all of its "node-local CNI things". The delegated IPAM would have a similar frame of mind, where the IPAM would have the ability to select which IP to use (e.g. which VNIC to pick an IP from), and pass that onto Cilium. 

Although I think DRA (mentioned in the thread) may provide a way for users to specify a multi-nic approach when it provides consumable-capacity (using DRA, a pod could request 2 different VNIC devices).

joestringer:
I understand from the community meeting that @HadrienPatte and @trungng92 have made some progress on this issue. Would be interested to hear the results of this investigation.

Will mark this as draft for now as we'd need to see an update in order to develop this idea further.

trungng92:
Thanks for the message. Ultimately OCI/OKE went in the direction of using the Kubernetes IPAM solution (populating v1.node.spec.podCIDR) for our short term solution.

This was the most attractive short term solution given that it was cloud agnostic and a relatively simple solution compared to others, which meant the least amount of long term maintenance.

I've updated the pull request proposal doc to reflect this.

We are holding off on evaluating long term Multi NIC solutions until we have a greater need for it, but it's still on our radar.

PR #79, opened 2025-11-05
CFP-42466-change-cilium-pod-ip-pool-crd.md

kyounghunJang:
# Related
- Issue: [cilium/cilium#42466](https://github.com/cilium/cilium/issues/42466)

### DIFF
--- a/cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
+++ b/cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
@@ -1,0 +1,118 @@
+# CFP-42466: change-multi-ip-pool-crd
+
+**SIG: SIG-NAME** ([View all current SIGs](https://docs.cilium.io/en/stable/community/community/#all-sigs))
+
+**Begin Design Discussion:** 2025-11-04
+
+**Cilium Release:** 1.19
+
+**Authors:** kyounghoonJang <matkimchi_@naver.com>
+
+**Status:** Implementable
+
+**Reference** this CFP was translated from Korean to English using AI.
+
+## Summary
+
+Currently, CiliumPodIPPool defines CIDRs as simple strings, preventing granular allocation control. This proposal enhances the schema to support intra-CIDR IP reservation via a reservedRange field. This approach allows users to block allocation for specific parts of a CIDR rather than the entire subnet, offering greater flexibility for handling orphaned IPs and managing pool migrations.
+
+## Motivation
+
+When using Multi-Pool IPAM, operators sometimes need to decommission or replace existing CIDRs.
+The only current way to stop IP allocation from a CIDR is to remove it from the pool entirely, which can cause existing pod IPs to become orphaned and trigger warnings in the operator.
+
+To support a two-step migration workflow, we propose enabling CIDRs to be marked as “non-allocatable”:
+
+1. Update the old CIDR with a reservedRange covering its available IP space.
+
+2. Drain and migrate workloads to nodes using the new CIDR.
+
+3. Once the old CIDR is no longer in use, remove it from the pool.
+
+This allows for controlled and predictable transitions between IP pools without service disruption.
+
+## Goals
+Enhance the CiliumPodIPPool CRD schema to support a reservedRange field, allowing users to exclude specific IP ranges within a CIDR from being allocated.
+## Proposal
+
+### Overview
+
+Current structure:
+```yaml
+apiVersion: cilium.io/v2alpha1
+kind: CiliumPodIPPool
+metadata:
+  name: green-pool
+spec:
+  ipv4:
+    cidrs:
+      - 10.20.0.0/16
+      - 10.30.0.0/16
+    maskSize: 24
+```
+Proposed structure:
+```yaml
+apiVersion: cilium.io/v2
+kind: CiliumPodIPPool
+metadata:
+  name: green-pool
+spec:
+  ipv4:
+    cidrs:
+      - cidr: 10.20.0.0/16
+        reservedRange: 10.20.0.0-10.20.0.99
+      - cidr: 10.30.0.0/16
+    maskSize: 24
+```
+
+### Implementation Details
+
+- **Struct Change**: The cidrs field in spec will be changed from a simple []string to a list of objects: []struct { cidr: string, reservedRange: string }.
+  - cidr: The network CIDR.
+  - reservedRange: (Optional) A specific IP range within the CIDR to exclude from allocation.
+- **Allocator Logic**: The allocator will parse the reservedRange and exclude those specific IPs from the available pool.
+- **Migration Strategy (Lazy Migration)**:
+    - **Single Source of Truth**: The Operator will prefer the `v2` version of CiliumPodIPPool when both `v2` and `v2alpha1` objects exist. If only `v2alpha1` is present, the Operator will ingest the `v2alpha1` object.
+    In other words, `v2` takes precedence whenever available, and the Operator uses `v2alpha1` only as a fallback when no `v2` resource exists.
+    - **Lazy Conversion**: If automatic migration is enabled, the Operator will perform lazy conversion in the reconcile loop:
+      1. Detect v2alpha1 objects that have not yet been migrated.
+      2. Convert them into the v2 format and create a corresponding v2 resource.
+      3. Leave the original v2alpha1 object intact (not deleted) to avoid unexpected disruptions.
+    - **Configuration**: The migration behavior will be controlled through configuration (e.g., a value in the Helm chart or a field in the Operator’s ConfigMap), allowing operators to enable or disable automatic v2 migration.
+- **Upgrade Guide**: Automated downgrades are not supported. We will include a cautionary note in the documentation, advising users to proceed with migration only when they are confident.
+
+## Impacts / Key Questions
+
+### Impact 1 : Breaking Change - Schema Migration
+
+Since the schema has changed from storing CIDRs as simple strings in a slice to using a struct that contains both the CIDR and the reservedRange field, existing v2alpha users will encounter a schema mismatch issue.
+
+#### Key Question: A way to apply this feature without affecting existing users
+
+Implement this feature while promoting the MultiPool IP functionality to v2.
+
+##### Pros
+- Enables us to promote the API to v2 while introducing the new feature.
+
+##### Cons
+- The schema changes (string → object format) will be included directly in the official v2 release and may not have full maturity at the time of promotion.
+
+**Note:** To mitigate this concern, the new reservedRange field could be introduced under a beta the v2 API until stability is validated.
+
+#### Key Question: How do we handle resource conflicts and precedence during migration?
+If both a v2alpha resource and a v2 resource exist with the same name, how should the operator handle this conflict?
+
+It makes sense to prefer v2 resources, as v2 is generally considered the stable version.
+For v2alpha resources, the operator will leave v2alpha objects and log a warning.
+
+###### Pros
+- The operator’s conflict-handling logic becomes simpler.
+- Since the user’s CRs are not deleted directly, potential accidents or data loss are avoided.
+
+###### Cons
+- Stale CRs may remain unattended, which could lead to confusion.
+
+#### Key Question: What happens if the user has to downgrade to an older version of Cilium again
+
+The v2alpha1 objects are not deleted automatically.
+If a downgrade occurs, the older Cilium version will continue to read and reconcile these v2alpha1 objects as before


xmulligan:
cc: @cilium/sig-ipam

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,198 @@
...
+### Implementation Details
+
+#### CRD Changes
+
+Update the `CiliumPodIPPoolSpec` struct to include a new
+```go
+// pkg/k8s/apis/cilium.io/v2/ippool_types.go
+type CiliumPodIPPoolSpec struct {
+    CIDR string `json:"cidr"`
+    Unallocated []string `json:"unallocated,omitempty"`
```
gandro: This looks inconsistent with the example

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,198 @@
...
+
+
+### Impact 1 : CRD Version Promotion (`v2alpha1` → `v2`)
+The CiliumPodIPPool resource currently exists only in cilium.io/v2alpha1.Introducing per-CIDR configuration fields (e.g., unallocatable) requires changing the schema from a list of strings to a list of structured objects.This will likely require promoting the CRD to cilium.io/v2.
+
+This change affects backward compatibility and upgrade paths, as clusters using existing v2alpha1 resources might need migration logic.
+
+### Key Question: Should we promote the CRD to v2 directly or extend v2alpha1 first for testing?
+
+### Option 1: Promote directly to `v2`
```
gandro: I don't know what this means. What do we do with existing resources?

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Option 2: Keep `v2alpha1` and introduce the new schema there first
+
+#### Pros
+
+- Allows iterative development and testing based on community feedback.
+- Reduces migration and compatibility risks before the API promotion.
+
+#### Cons
+- May cause confusion or compatibility issues for existing users already relying on v2alpha1.
```
gandro: Multi-Pool has already been declared stable, so this is a non-starter.

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Impact 2: Operator Behavior Change
+The Cilium Operator currently assumes that all CIDRs in the pool are allocatable.
+Adding an unallocatable: true field changes allocation logic and reconciliation behavior.
+
+This will affect IPAM logic — especially how nodes select pools and how the operator reports “available IPs.”
+There may be a need to modify the allocator to skip unallocatable ranges.
```
gandro: I don't know what this section tries to say. Of course the operator needs to be changed, isn't that the whole point of the feature?

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Key Question: How should the operator treat CIDRs marked as unallocatable: true?
```
gandro: Isn't this the same as the above?

gandro (changes requested):
Thanks for the CFP. Did you AI to generate the CFP or parts of it? If so, please declare this.

I have a really hard time following the intent behind certain sections.

In general, I think the CFP should focus clearly on the CR changes, the intended behavior and most importantly on the migration path. Low-level implementation details are less relevant.

kyounghunJang:
@gandro 
Sorry for providing unclear context
I’ll make sure to clarify and refocus the CFP to clearly describe the CR changes, intended behavior, and migration path.
Thanks a lot for the feedback — I’ll update the proposal accordingly.

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,102 @@
...
+spec:
+  ipv4:
+    cidrs:
+      - 10.20.0.0/16
+      - 10.30.0.0/16
+    maskSize: 24
+```
+Proposed structure:
+```yaml
+apiVersion: cilium.io/v2alpha1
```
gandro: I think this should be `cilium.io/v2` - I don't think we can change the type of the `cidrs` list type otherwise, because it would break all users using `v2alpha1`.

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,102 @@
...
+- Create a struct to hold CIDR and the Unallocatable Flag.
+- Implement logic in the allocator to disallow pool allocation when the Unallocatable flag is set.
+
+## Impacts / Key Questions
+
+### Impact 1 : Breaking Change - Schema Migration
+
+Since the schema has changed from storing CIDRs as simple strings in a slice to using a struct that contains both the CIDR and the Unallocatable flag, existing v2alpha users will encounter a schema mismatch issue.
+
+### Key Question: A way to apply this feature without affecting existing users
```
gandro: I'm a bit confused by the headline structure here. If this belongs to the `Breaking Change - Schema Migration` section, it should likely be `#### Question: A way to apply this feature without affecting existing users` or similar.

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,102 @@
...
+
+To make this work, a Conversion Webhook must be implemented to translate API objects in real-time,
```
gandro: Not a big fan of webhooks, they can be very hard to operate. While it certainly can be an option, I think we have to have a mechanism that also works without a webhook. For example, I think we can also run some migration code in the operator.

However, I think there are a few questions how migration should work that need to be fleshed out before I think this CFP is implementable.

A few questions to solve out:

- If we translate all `v2alpha1` CRs to `v2` (or `v2alpha2`) CRs, what happens if the user already had a mix of objects and some objects cannot be translated because there already exists a `v2` version with the same name?
- What happens if the user has to downgrade to an older version of Cilium again? The `v2` schema would no longer work. We probably should only migrate resources for the user once they are certain they don't have to downgrade anymore.
- What type should the multi-pool allocator in the operator ingest by default? If I have two pools `v2/blue` and `v2alpha/blue` and a pod with the `ipam.cilium.io/ip-pool=blue` annotation, which one should it pick? Does it ingest both versions and give precedence to the new one? Does it only ingest one version that may be user configurable?

Unfortunately I don't have any experience migrating CRD schemas, but we've done it before. It might be worth looking at how Cilium (or other projects) have done this in the past.

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,102 @@
...
+
+This does not affect users who were using `v2alpha1`.
+
+### Option 1: Implement this feature while promoting the Multipool IP functionality to `v2`.
+
+#### Pros
+- It is possible to promote the API to v2 while introducing the new feature.
+
+#### Cons
+- The changes converting to objects would go into the official v2 release without stability verification.
```
gandro: I appreciate this concern. One potential option to mitigate is to mark the `unallocatable` flag in v2 as beta. That way, we can still deprecate it later.

gandro (commented):
Thanks for reworking the CFP! I found it much easier to follow. I think we need more details on the migration though.

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
kyounghoonJang: @gandro 
Thank you for the feedback.

Here are my thoughts regarding the questions:

Q1 & Q3: I believe applying v2 is the correct approach, as v2 is generally considered the stable version compared to v2alpha. If both exist, perhaps we can provide a warning log to users indicating that the v2alpha CRD will be ignored or deleted, and let the operator handle this logic.

Q2: I looked into ways to automate downgrades but couldn’t find a suitable approach. There are many considerations to take into account. In my opinion, the best course of action is to include a cautionary note in the documentation, advising users to proceed with migration only when they are confident.

Would it be okay to update the CFP with this content?

gandro: > Would it be okay to update the CFP with this content?

Yes, sounds good to me.

With regards to downgrade - maybe we could leave the v2alpha1 objects around? They might be stale of course, but if the v2 and v2alpha1 objects are identical, then one can downgrade without issues. After all, changes should be rather rare.

kyounghoonJang: @gandro 
I also think that approach makes sense!

gandro: (@kyounghunJang Just a quick heads-up - I'll be away for a week. I hope other members of @cilium/sig-ipam are able to take a look, otherwise I'll review again once I'm back)

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,132 @@
...
+  
+  **Note:** To mitigate this concern, we could introduce the `unallocatable` flag in `v2` under a beta designation
+
+##### Option 2: Upgrade to `v2alpha2` first
+
+###### Pros
+- Allows testing the object conversion in a pre-release environment.
+
+###### Cons
+- Additional maintenance and migration steps are required before the official v2 release.
```
gandro: It's also still not clear how this can be achieved without loss of data.

gandro (commented):
Thanks for the update. One thing that still seems unaddressed is the question of what the actual migration process is.

I would expect that the operator converts the existing `v2alpha1` objects to `v2`, and then ingests the `v2` objects as described in the CFP. But when does it do that conversion? How do we make sure it is only done once?

kyounghunJang:
> Thanks for the update. One thing that still seems unaddressed is the question of what the actual migration process is.
> 
> I would expect that the operator converts the existing `v2alpha1` objects to `v2`, and then ingests the `v2` objects as described in the CFP. But when does it do that conversion? How do we make sure it is only done once?

I think we can prevent duplicate processing by using an annotation flag.
For example, something like `cilium.io/migrated-to-v2="true"` should work.

Regarding when the operator should perform the conversion, I believe there are two possible approaches:

1. **Run the conversion during the operator’s initialization phase**
This reduces the number of management points, but if the amount of data to migrate is large, it may delay the operator’s startup.

2. **Perform a lazy migration**
Add the migration logic to the reconcile loop. This way, the operator’s initialization is not impacted, and the migration can be handled more flexibly over time.

Personally, I think the second approach is better.
What are your thoughts on this?

Thank you!

gandro:
In the current operator architecture, we don't really have an initialization phase. There are Hive start hooks, but they cannot be used for long running tasks. Therefore I don't think option 1 is really feasible.

The important bits to consider are:
a) Make sure migration can be turned off for users who don't want/need it.
b) Make sure the multi-pool allocator correctly copes with pools being migrated (which I think should just work)
c) What's the semantics of the annotation? Does it mean we no longer look at that pool at all? What happens if v2alpha1 and v2 differ? Probably best to assume v2 objects are the source of truth post migration.
d) Document clearly when it it safe for the user to delete the old v2alpha1 objects (i.e. migration has succeeded without errors, and a downgrade to earlier Cilium versions for any reason (potentially unrelated to multi-pool) no longer is needed).

kyounghunJang:
@gandro

Sorry for the late reply. I’d appreciate your thoughts on whether the following approach makes sense.

a) I think the migration behavior can be controlled through Helm values or a ConfigMap. For example, we could introduce a flag like enableV2Migration.

c) The annotation I had in mind would indicate whether the user’s CRD has been successfully migrated from v2alpha to v2.
If the flag is set to true, it means the migration from v2alpha to v2 was completed successfully, and the object should now follow the v2.
Additionally, if both v2 and v2alpha objects exist at the same time, I believe we should ensure that only the v2 object is used.

Thank you!

gandro:
>  Additionally, if both v2 and v2alpha objects exist at the same time, I believe we should ensure that only the v2 object is used.

Yes. I don't think we want to ingest both v2 and v2alpha objects concurrently, that will lead to unnecessary complexity. I think the operator should only ingest v2, and the migrator ingests v2alpha and produces v2.

Given that I think we have discussed the major discussion points now, could you update the CFP the reflect the decisions being taken? In other words, I don't think some of the questions are open questions anymore, and we have decided on an approach. 

Let's document that approach that you are planning to take for the implementation, and then once everyone is happy with that plan we can move the CFP into implementable state.

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,120 @@
...
+      - cidr: 10.30.0.0/16
+    maskSize: 24
+```
+
+### Implementation Details
+
+- **Struct Change**: The `cidr` field in `spec` will be changed from `[]string` to `[]struct { cidr: string, unallocatable: bool }`.
+- **Allocator Logic**: The allocator will check the `unallocatable` flag and skip those CIDRs during IP allocation.
+- **Migration Strategy (Lazy Migration)**:
+    - **Single Source of Truth**: The Operator will prefer the `v2` version of CiliumPodIPPool when both `v2` and `v2alpha1` objects exist. If only `v2alpha1` is present, the Operator will ingest the `v2alpha1` object.
```
gandro, line 75: IMO, we should only only ingest v2. If only `v2alpha` objects are present, let's rely on the conversion.

gandro (approved):
Unblocking the "changes requested". I think this is in a good enough state to be implementable.

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,120 @@
...
+
+**SIG: SIG-NAME** ([View all current SIGs](https://docs.cilium.io/en/stable/community/community/#all-sigs))
+
+**Begin Design Discussion:** 2025-11-04
+
+**Cilium Release:** 1.19
+
+**Authors:** kyounghoonJang <matkimchi_@naver.com>
+
+**Status:** Draft
```
xmulligan:
```suggestion
**Status:** Implementable
```

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,120 @@
+# CFP-42466: change-multi-ip-pool-crd
+
+**SIG: SIG-NAME** ([View all current SIGs](https://docs.cilium.io/en/stable/community/community/#all-sigs))
```
xmulligan, line 3:
```suggestion
**SIG: SIG-Datapath**
```

cilium/CFP-42466-change-cilium-pod-ip-pool-crd.md
```diff
@@ -0,0 +1,120 @@
...
+
+**Note:** To mitigate this concern, the new unallocatable field could be introduced under a beta the v2 API until stability is validated.
+
+#### Key Question: How do we handle resource conflicts and precedence during migration?
+If both a v2alpha resource and a v2 resource exist with the same name, how should the operator handle this conflict?
+
+It makes sense to prefer v2 resources, as v2 is generally considered the stable version.
+For v2alpha resources, there are two possible handling options
+
+##### Option 1: Leave v2alpha objects and log a warning
```
pippolo84: Nit: My understanding is that we agreed on keeping the v2alpha1 resources around (see [this comment](https://github.com/cilium/design-cfps/pull/79#discussion_r2522211335)), so this should be updated to describe only the `Leave v2alpha objects and log a warning` option.

pippolo84 (approved):
LGTM minus a minor nit. Thanks! 💯

networkop:
what do you think about the following use case: instead of only controlling the availability of the entire CIDR, allow users to specify parts of that CIDR? so something like this:

```yaml
apiVersion: cilium.io/v2
kind: CiliumPodIPPool
metadata:
  name: green-pool
spec:
  ipv4:
    cidrs:
      - cidr: 10.20.0.0/16
        reservedRange: 10.20.0.0-10.20.0.99
      - cidr: 10.30.0.0/16
```

This would enable a use case where PodIPPool CIDR is shared with an external entity, e.g. Pods are attached and use [L2 pod announcement](https://docs.cilium.io/en/stable/network/l2-announcements/#l2-pod-announcements) while the default gateway resides in both Cilium and the TOR switch. This would allow users to disable the assignment of default gateway IP and potentially any other IPs from the same CIDR/prefix. (there was a similar use case described in https://github.com/cilium/cilium/issues/31600)

kyounghunJang:
@gandro @pippolo84 
I also agree with the suggestion that using reservedRange allows for more flexible handling.
If everyone is aligned, I can update the CFP based on this approach.


> what do you think about the following use case: instead of only controlling the availability of the entire CIDR, allow users to specify parts of that CIDR? so something like this:
> 
> ```yaml
> apiVersion: cilium.io/v2
> kind: CiliumPodIPPool
> metadata:
>   name: green-pool
> spec:
>   ipv4:
>     cidrs:
>       - cidr: 10.20.0.0/16
>         reservedRange: 10.20.0.0-10.20.0.99
>       - cidr: 10.30.0.0/16
> ```

gandro:
@kyounghunJang Thanks for addressing the feedback. 

Could you change the state to `Implementable`? And maybe address the feedback by Fabio here https://github.com/cilium/design-cfps/pull/79#discussion_r2602987034

> [!NOTE]
> Implementable: CFPs that have been approved by one committer and merged into this repository are given the implementable status. This means that the ideas in the CFP have been agreed in principle and the coding work can now be started. When merged, the CFP should be up to date and the relevant stakeholders should be in alignment even if they are still going through the practical ramifications of how to implement it.

I assume while implementing the CFP, there might be some changes to the schema that may be required during implementation or code review, but I feel we have overall agreement on what to change and how to get there. We can modify further changes to the CR in this document even after this PR has been merged.

Once you change the `Status` to implementable, I will merge it.

Side-note: I'm about to be out of office yet again until next year, feel free to already start working on the code even if it isn't merged yet. Assuming no last minute discussion emerge, I will merge it once I'm back.

gandro:
Merging this as implementable as per the comment above. Thanks!

PR #80, opened 2025-11-05
CFP-4191: Tetragon per-workload policies

Andreagit97:
## Summary

Today, with Tetragon, it is not possible to share a common enforcement logic across many Kubernetes workloads. For each workloa,d users need to create a separate `TracingPolicy` with the same enforcement but with different values. This design proposes a new model to decouple the enforcement logic from per‑workload values, enabling significant reductions in eBPF programs and map memory usage.

...

[continued in CFP]

### DIFF
--- a/tetragon/CFP-4191-per-workload-policy.md
+++ b/tetragon/CFP-4191-per-workload-policy.md
@@ -1,0 +1,143 @@
+# CFP-4191: Tetragon per-workload policies
+
+**SIG:** SIG-TETRAGON
+
+**Begin Design Discussion:** 2025-11-05
+
+**Tetragon Release:** X.XX
+
+**Authors:** Andrea Terzolo <andrea.terzolo@suse.com>, Kornilios Kourtis <kkourt@cisco.com>
+
+**Status:** Draft
+
+## Summary
+
+Today, with Tetragon it is not possible to share a common enforcement logic across many Kubernetes workloads. For each workload users need to create a separate `TracingPolicy` with the same enforcement but with different values. This design proposes a new model to decouple the enforcement logic from per‑workload values, enabling significant reductions in eBPF programs and map memory usage.
+
+## Motivation
+
+The current approach of one `TracingPolicy` per workload leads to:
+
+- Program scaling limits: At least one eBPF program per policy. In clusters with hundreds of workloads, this may add latency when the hook is the hot path.
+- Map memory scaling: Tens of maps per policy and several large maps (for example, policy filters, socktrack_map, override_tasks) lead to multi‑MB memlock per policy.
+- Operational friction: Logic is identical but values differ (for example, allowed binaries), yet the model duplicates programs and maps.
+
+## Goals
+
+- Decouple shared enforcement logic from per‑workload values to avoid linear growth in programs and map memory as clusters scale.
+- Retain TracingPolicy expressiveness (selectors, filters, actions).
+
+## Non-Goals
+
+- TBD
+
+## Proposal
+
+### Overview
+
+This design is accompanied by a [proof of concept](https://github.com/cilium/tetragon/issues/4191). It introduces two concepts: templates and bindings.
+
+A template is a `TracingPolicy` that declares variables populated at runtime rather than hardcoded at load time. Selectors reference these variables by name.
+
+```yaml
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "block-process-template"
+spec:
+  variables:
+  - name: "targetExecPaths"
+    type: "linux_binprm" # this could be used for extra validation but it's probably not strictly necessary
+  kprobes:
+  - call: "security_bprm_creds_for_exec"
+    syscall: false
+    args:
+    - index: 0
+      type: "linux_binprm"
+    selectors:
+    - matchArgs:
+      - index: 0
+        operator: "Equal"
+        valuesFromVariable: "targetExecPaths"
+```
+
+Deploying a template alone has no runtime effect because it lacks concrete values for comparisons.
+
+A binding is a new resource (for example, `TracingPolicyBinding`) that supplies concrete values for a template’s variables and targets specific workloads through a selector. The selector is not the same as a podSelector; it is intended to be mutually exclusive across bindings.
+
+```yaml
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicyBinding
+metadata:
+  name: "block-process-template-values-1"
+spec:
+  policyTemplateRef:
+    name: "block-process-template"
+  exclusiveSelector: # it should be mutually exclusive across bindings
+    matchLabels:
+      app: "my-app-1"
+  bindings:
+  - name: "targetExecPaths"
+    values:
+    - "/usr/bin/true"
+    - "/usr/bin/ls"
+```
+
+Policy logic becomes active only when a `TracingPolicyBinding` is present. The binding populates eBPF maps with the specified values for cgroups that match its selector.
+
+### Details
+
+- When a template is deployed, Tetragon loads the same eBPF programs and maps as today. Additionally, it creates a `BPF_MAP_TYPE_HASH` map (`cg_to_policy_map`) to map cgroup IDs to binding `policy_id`s. Initially, the map is empty, so the template has no effect.
+- When a binding is deployed:
+  - It receives a new `policy_id`.
+  - For each cgroup matching its selector, Tetragon inserts a (`cgroup_id` → `policy_id`) entry into `cg_to_policy_map`. If a cgroup already has a binding for that template, it is rejected.
+  - This mapping activates the template logic for the targeted cgroups.
+  - Binding values are stored in `BPF_MAP_TYPE_HASH_OF_MAPS` instances (`pol_str_maps_*`). This implementation is very specific to string/charbuf/filename types and the eq/neq operators, but the concept can be extended to other types/operators, more on this later.
+  - These maps are keyed by `policy_id` (looked up via `cg_to_policy_map`). The value is a hash set of strings sourced from the binding, using the same 11‑bucket sizing technique as existing `string_maps_*`.
+
+### Results
+
+This design enables:
+
+1. Single eBPF program: One shared program can serve many bindings (for example, 512–1024+), as they reference the same template. This drastically reduces programs loaded into the kernel.
+2. Low memory overhead: Per‑binding cost is small—entries in `cg_to_policy_map` plus `pol_str_maps_*` (typically a few KB per binding for modest value lists).
+
+## Impacts / Key Questions
+
+### Impact: new `TracingPolicyBinding` CRD
+
+Adding a `TracingPolicyBinding` CRD introduces a new concept. Users must understand the relationship between templates and bindings and how to manage them effectively.
+
+### Key Question: Is the new CRD necessary?
+
+Could existing `TracingPolicy` mechanisms be extended to carry only values and reference a shared logic template, or is a dedicated CRD essential for clarity and usability?
+
+### Impact: selectors in bindings are intended to be mutually exclusive
+
+In this model, a `cgroup_id` can be associated with at most one binding (`policy_id`) per template. A new binding for the same cgroup is rejected. Binding the same cgroup to multiple policies simultaneously is not allowed.
+
+### Key Question: Is this an acceptable use-case?
+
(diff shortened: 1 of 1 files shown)


tetragon/CFP-4191-per-workload-policy.md
```diff
@@ -0,0 +1,143 @@
+# CFP-4191: Tetragon per-workload policies
```
kkourt, line 1: Based on reading the first proposal draft and considering the discussion we already had at https://github.com/cilium/tetragon/issues/4191,  there are two different sub-problems.

- *P1*: The  BPF implementation to support multiple policies for different workloads scales poorly (e.g., program scaling limits and map memory scaling) 
- *P2*:  The UI/UX of writing policies for different workloads is cumbersome (operational friction)

My suggestion would be to treat them as separate problems when discussing solutions. 

One way to think about Tetragon policies that, I think, is useful is as

```
(workload, operation, object) -> action
```

_Workloads_ are defined with selectors such as `podSelector`, `matchBinary`, `matchNamespace`, `matchCredentials`.
_Operations_ are defined with things like `kprobes.call`,  _objects_ are defined with things like `matchArgs/matchBinaries`, while _actions_ with things like `Post`, `NoPost`, `Override`, `NotifyEnforcer`. 


### Scaling the BPF implementation (P1)

The current Tetragon implementation creates (and attaches) one BPF program per TracingPolicy resource.
This has limitations. For one, a limited number of programs can be attached to every hook. Moreover, there is a lot of redundant computation  (e.g., if we have `100` policies, we will check `100` times if the cgroup of the workload is part of the corresponding policy; this can be optimized so we control the path so we don't have to check every time), as well as duplicated information in case the policies are similar (e.g., only the values in `matchAgs/matchBinaries` differ) 

Potential solutions:
 - Avoid redundant computations and reduce memory footprint by sharing maps across different TracingPolicy resources
 - tail-call only into the programs of the policies matching a given workload
 - ??? 

### The UI/UX of writing policies for different workloads is cumbersome (P2)

The challenge here, I think, is one where policies might have common sections of `Workloads`, `operations`, or `objects`, and we want to avoid having the user re-write them.

Potential solutions:
 *  Split what is now a single TracingPolicy into a parent/template and multple children policies. The parent/template contains the common parts, while the children parts contain the different parts (typically per workload).
 * Extend the current selectors (e.g., allow for `PodSelectors` not only at the top of the policy, but also under `selectors`)
 * ???
 
 
 ### Approach
 
Given all the above, I would see this CFP as a way to specify solutions to the above problems (as well as how they interact), their tradeoffs, and how we (i.e., the Tetragon OSS project) can move forward to address them. An alternative approach would be to focus on the specific solution that you propose in https://github.com/cilium/tetragon/issues/4191. Either approach is fine by me, but we should define (and agree on) the goals/scope.

kkourt (commented):
Thanks!

tetragon/CFP-4191-per-workload-policy.md
Andreagit97: I agree, we can split this into 2 different sub-problems. More in detail, I would address P1 first since the solution to this problem could give us precious hints on how to design the UX.

> Given all the above, I would see this CFP as a way to specify solutions to the above problems (as well as how they interact), their tradeoffs, and how we (i.e., the Tetragon OSS project) can move forward to address them.

I would pick this approach

> An alternative approach would be to focus on the specific solution that you propose in https://github.com/cilium/tetragon/issues/4191. Either approach is fine by me, but we should define (and agree on) the goals/scope.

I would not focus on the proposed solution cilium/tetragon#4191 but would seek to explore alternatives like the "tail-call" approach you suggested, to understand the advantages it can provide.

If you agree, as a next step, I would compare the tail call solution with the proposed one, evaluating their pros and cons, and see if any other ideas come up from the discussion. WDYT?

kkourt: Cool!

Maybe as a first step, we can structure the CFP in a way that reflects the above? That is, separate the two problems and have sections for proposed solutions?

CC: @tpapagian who can provide more details on the tail call solution.

tpapagian: First I will provide some background on what we have done already (for a different context). It seems similar to that but it also has some differences.

Assuming a policy similar to the following
```yaml
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "example-policy"
spec:
  podSelector:
    matchLabels:
      app: "workload-a"
  kprobes:
  - call: "some_call"
    syscall: true
    args: [...]
    selectors: [...]
```
we wanted to have a large number of such policies, each of them for a separate workload (defined under `podSelector`) that all hook at the same function (same `kprobes.call`) but have different selectors. As you already mentioned, the problem is that this will result in multiple programs, each of them attached on that function, and each program has its own maps etc. The main issue was that if we have 100 such policies, we always need to run all 100 programs and this will affect the performance of the users workload.

The approach that we took was to have a mapping from `cgroup_ids` to `policy_ids`. This was introduced in https://github.com/cilium/tetragon/pull/3122 (map named `policy_filter_cgroup_maps`). We also need a map from `policy_id` to `program` (type `BPF_MAP_TYPE_PROG_ARRAY`) that has all the mapping from `policy_id` to the actual eBPF programs. 

Now we need a single eBPF program that is attached to the hook that we care about. This will get the policy ids (from `policy_filter_cgroup_maps`) and call `bpf_tail_call` helper to the first one. When the first program is done, this will again call `bpf_tail_call` to the second, etc.

When a new policy is loaded, we just need to load a new program with its maps, that corresponds to the selectors for this tracing policy, and update the `BPF_MAP_TYPE_PROG_ARRAY` to add a new mapping. But we don't need to attach that to any hooks.

On the one hand, this makes scalability (in terms of CPU overheads) with the number of tracing policies to be constant (only the programs that needs to run will be called). We also managed to do that without the need to change the UI/UX. It was also relatively easy to implement. On the other hand, this will not use less maps (i.e. less memory) compared to what we have now. 

I tried to keep it at high-level to make that easier to follow. It you feel that anything is not clear, just let me know to provide more details.

holyspectral: (Sorry for deleting the previous messages because it wasn't thought through well)

@tpapagian thanks for your reply and I think it's a great idea. It's also great to see that Tetragon already has part of the implementation.

I have a question about the bpf_tail_call. If I understand it correctly, Linux has a maximum 33 bpf_tail_call depth limit from [MAX_TAIL_CALL_CNT](https://elixir.bootlin.com/linux/v6.14.3/source/include/linux/bpf.h#L1976) macro.

I think Tetragon has used many `bpf_tail_call`.  For example, `TAIL_CALL_ARGS` and `TAIL_CALL_FILTER` would be called many times depending on the policy. Do you think it would fit the 33 maximum if we have multiple `policy_ids`?

Andreagit97: Thank you @tpapagian, for the detailed explanation. 
Today, in the code, I see that the `policy_filter_cgroup_maps` map is instantiated once in the policy filter state, so it is shared by all eBPF programs. Is this map also global in your approach?

Let's say we are in a specific hook `security_bprm_creds_for_exec` and we are the `cgroup1`. When we look up the policy IDs in `policy_filter_cgroup_maps`, how can we understand which policies are associated with the particular hook we are calling from (`security_bprm_creds_for_exec` in this example)? I imagine that a cgroup could be associated with different policies using different hooks.


> On the one hand, this makes scalability (in terms of CPU overheads) with the number of tracing policies to be constant (only the programs that needs to run will be called). We also managed to do that without the need to change the UI/UX. It was also relatively easy to implement. 

This sounds really promising! What I really like is that all selectors are supported out of the box without the need to create new maps of maps to handle values from different policies

> On the other hand, this will not use less maps (i.e. less memory) compared to what we have now.

Do you think that this limitation can be overcome in some way, or do you believe it is not feasible with this approach? Having 2/3 MB for each policy still seems a little bit too much when you have many workloads in the cluster. Putting it another way, do you already have any plans for how to mitigate it, and how we might help?

Andreagit97:
> Maybe as a first step, we can structure the CFP in a way that reflects the above? That is, separate the two problems and have sections for proposed solutions?

Sorry for the delay. Done in the last commit

tetragon/CFP-4191-per-workload-policy.md
Andreagit97: Any update on this one?

kkourt: > Let's say we are in a specific hook security_bprm_creds_for_exec and we are the cgroup1. When we look up the policy IDs in policy_filter_cgroup_maps, how can we understand which policies are associated with the particular hook we are calling from (security_bprm_creds_for_exec in this example)? I imagine that a cgroup could be associated with different policies using different hooks.

That's a good point. Not sure what the best answer to this is. My first thought is to have per-hook information about the policies that use that hook. Let us think about it and we can add a concrete proposal under Option "2: Tail Call policy chains"

tpapagian: Hey @Andreagit97.

> Is this map also global in your approach?

Yes.

> how can we understand which policies are associated with the particular hook we are calling from (security_bprm_creds_for_exec in this example)?

Good question. As @kkourt said we need to have some metadata about that. We can change `policy_filter_cgroup_maps` to have what we need there.

> Do you think that this limitation can be overcome in some way, or do you believe it is not feasible with this approach? Having 2/3 MB for each policy still seems a little bit too much when you have many workloads in the cluster. Putting it another way, do you already have any plans for how to mitigate it, and how we might help?

I don't have all the details for generic tracing policies in mind, but I agree that 2/3 MB per-policy can be too much. I believe that there are ways to improve that (possibly orthogonal to what we are discussing here) but I don't have a concrete proposal yet.

tpapagian: Hey @holyspectral,

> I think Tetragon has used many bpf_tail_call. For example, TAIL_CALL_ARGS and TAIL_CALL_FILTER would be called many times depending on the policy. Do you think it would fit the 33 maximum if we have multiple policy_ids?

In what I have already done, we do not have any additional tail calls. For generic tracing let's say that at maximum 10 tail calls are already used (depends on the number of arguments as far as I am aware of) which leave us with ~23. But this also depends on the use-case that you have in mind. This limit is on how many policies will run for a specific cgroup-id and not how many policies are loaded in general. 

If you try to optimise things, you also try to run a small amount of policies per workload. So in general this would be enough I believe. Do you have in mind a use-case where this should not be enough?

holyspectral: Hi @tapapagian, 

Thanks for your reply!  It's great to know that the proposal won't introduce new tail calls.  I think my concern was that, in the existing design, each read_arg or argSelector is already a tail call.  For a simple policy with 2 arguments to read, 2 argSelector and 1 action, it would take 1 `TAIL_CALL_SETUP` call, 2 `TAIL_CALL_PROCESS` calls, 2 `TAIL_CALL_ARGS` calls.  If any of them matches, there will be `TAIL_CALL_ACTIONS` call and `TAIL_CALL_SEND` call.

So totally 5-7 tail calls for each policy in our typical use case.

I mean, it's a significant improvement to have 6x less ebpf hooks, but it's a little less than what we expected.    

Do you think it's feasible to make these policies share the same `TAIL_CALL_SETUP` call, or even `TAIL_CALL_PROCESS` calls?  By sharing these calls we may be able to also reduce the memory utilization from percpu arrays.

joestringer:
@cilium/tetragon-maintainers this PR came up during periodic review during the Cilium community meeting, but that meeting is mainly focused on cilium/cilium so we skipped past it. But maybe for clarity has this moved forward? Do we think this proposal needs updating or changing in some way, or is the investigation into this solution ongoing?

kkourt:
> @cilium/tetragon-maintainers this PR came up during periodic review during the Cilium community meeting, but that meeting is mainly focused on cilium/cilium so we skipped past it. But maybe for clarity has this moved forward? Do we think this proposal needs updating or changing in some way, or is the investigation into this solution ongoing?

It hasn't moved forward, unfortunately. @Andreagit97 any plans to continue working on this?

Andreagit97:
Ei! No, not in the immediate future. At the moment, it’s not one of our top priorities.

joestringer:
No worries! I'll mark this PR as draft then, but feel free to bring it back up for discussion whenever you're ready.

PR #81, opened 2025-11-05
Add CFP for ordered policy

squeed:
This demonstrates a possible implementation of ordered policy.

### DIFF
--- a/cilium/CFP-23380-ordered-policy.md
+++ b/cilium/CFP-23380-ordered-policy.md
@@ -1,0 +1,265 @@
+# CFP-23380: Ordered network policy
+
+**SIG: sig-policy**
+
+**Begin Design Discussion:** 2025-11-01
+
+**Cilium Release:** 1.19
+
+**Authors:** Casey Callendrello <cdc@isovalent.com>, Jarno Rajahlme <jarno@isovalent.com>
+
+## Summary
+
+Add explicit, user-supplied rule ordering to the policy engine.
+
+## Motivation
+
+Most firewall engines present an _iterative_ interface to their users. Users provide an _ordered_ list of rules. In comparison, Cilium policy is _commutative_, which is to say, the ordering of rules does not matter.
+
+Iterative policy languages are more expressive. For example, consider the rules:
+
+> 1. `env=qa, role=grafana`: allow access to `env=prod, role=artifacts`
+> 2. `env=qa`: deny access to `env=prod`
+
+These sorts of rules are awkward to express via the Cilium policy language, where Deny-based rules always take priority.
+
+## Goals
+
+* Provide a foundation for the ClusterNetworkPolicy effort
+* Allow expressing L3/L4 rules with order
+* Add an explicit priority field to policy Rules
+* Keep the datapath O(1)
+
+## Non-Goals
+
+* Add an ordered engine to the datapath
+
+## Existing mechanism
+
+Before outlining the proposed changes, it is important to understand the current policy resolution mechanism in detail. The basic shape:
+
+1. From API objects, policy is lightly translated to an `[]PolicyEntry`
+2. These rules are converted to a `L4Policy`, which is a `map[(proto,ports)]L4Filter`. The `L4Filter` is a `map[selector]PerSelectorPolicy`. `L4Filter` contains some bookkeeping information. The "policy judgement" is the `PerSelectorPolicy`. Informally, this can be considered a `map[(proto, port(s), selector)]PerSelectorPolicy`.
+3. The `L4Policy` is then converted to a `MapState`, which is (logically) a `map[port+proto]map[identity]MapStateEntry`. The `MapStateEntry` is the actual configuration for the datapath.
+
+
+### Existing logic: PerSelectorPolicy generation
+
+When there are no existing entries, the translation from a `PolicyEntry` to its `L4Policy` and `PerSelectorPolicy` is straightforward. Most of the logic is in how to merge two `PerSelectorPolicies`. The `PerSelectorPolicy` merge logic primarily evaluates per-field precedence. For example, deny overrides all, auth overrides no-auth, L7 overrides no L7. In the event of a conflict (e.g. different L7 parsers for the same port), policy resolution fails. L7 rules are the exception: they are vectors that support merging.
+
+Envoy directly consumes the `L4Policy`, the distilled per-selector policy. However, the BPF policy engine expects a different structure: the PolicyMap.
+
+### Existing logic: Generating the PolicyMap
+
+The policy map as stored in BPF is a map from `(identity, proto, port) -> MapStateEntry`. Thus, the `L4Policy` must be partially inverted to a per-identity policy result. This translation is not purely mechanical; it must also be priority aware.
+
+Why must PolicyMap generation be priority aware? After all, `PerSelectorPolicy` merging *should* have included all the precedence logic. However, disjoint selectors may select the same identity. Thus, when resolving the `MapStateEntry` for a `(id, proto, port)` key, it may be that a tiebreaker is needed between two `MapStateEntries`.
+
+## Proposal
+
+This proposal outlines a way to add explicit order-based precedence to L3/L4 policies.
+
+There are two key sections for this proposal. The first is the technical implementation of policy ordering. The second is an outline of the algorithm for generating an endpoint-specific BPF PolicyMap from an ordered set of policies.
+
+### Ordered PerSelectorPolicy resolution
+
+Quick summary:
+- Generate a partial order of PerSelectorPolicies, based on specified order and tie-breakers. Assign an integer priority.
+- Merge PerSelectorPolicies with the same order number, using existing precedence logic.
+
+The first step is to generate an ordered list of PerSecectorPolicies. This should be sorted according to explicit order, then tie-broken according to the existing priority logic. With this order, a priority is assigned.
+
+Note that the end-user order need not be the internal priority referenced in this design. In fact, most APIs support non-integer orderings.
+
+Then, PerSelectorPolicies with the same selector should be combined. This means there will be exactly one PerSelectorPolicy for the key `(proto, port, selector)`. The priority should be used to select the "winning" selector. If both have the same priority, the existing logic for merging selectors will be used. This prevents conflicts while determining L7 policies, and preserves existing behavior, where each policy can be considered to have priority 0.
+
+
+### Generating the MapState
+
+Quick summary: When inserting entries in to the MapState, any longer-match, lower-priority entries must be deleted.
+
+Once the flattened set of PerSelectorPolicies has been created, it must be translated to the MapState. Recall that the MapState is almost equivalent to `map[proto + port]map[identity]MapStateEntry`, where `MapStateEntry` is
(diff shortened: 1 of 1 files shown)


squeed:
@TheBeeZee, @jrajahalme  PTAL

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,263 @@
...
+    - For every selected identity:
+        - Insert in to the MapState if valid
+        - Delete any now-invalid entries
+        - Propagate authentication mode to descendants
+
+The existing MapState already has a form of this algorithm, in the implicit priorities brought by deny and proxy policies. This CFP proposes eliminating the split logic for deny and proxy priorities, unifiying them in a single 32-bit Priority field that merges user order, proxy priority, and deny.
+
+### MapState conflict resolution
+
+The key observation is this: **we cannot insert a lower-priority, higher-precedence entry in to the MapState**. Since MapState precedence mirrors datapath precedence, determining this is relatively straightforward: to insert a given key, we scan all ancestor keys, checking if the same identity is selected. If a lower-precedence, higher-priority entry is found, we cannot insert the higher-precedence, lower-priority entry. Conversely, if any descendants of the newly proposed entry are higher-precedence and lower-priority, they must be deleted.
```
jrajahalme: We should define what we mean with `priority` and `precedence`. In ClusterNetworkPolicy lower numeric value "takes precedence" (is considered before any rules with higher numeric "priority"). So "high priority" rules have lower numeric priority value. This is in agreement with language like "this is priority 1 task" (i.e., the highest priority 
task in general language).

Furthermore, "higher-precedence" reads as "taking precedence", i.e., being *more* important. So in written language "higher-priority" has the same meaning as "higher-precedence".
 
 This is hard get right, but we should aim for clarity here by e.g., using the general sense that "higher-priority" == "higher-precedence", and refer to a "numerically low priority value" explicitly.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,263 @@
...
+
+Note: higher numerical priority overrides lower numerical priority.
```
jrajahalme: ref above. In my PoC implementation I have adopted the k8c CNP view on the API side, where lower priority values override higher ones, but on the datapath I have expanded the notion that higher numerical values override lower numerical values from `ProxyPortPriority` when generalizing it to a single `Precedence` field. The reason for the latter is the clarity in code and comments that comes from the ability to have "higher precedence" mean "numerically greater" and using the greater than operator (`>`) for comparisons in straightforward manner.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,263 @@
...
+- if no authentication is specified, inherit it from the closest ancestor of equal or higher priority
+- if authentication is specified, "push" it to all descendants of equal or lower priority
+
+### Processing incremental updates
+
+In incremental update is when a complete policy newly selects, or no longer selects, one or more identities. These incremental updates must then be propagated down to the MapState and the BPF map updated accordingly.
+
+Incremental updates are aggregated by selectors and then inserted in to the MapState. The MapState is responsible for deconflicting overlapping selectors. This remains the same in the ordered version.
+
+It is safe to insert incremental updates in this manner because identities and selectors are immutable. As a direct consequence, two selectors S1 and S2 might overlap, a new identity I1 can never change state from non-overlapping to overlapping.
```
jrajahalme: S1 and S2 might be non-overlapping to begin with, e.g., by selecting nothing. Then I1 appears and is selected by both S1 and S2, making them overlapping. The key for correct incremental updates is that, for the purposes of mapstate updates, S1 and S2 both select any overlapping identities (e.g, I1). This would naturally be the case if we evaluated the selections from ground up each time, but with our use of selector cache we must make sure that both cached selectors are updated at the same transaction, so that the readers (the mapstate update procedure) observe the new, overlapping identity in both S1 and S2. Historically this was not the case which made the mapstate update much more complicated than it is now.

jrajahalme (changes requested):
Need some clarifications.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,263 @@
...
+
+Whether or not a connection requires mututal authentication is *also* managed by the network policy engine. However, authentication is not a "traffic verdict". (Traffic verdicts are `allow`, `deny`, and `proxy`). Rather, it is applied to all traffic, regardless of selectors.
+
+When inserting a mapstate entry, the authentication option needs to be resolved:
+- if no authentication is specified, inherit it from the closest ancestor of equal or higher priority
+- if authentication is specified, "push" it to all descendants of equal or lower priority
+
+### Processing incremental updates
+
+In incremental update is when a complete policy newly selects, or no longer selects, one or more identities. These incremental updates must then be propagated down to the MapState and the BPF map updated accordingly.
```
TheBeeZee: In -> An

cilium/CFP-23380-ordered-policy.md
TheBeeZee: +1

There were many discussions in the network-policy-sig about this and in the end it was decided that going with "priority" with the associated semantics (lower numeric value is higher priority) is the most intuitive approach (despite the inversion of numeric value and the real meaning of "priority").

It is of course possible to invert the k8s Cluster Network Policy priorities when translating into an internal representation, but it seems like keeping Cilium internals in sync with what the public APIs do is a net-win.

Explicitly defining what "precedence" and "priority" mean in this context would help the readability of the document (and code). Also, if the choice will be for priority to be "numerically higher means higher priority", then I would recommend against using the name "priority", as this will conflict with Cluster Network Policy semantics.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,263 @@
...
+    rule2-. deletes .-> rule1
+```
+
+#### Wildcard and non-wildcard interaction
+
+Special care must be taken for identity wildcards, both when higher and lower precedence.
+
+If a lower-priority non-wildcard PerSelectorPolicy would override an existing wildcard, higher-priority, lower-precedence selector, adding it is blocked.
+
+If a lower-priority wildcard PerSelectorPolicy would override an existing non-wildcard, higher-priority, lower-precedence selector, then it must be inserted, but this will have to be separately resolved via the datapath. Specifically, the datapath must look up both wildcarded and non-wildcarded entries, and select the one with the highest priority, using precedence length as a tiebreaker.
```
TheBeeZee: This might be my lack of understanding of the Cilium datapath, but is this currently handled or is this CFP proposing to implement it?

Also, how about non-wildcard overlaps, for example with CIDR rules? Those are already handled by LPF matching in the dataplane? For example, low priority deny for 10.0.0.0/8 and high priority allow for 10.10.10.0/24?

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,263 @@
...
+    rule1 -. invalid .-> rule2 & rule3
+```
+
+#### Example 2b: Descendant wildcard
+
+Two rules:
+1. `(priority 2, id 5, tcp, port *)`
+2. `(priority 1, id * tcp, port 80)`
+
+Rule 2 has a longer match. For ID 5, rule 2 overrides rule 1, which is an error. However, rule 2's wildcard selector selects more identities, so it cannot be elided. Rather, at policy resolution time, the numeric priorities will be compared.
```
TheBeeZee: Why is this an error? Again, might be my lack of understanding of the current semantics. Seems like something a user might configure. Also, this says "at policy resolution time, the numeric priorities will be compared". Is this "policy resolution time" the control plane or dataplane?

cilium/CFP-23380-ordered-policy.md
jrajahalme: Cilium datapath does this currently for deny and proxy port precedence, the proposal is to expand this to a more generic notion of "precedence".

Cilium policy engine operates on "security identities", not on CIDRs directly. Each unique CIDR (or a group of them via CIliumCIDRGroup) will get mapped to a unique security identity, in the datapath before policy enforcement. In your example `10.0.0.0/8` would have a different security identity than `10.10.10.0/24`, and if a packet would be covered by both of them, the ipcache will assign the security identity of `10.10.10.0/24` to the packet for policy enforcement.

On the policy computation side, the selector for `10.0.0.0/8` will select both of these security identities since it covers also the `10.10.10.0/24` CIDR, so the low priority deny entry is inserted for both of these identities. For the identity of `10.10.10.0/24` that low priority deny entry would be overridden by the high priority allow entry, so the bpf datapath policy map will only have the high priority allow entry for the `10.10.10.0/24` CIDR identity.

This also expands to port (range) policies due to the LPF policy map.

cilium/CFP-23380-ordered-policy.md
jrajahalme: "policy resolution time" here is at the dataplane (cilium bpf datapath). Maybe it would be clearer to write "policy enforcement time" instead? "policy resolution" could be understood to mean "policy map computation time".

I think the "is an error" here should be maybe written as "would be an error if the numeric identities were not compared at the policy enforcement time" etc.

cilium/CFP-23380-ordered-policy.md
squeed: Right, based on the existing PoC implementation, priority is an unsigned integer value, where a higher number overrides a lower number. This is unrelated to any particular API convention, where, generally, a lower number overrides.

While this inversion is a bit confusing, it does ultimately simplify the low-level core. I'll add some commentary about it.

squeed: Clarified the wording, PTAL

cilium/CFP-23380-ordered-policy.md
squeed: I added a distinction between user API ordering and datapath priority, and made it clear that an ordering API is not part of this CFP.

cilium/CFP-23380-ordered-policy.md
squeed: Clarified a bit.

cilium/CFP-23380-ordered-policy.md
TheBeeZee: I don't see any changes, did you forget to push?

squeed: I did. Sorry, PTAL.

cilium/CFP-23380-ordered-policy.md
```diff
@@ -0,0 +1,265 @@
...
+
+**Example 1d:** Can we insert `(priority 1, identity 5, udp, port 53)`?
+    Yes, because this does not cover a higher-priority entry.
+    
+```mermaid
+---
+title: "Example 1d: valid"
+---
+graph TD
+    rule1[prio 1, id *, tcp, 0-1023]
```
TheBeeZee, line 167: I think the priorities are swapped here, rule1 should be prio2 and rule2 should be prio1, no?

joestringer:
Was this synced to the latest design? Are you still looking for feedback?

joestringer:
Moved to draft due to lack of activity.

PR #85, opened 2025-12-12
CFP-43295: Cilium Network Driver

bersoare:
adding CFP for the Cilium Network Driver feature
https://github.com/cilium/cilium/issues/43295


cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,304 @@
...
+```
+---
+apiVersion: cilium.io/v1
+kind: CiliumNetworkDriverConfig
+metadata:
+  name: cilium-network-driver-config
+spec:
+  selectors:
+    labels: 
+       - cilium.io/network-driver
```
YutaroHayakawa: As we agreed on the [PR](https://github.com/cilium/cilium/pull/43797), let's clarify the behavior of the conflicting nodeSelector. The idea coming up in the thread now is:

1. Choose the oldest configuration (with oldest creationTimestamp). If that doesn't break the tie, compare the alphabetical order of the name.
2. Don't make a choice when there's a conflicting nodeSelector. Whenever we see multiple configuration selecting the same node, nothing will be applied.

I don't have much opinion around which is better. Just for the reference, in the BGP Control Plane, [we chose 2](http://docs.cilium.io/en/latest/network/bgp-control-plane/bgp-control-plane-troubleshooting/#additional-new-ciliumbgpclusterconfig-is-not-working). The main limitation here is when there's 1 configuration and another conflicting configuration appears, it will clear all applied configuration and report error. Meaning the on-going BGP session will be broken. The upside is it is easier for users to notice the misconfiguration. With 1. it's difficult for users to notice it as it "works" somehow.

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,304 @@
...
+          - ifName: enp2s0f1np1
+            vfCount: 6
+```
+
+Under the deviceManagerConfigs section, an operator is able to control how a specific device manager is set up. 
+In this context, the device manager is an abstraction of a certain type of resource. In the example below, 
+you can see that we’re working with the `sriov` devicemanager - implying that we must account for extending 
+the feature set by introducing device managers.
+
+Updating the configuration at runtime is possible for the most part. Some settings, for example,
```
YutaroHayakawa, line 174: In the [PR](https://github.com/cilium/cilium/pull/43797) the configuration is not updatable at all until the agent restarts. Is it part of the specification or just an implementation detail? What makes it challenging to update the configuration?

bersoare: in the sriov device manager, a user is allowed to express how PFs should be configured. specifically, the number of sr-iov vfs that are to be created. adjusting this number can't be done without unconfiguring the device first (deleting all VFs), and if there is any workload making use of VFs in there would see their interface disappearing. 
on the other hand, another part of the configuration relates to how devices are grouped and advertised, and it is totally fine to change that at runtime and shouldn't even require a re-initialization of the whole thing.
it is the combination of these two different properties that makes a decision on how to handle configuration selection tricky.
in the PR, it is true that we're not dealing with running updates but that is intentional simply because we want to keep the scope small on each piece of work (smaller PRs). to me, it makes logical sense to figure out how to choose a config among many first, and only then deal with dynamic updates.
hope it makes sense

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: approach 2 sounds better to me (except for the part of clearing the running config) - and i wonder if there is a way we can signal the error state without affecting workloads (with taints? statuses?)

i think this is probably what @pippolo84 suggested

bersoare: took some time to write down thoughts about it, and here is what i came up with:

Requirements:

- operators must be able to apply different configuration to different nodes
- only one config is selected at a time
- operators must be notified in case conflicts occur
- operator is ultimately in control of the selection
- documentation is explicit about the possible scenarios and caveats

Problem:

Using labels matching in order to select which object applies to which node is the standard way of indicating where certain things should go. In the network driver, the CiliumNetworkDriverConfigSpec takes in a nodeSelector, which has the purpose of making an agent aware that a certain configuration should be applied by matching the node labels.

Problem arises when we consider the lifecycle of an agent and how configurations are handled:

- Running cilium without any configuration detects a set of configurations coming in
    - Depending on the order which the configurations arrive, if they are created separately or together, or if they are deployed sparsely for a period of time can affect which one is chosen.
- If a running agent with a configuration applied restarts, it may select a different configuration from the one it had selected before. this could lead to unintended behaviors in the agent.
- Merging configurations add to complexity as there has to be some logic to know what is supposed to be done about conflicts, and the more configuration options there are, the more complexity is added.
    - also a possible footgun for the operator
- Assigning quality of matching logic can be un-intuitive, as a “best match” when it comes to label selection is sometimes subjective
    - `nodeName=A` vs (`nodeName in [A,B]` and `cluster in [A,B,C,D]` )
    

Proposal:

That said, in order to give full control to the operator on which configuration should be applied, we can make the agent:

- Select config only once on startup
    - requiring a restart needs an explicit action from the operator.
- When starting, if no config is found right away - terminate the network driver initialization. Again, require a restart in case the feature is to be enabled.
    - As an operator, I deploy my configurations and when I am ready, i restart the cilium pods to apply the configuration. This allows me to prepare for setting up changes and control the deployment/rollback procedure with measurable impact.
- If a running agent with a config detects a new config with a nodeSelector coming in that could also match the node - the operator must be made aware but without affecting the running workloads (no eviction, no reconfiguration trigger).
    - from here, if a new configuration is picked after a restart is because the operator wanted to. they are in control

cilium/CFP-43295-cilium-network-driver-dra.md
YutaroHayakawa: Ok, I understand this is more like a matter of the controller implementation and not by design. Just for your information, there's a way to make certain fields in the CRD immutable [with CEL](https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/). Perhaps you can leverage this kind of the mechanism to avoid undesirable change (like changing the number of VF allocation).

cilium/CFP-43295-cilium-network-driver-dra.md
YutaroHayakawa: What happens when there are conflicting configurations at start-up? If we follow the [Fabio's suggestion](https://github.com/cilium/cilium/pull/43797#discussion_r2713942356), we don't select any configuration until the conflict is resolved? And during the resolution, we won't touch to any already published ResourceSlice?

I think it's ok to go with this restart-driven reconciliation mechanism as long as we have a plan to support dynamic update in the future. One thing we should keep in mind this restart-driven reconciliation will become a part of user interface. Supporting dynamic update while users are anticipating the update only happens on restart is a breaking change. We need a proper opt-in knob and deprecation process at that point.

YutaroHayakawa: Wrt error reporting, we could rely on the status field of the CiliumNetworkDriverConfig resource, but in general, writing to the status of the single resource from multiple controller (in this case, all selected node's Cilium agent) is not a good idea. This is because each status update fan-outs the watch notification to all controllers watching this resource (in this case all nodes in the cluster) which does not scale in the large cluster.

I think what we can do is like:

1. Introducing per-node resource derived from the CiliumNetworkDriverConfig (say CiliumNetworkDriverNodeConfig) and individual node writes to this per-node resource (this is what BGP Control Plane does for `CiliumBGPClusterConfig` vs `CiliumBGPNodeConfig`).
2. Report status in the agent health and users need to check it with `kubectl exec` or via dedicated tool that does `kubectl exec` (`cilium bgp peers` cilium-cli command is implemented in a similar way).

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,304 @@
...
+  driverName: "sriov.cilium.k8s.io"
+  deviceManagerConfigs:
+      sriov:
+        enabled: true
+        ifaces:
+          - ifName: enp2s0f0np0
+            vfCount: 6
+          - ifName: enp2s0f1np1
+            vfCount: 6
+```
```
joestringer, line 127: Side note related to the [thread above](https://github.com/cilium/design-cfps/pull/85/files#r2715257457), but I didn't want to derail the detailed discussion: If we had status here we can have the cilium-operator scan nodes, check for label constraints, and update the status on the object to inform of conflicts. Similar pattern to the policy validation we do from cilium-operator.

````suggestion
status:
  conditions:
    type: valid
    status: false
    reason: validation check
    message: This resource conflicts with resource Y
    lastTransitionTime: XXXX-XX-XX XX:XX:XX+00:00
```
````

Comment (bersoare) [reply cilium/CFP-43295-cilium-network-driver-dra.md:127]: that solves the problem of writing to the status field safely....but at the cost of involving the cilium-operator. i'd prefer to avoid adding one more element to the mix and keep the complexity contained to the agent itself.

@YutaroHayakawa [suggested](https://github.com/cilium/design-cfps/pull/85#discussion_r2719190342) that we could use a per node object to report the status for each node without dealing with concurrent writes, and i think it is a good one given that we wont need to involve the operator. any thoughts on this?

File: cilium/CFP-43295-cilium-network-driver-dra.md (hunk 1 above)
Comment (bersoare) [reply cilium/CFP-43295-cilium-network-driver-dra.md]: > What happens when there are conflicting configurations at start-up? If we follow the [Fabio's suggestion](https://github.com/cilium/cilium/pull/43797#discussion_r2713942356), we don't select any configuration until the conflict is resolved? 
> 

yes, this is what i was thinking. it should not impact any running workloads as there won't be another scheduling event due to the missing resourceslices

> And during the resolution, we won't touch to any already published ResourceSlice?
> 

no, on a restart, the DRA plugin wont register itself and it wont advertise anything. so the node will essentially present no resources to the cluster.


> I think it's ok to go with this restart-driven reconciliation mechanism as long as we have a plan to support dynamic update in the future. One thing we should keep in mind this restart-driven reconciliation will become a part of user interface. Supporting dynamic update while users are anticipating the update only happens on restart is a breaking change. We need a proper opt-in knob and deprecation process at that point.

with 


> 1. Introducing per-node resource derived from the CiliumNetworkDriverConfig (say CiliumNetworkDriverNodeConfig) and individual node writes to this per-node resource (this is what BGP Control Plane does for `CiliumBGPClusterConfig` vs `CiliumBGPNodeConfig`).
> 2. Report status in the agent health and users need to check it with `kubectl exec` or via dedicated tool that does `kubectl exec` (`cilium bgp peers` cilium-cli command is implemented in a similar way).

even better, i believe we can then use the per-node resource to remember the last configuration picked up. we can avoid any surprise changes by making the agent using a `CiliumNetworkDriverNodeConfig` if it had written one in the past (and the selected configuration still exists). if one doesn't exist, select a config and write one out.
so, in the case of configuration updates, we can have the operator in control of the config selection (delete the `CiliumNetworkDriverNodeConfig` in order to re-select a config) while avoiding the requirement of an agent restart.

so, back to the `Proposal` part of my previous comment, we could do something like:

- Sync all configurations matching the local node by labels.
- On startup, find `CiliumNetworkDriverNodeConfig` for the local node.
  - if found, ensure the configuration referenced in the `CiliumNetworkDriverNodeConfig` is selected
  - if not, perform the selection algorithm* 
- if any configuration is selected, write `CiliumNetworkDriverNodeConfig` with the selected config and any statuses (ex: conflicts or other errors)

If there is no configuration or `CiliumNetworkDriverNodeConfig` resources matching the node, block until a `CiliumNetworkDriverConfig` comes in.
  
 
* algorithm specifics still TBD 

let me know what you think

Comment (pippolo84) [reply cilium/CFP-43295-cilium-network-driver-dra.md]: I agree with the usage on a node-specific `CiliumNetworkDriverNodeConfig`. I think having that additional object makes it immediately clear to the operator what is happening even with conflicting multiple configs.

However, I'm thinking about [Joe's suggestion](https://github.com/cilium/design-cfps/pull/85#discussion_r2719192830). I feel that the "filtering and selection" logic that ultimately leads to the creation of the node-specific config object should leave in the operator. IMHO this simplifies the implementation in the agent/network driver that should care only about implementing a certain config, without bloating it with a possibly complex selection logic. IOW I'd move all of this:

> - Sync all configurations matching the local node by labels.
> - On startup, find `CiliumNetworkDriverNodeConfig` for the local node.
>   - if found, ensure the configuration referenced in the `CiliumNetworkDriverNodeConfig` is selected
>   - if not, perform the selection algorithm* 
> - if any configuration is selected, write `CiliumNetworkDriverNodeConfig` with the selected config and any statuses (ex: conflicts or other errors)

in the operator and let the agent only watch its `CiliumNetworkDriverNodeConfig` object, reacting accordingly.

Moreover, the operator is already (at least indirectly) involved into the network driver implementation, since Multi Pool Resource IPAM requires an operator component.

File: cilium/CFP-43295-cilium-network-driver-dra.md (hunk 3 above)
Comment (joestringer) [reply cilium/CFP-43295-cilium-network-driver-dra.md:127]: Only thought is that as a general rule we avoid reporting per-node status from cilium agent into the k8s apiserver because it adversely impacts scale. We've had this problem in the past for other object types that we generate too much load on the control plane. That's the motivation for the operator-based approach used by network policy, as long as the validation can be done statically. If there is node-specific information though that's not available centrally, we've just opted to report the status within the agent / via logs etc instead. The usability is poorer but then we don't have to deal with scale implications.

Comment (YutaroHayakawa) [reply cilium/CFP-43295-cilium-network-driver-dra.md:127]: It definitely has a scale implication for the large cluster with many nodes. In BGP Control Plane, we have a rate limit (once per 5 sec) for status write and can [opt-out](https://docs.cilium.io/en/latest/network/bgp-control-plane/bgp-control-plane-operation/#disabling-crd-status-report) it as needed (because this is just informational and ultimately can obtain the same information using `kubectl exec`). If we go with per-node resource way and report the status, maybe better to have this kind of safe guard.

File: cilium/CFP-43295-cilium-network-driver-dra.md (hunk 1 above)
Comment (YutaroHayakawa) [reply cilium/CFP-43295-cilium-network-driver-dra.md]: > no, on a restart, the DRA plugin wont register itself and it wont advertise anything. so the node will essentially present no resources to the cluster.

But, the ResourceSlice from the previous agent run is already persisted on the k8s API right? I guess then we need to retain it without touching it.

Comment (YutaroHayakawa) [reply cilium/CFP-43295-cilium-network-driver-dra.md]: > On startup, find CiliumNetworkDriverNodeConfig for the local node.
if found, ensure the configuration referenced in the CiliumNetworkDriverNodeConfig is selected
> if not, perform the selection algorithm*

I think what I suggested (writing conflict status to the NodeConfig) was a bit weird, sorry. I think we can take the mixture of what Joe suggested below. Operator can deal with the NodeConfig creation, detect the selector conflict during the reconciliation and write conflict status of the `CiliumNetworkDriverConfig` (maybe better to rename it like `CiliumNetworkDriverClusterConfig`?).

**Comment by YutaroHayakawa** [comment]
I think another thing I noticed was there's no much discussion done in this CFP about how to deal with the config update/delete. Since this CRD controls how we expose the DRA ResourceSlice, the ResourceSlice generated based on `CiliumNetworkDriverConfig` has a dependency to the specific version of the `CiliumNetworkDriverConfig` and we might have a DeviceClasses, ResourceClaim(Template)s, and Pods that are also depending on the specific version of the ResourceSlice.

```mermaid
flowchart LR
  dc0["CiliumNetworkDriverConfig"]
  rs0["ResourceSlice node0"]
  rs1["ResourceSlice node1"]
  devc0["DeviceClass"]
  rc0["ResourceClaim"]
  pod0["Pod"]
  dc0---rs0---devc0---rc0---pod0
  dc0---rs1---devc0
```

So, whenever the update/delete (and potentially selection/deselection by Node's label change, equivalent to update/delete from single node's point of view) happens, before performing the actual update/delete, we need to wait for all dependencies to be disappeared. I think it's worth covering how we want to handle this.

bersoare:
> I think another thing I noticed was there's no much discussion done in this CFP about how to deal with the config update/delete. Since this CRD controls how we expose the DRA ResourceSlice, the ResourceSlice generated based on `CiliumNetworkDriverConfig` has a dependency to the specific version of the `CiliumNetworkDriverConfig` and we might have a DeviceClasses, ResourceClaim(Template)s, and Pods that are also depending on the specific version of the ResourceSlice.
> 
> ```mermaid
> flowchart LR
>   dc0["CiliumNetworkDriverConfig"]
>   rs0["ResourceSlice node0"]
>   rs1["ResourceSlice node1"]
>   devc0["DeviceClass"]
>   rc0["ResourceClaim"]
>   pod0["Pod"]
>   dc0---rs0---devc0---rc0---pod0
>   dc0---rs1---devc0
> ```
> So, whenever the update/delete (and potentially selection/deselection by Node's label change, equivalent to update/delete from single node's point of view) happens, before performing the actual update/delete, we need to wait for all dependencies to be disappeared. I think it's worth covering how we want to handle this.

not sure i follow - can you elaborate more on that?

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: approach makes sense to me. i think we all managed to agree on a direction. i'll update the CFP on a commit, but tl;dr:

- Implement configuration selection and conflict management in the operator by looking at `CiliumNetworkDriverConfig` / `CiliumNetworkDriverClusterConfig` objects.
- Operator reports conflicts and issues back to the `CiliumNetworkDriverConfig`/`CiliumNetworkDriverClusterConfig` (with an on/off toggle) 
- Operator chooses matching configuration for each node and writes a `CiliumNetworkDriverNodeConfig` for each node.
- Node picks up the config and handles any necessary updates (details omitted for now as it is a separate problem to be tackled - i think this is what @YutaroHayakawa mentioned on https://github.com/cilium/design-cfps/pull/85#issuecomment-3792320815)

this way we can simplify the agent logic by moving the harder decision making to the operator, while also benefiting from a centralized management and reporting which improves the scaling

bersoare:
@pippolo84 , @YutaroHayakawa , @joestringer ,

i have updated the cfp taking into account the discussions we had in the PR on this [commit](https://github.com/cilium/design-cfps/pull/85/changes/de60b0277cbf197676a44250365e630c42cbc52d).

let me know what you think, and if there's anything i missed.
thank you so much for the 🥇 feedback

pippolo84:
> @pippolo84 , @YutaroHayakawa , @joestringer ,
> 
> i have updated the cfp taking into account the discussions we had in the PR on this [commit](https://github.com/cilium/design-cfps/pull/85/changes/de60b0277cbf197676a44250365e630c42cbc52d).

LGTM! :100:

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,360 @@
...
+metadata:
+  name: sriov-nodes-cluster-config
+nodeSelector:
+    matchLabels: 
+      sriov: true 
+spec:
+  driverName: "sriov.cilium.k8s.io"
+  deviceManagerConfigs:
+      sriov:
+        enabled: true
```
YutaroHayakawa: Nit: What is the use case of this explicit flag? I think defining this `sriov` stanza already implies that it is already enabled.

YutaroHayakawa:
> not sure i follow - can you elaborate more on that?

Sorry, never mind. I again started driver implementation discussion in the CRD spec discussion and I resolved all questions myself. Just for the reference, concrete scenarios in my mind (and my current understanding) are:

1. What happens to ResourceSlice if we delete the CiliumNetworkDriverClusterConfig while there are ResourceClaim using the devices derived from that ClusterConfig?

In this case, we can just delete the ResourceSlice and delay the actual deletion of the actual PF/VFs until all ResourceClaim disappears at driver level.

2. What happens to ResourceSlice if we increase/decrease the number of vfCount while there are ResourceClaim using the devices derived from that ClusterConfig?

In both of increase/decrease case, we should advertise the actual state of the SRIOV. When the driver is waiting for the device reconfiguration to prevent traffic disruption, old VFs are still advertised. When all dependency ResourceClaim disappeared, it performs the device reconfiguration and update the ResourceClaim with new VFs.

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: yea, i can change that - it was an unconscious decision

bersoare:
> > not sure i follow - can you elaborate more on that?
> 
> Sorry, never mind. I again started driver implementation discussion in the CRD spec discussion and I resolved all questions myself. Just for the reference, concrete scenarios in my mind (and my current understanding) are:
> 
> 1. What happens to ResourceSlice if we delete the CiliumNetworkDriverClusterConfig while there are ResourceClaim using the devices derived from that ClusterConfig?
> 

If a config delete happens, the node should re-initialize with an empty config. this will cause the node to stop advertising resources to the kube api, and that node will no longer be considered for scheduling if a device is needed for a workload.

> In this case, we can just delete the ResourceSlice and delay the actual deletion of the actual PF/VFs until all ResourceClaim disappears at driver level.
> 

the driver will not change the existing configuration on the PF - the VFs will not be cleaned up. and based on my experiments so far, existing pods can still function as they were. only new workloads will be affected - the scheduler will no longer consider the node for workloads coming in simply because it doesn't advertise any resources.

> 2. What happens to ResourceSlice if we increase/decrease the number of vfCount while there are ResourceClaim using the devices derived from that ClusterConfig?
> 

the resourceslice is a reflection of what currently is present on the node, regardless of anything in the sriov interfaces configuration. in other words, the devices found in the pci bus (a vf is a lightweight pci device) are advertised as is, even without their PF being referenced in the interface configuration. the only thing that can affect this is a matching pool.
that means that a cluster operator may prefer to manage VF creation on their own (via systemd or otherwise) and simply get them advertised in cilium.

> In both of increase/decrease case, we should advertise the actual state of the SRIOV. When the driver is waiting for the device reconfiguration to prevent traffic disruption, old VFs are still advertised. When all dependency ResourceClaim disappeared, it performs the device reconfiguration and update the ResourceClaim with new VFs.

increasing and decreasing is not possible. i chose to not implement that because changing the number of VF for a PF needs a reset of all the VFs (set number to zero -> set to the desired number). this action, if it takes place on a running node, will cause VF devices to disappear from running pods and this is pretty bad. see https://github.com/cilium/cilium/blob/feature/dra-driver/pkg/networkdriver/sriov/sriov.go#L475-L487


hope it clarifies! lmk if there's any questions

YutaroHayakawa:
> increasing and decreasing is not possible

JFYI we can enforce this at API validation level [with CEL](https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/). I guess this is way easier than tracking the state transition at controller level.

squeed:
>  in other words, the devices found in the pci bus (a vf is a lightweight pci device) are advertised as is, even without their PF being referenced in the interface configuration.

Perhaps I'm mis-understanding this statement, but AIUI the published ResourceSlice for a given node should only contain devices we are making available for pods.

bersoare:
> > in other words, the devices found in the pci bus (a vf is a lightweight pci device) are advertised as is, even without their PF being referenced in the interface configuration.
> 
> Perhaps I'm mis-understanding this statement, but AIUI the published ResourceSlice for a given node should only contain devices we are making available for pods.

that is correct - the sr-iov device manager currently finds all devices of that kind. but they are only advertised to the kube api in a resourceslice if the devices also match a pool (user configured). this part over here: https://github.com/cilium/design-cfps/pull/85/changes#diff-d7847da6e71b01e3ca0e845e7f46483c7b7ffc376792e06867214d9d5e433cc9R196-R204

let me know if it doesnt make sense

cilium/CFP-43295-cilium-network-driver-dra.md
squeed: Rather than a new type, what about using the `CiliumNode` object?

It could publish a list of candidate devices and their attributes. The operator could then evaluate the selectors and generate the per-node ResourceSlices.

squeed:
Hm. Now that I think a bit further about this, I wonder if we're reinventing the ResourceSlice / DeviceClass wheel. Perhaps you're right: we should publish all devices with all relevant attributes, and let the CEL in DeviceClass take care of selection.

Which brings me to my next thought: we should definitely make more device attributes available for selection. My first thought would be

- ifname
- mac (CEL allows selecting by prefix)
- driver
- PCI bus
- PCI vendor
- PCI model
- presence of default route

Either we can select these via the CiliumNetworkDriverClusterConfig or via DeviceClass.

Thoughts?

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,360 @@
...
+        ifaces:
+          - ifName: enp2s0f0np0
```
squeed: What sort of attributes would you expect users to be able to select? For example, the CNI host-device plugin supports selecting by device name, mac, sysfs path, or PCI ID. And UDEV supports all sorts of things, most useful being PCI vendor / model and driver.

bersoare: Right now, this is the set of attributes we're exposing for SR-IOV devices: https://github.com/cilium/cilium/blob/feature/dra-driver/pkg/networkdriver/sriov/sriov.go#L53-L63

a customer may choose any combination they see fit from that list. from my experience, picking a device based on the topology (ex: referencing a parent PF), vendor:device and driver seem to be the most common. but really it all comes down to how an operator wishes to pick devices and express it in a device class

pippolo84:
> Which brings me to my next thought: we should definitely make more device attributes available for selection. My first thought would be
> 
> * ifname
> * mac (CEL allows selecting by prefix)
> * driver
> * PCI bus
> * PCI vendor
> * PCI model
> * presence of default route
> 
> Either we can select these via the CiliumNetworkDriverClusterConfig or via DeviceClass.
> 
> Thoughts?

I definitely agree with exposing more attributes: that will give more flexibility to driver users that would like to be more specific when claiming device.
Adding new attributes is pretty easy: it is just a matter of extending the list of `DeviceAttribute` when scanning for devices on the node. IOW having a longer list does not need any change in the proposed driver structure, and we can extend the list even further in case new needs arise. :+1:

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,361 @@
...
+apiVersion: cilium.io/v1
+kind: CiliumNetworkDriverClusterConfig
+metadata:
+  name: sriov-nodes-cluster-config
+nodeSelector:
+    matchLabels: 
+      sriov: true 
+spec:
+  driverName: "sriov.cilium.k8s.io"
+  deviceManagerConfigs:
```
tamilmani1989: if deviceManagerConfigs not specified, would cilium dra driver finds all devices in node and publish it?

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,361 @@
...
+metadata:
+  name: mynode-hostname
+spec:
+  driverName: "sriov.cilium.k8s.io"
+  deviceManagerConfigs:
+      sriov:
+        enabled: true
+        ifaces:
+          - ifName: enp2s0f0np0
+            vfCount: 6
```
tamilmani1989, line 144: what does vfCount imply here? nic shared by 6 pods or this particular device has 6 VFs?

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,361 @@
...
+  name: enp2s0f0v0
+driver: sriov.cilium.k8s.io
+nodeName: c3-small-x86-01-bernardo
+pool:
+  generation: 1
+  name: a-side
+  resourceSliceCount: 1
+```
+
+Devices can be assigned to Pods by creating pods with a ResourceClaim statement in the pod manifest.
```
tamilmani1989, line 266: how devices are assigned, will the interface be moved directly inside pod via nri hook?

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: if the deviceManagerConfigs section is empty, no devices will be published.
in order to, for instance, publish sr-iov devices, one must enable the sr-iov device manager in the config

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: it means that we want the interface named enp2s0f0np0 to spawn 6 VFs - the vfs will then be advertised to the kube api in as long as they are assigned to a device pool

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: yes, the VF interface is moved to the pod NS if using a kernel driver.

cilium/CFP-43295-cilium-network-driver-dra.md
tamilmani1989: wouldn't be better if dra driver automatically scans all devices in node and publish it? why user has to specify devices to be published through CRD

bersoare: i prefer to not decide which devices can be advertised as available, and let the user set it up as they see fit

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: i have not added any details on the exact approach yet on the CFP - @pippolo84 has something in the code for it and i'll leave it to him to 👀  and see if all this still make sense

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -313,17 +333,15 @@ spec:
           opaque:
             driver: sriov.cilium.k8s.io
             parameters:
-              vlan: 123
-	            ipam_pools:
-                - pool-a
+              vlanID: 123
```
pippolo84, line 342: Since `vlanID` is a detail that might not always be useful for the common deployments, I think it belongs in the `CiliumResourceNetworkConfig` obj.

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,451 @@
...
+  spec:
+    devices:
+      config:
+        - requests:
+            - a-side
+          opaque:
+            driver: sriov.cilium.k8s.io
+            parameters:
+              vlanID: 123
+	            ipv4Address: 192.0.2.1/30
```
pippolo84: Nit: fix indentation

cilium/CFP-43295-cilium-network-driver-dra.md
pippolo84: The section about the driver configuration is up to date and still relevant. I think we detailed all the relevant aspects of the proposed approach. 👍

pippolo84 (commented):
Made a quick pass to:

- confirm the correctness of the configuration approach
- review the new CIliumResourceNetworkConfig object proposal

Just a very minor nit and a minor proposal about `vlanID` left inline 🚀

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,451 @@
...
+This mode works the same way as the [Multi pool IPAM](https://docs.cilium.io/en/v1.18/network/concepts/ipam/multi-pool/) for pods,
+but it manages IP pools reserved for DRA resources.
+Similarly to CiliumPodIPPool, the CiliumResourceIPPool contains cluster-wide IP pools reserved for DRA resources addresses
+allocations. CIDRs from those pools are allocated by the operator to the agents on a per-need basis: whenever a resource needs an
+IP address from a pool, the agent requests additional CIDRs in the CiliumNode `spec.ipam.resourcepools.requested` field and the operator
+writes the allocated CIDRs in the CiliumNode `spec.ipam.resourcepools.allocated` field.
+
+This mode allows for great flexibility in defining multiple IP pools for different DRA resource types and it allows the reuse of most of
+the implementation for the Multi Pool Pod IPAM. Using separate k8s resources to define IP pools ensures there is no interference between
+the DRA Resources and Pod IPAM.
```
pippolo84: After trying the multi-pool resource approach for IPAM, we realized that reserving CIDRs to each node may not be the best solution for DRA resources. We would like to also prototype other approaches where the CIDRs are shared between subset of nodes (or all the nodes) and the addresses are assigned by the operator directly, to maximize the pool utilization and avoid addresses management in the agent. Also, there might be better ways to represent a CIDR in a pool: using something like `startAddr` and `endAddr` allows the cluster operator to easily describe any available allocation space.

Since IPAM is a pretty complicated topic to be addressed, I think we should remove this section and leave the discussion for a follow-up CfP.

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: so, i would prefer to keep both options since the simplest deployments might not need the flexibility we bring with `CiliumResourceNetworkConfig`. it offers a plain and direct way to pass this value down.

dylandreimerink (approved):

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,437 @@
...
+any given node, the cilium-operator is responsible for choosing
+the right configuration for each node, and once it selects
+a suitable configuration among the candidates, an object of type
+`CiliumNetworkDriverNodeConfig` is created for each node, containing
+the configuration specifics for the node.
+If the operator detecs that more than one configuration suits a node
+based on the node labels, it reports a conflict back to the 
+status field of the offending `CiliumNetworkDriverClusterConfig` object,
+and proceeds to either using the oldest one found - and if there is a tie,
+no configuration is selected and the status is reported for both.
```
YutaroHayakawa: Nit: There's a dedicated section that describes this conflict resolution mechanism below. Maybe we can delete this sentence or link to that section.

cilium/CFP-43295-cilium-network-driver-dra.md
YutaroHayakawa: Seems like this is not resolved yet.

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,437 @@
...
+        enabled: true
+        ifaces:
+          - ifName: enp2s0f0np0
+            vfCount: 6
+          - ifName: enp2s0f1np1
+            vfCount: 6
+```
+
+Note that an operator can also create `CiliumNetworkDriverNodeConfig` matching
+for each of the nodes manually and skip the cilium-operator config selection.
```
YutaroHayakawa: I think we should prohibit this instead. With this, we have multiple writers to the NodeConfig. When user create a NodeConfig manually and later select the node from ClusterConfig, then the NodeConfig will be overridden. In the large cluster, I think it's very easy to make this kind of the mistake.

Users can create the ClusterConfig that only selects a single node (e.g. using `kubernetes.io/hostname` label) instead. So that we can rely on the operator's tie-breaking/conflict-detection.

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,437 @@
...
+      - name: a-side
+        exactly:
+          deviceClassName: a-side.sriov.cilium.k8s.io
+---
+apiVersion: cilium.io/v2alpha1
+kind: CiliumResourceNetworkConfig
+metadata:
+  name: network-config-a
+spec:
+  - nodeSelector:
```
YutaroHayakawa: This is an array of spec. I feel like we should avoid this because:

1. Having an slice of spec is unconventional. Also, it should be plural (specs) if this is a slice.
2. This kills the extensibility of the resource. We cannot introduce `spec.someExtension` field in the future for example.

We can introduce `.spec.configs`, `.spec.nodeConfigs`, etc instead.

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,437 @@
...
+      matchLabels:
+        kubernetes.io/hostname: kind-worker
+    ipv4:
+      ipv4Address: 10.10.0.10/24
+      staticRoutes:
+        - destination: 10.0.0.0/8
+          gateway: 10.10.0.1
+  - nodeSelector:
+      matchLabels:
+        kubernetes.io/hostname: kind-control-plane
```
YutaroHayakawa: This nodeSelector smells to me. It's pretty hard to get it right.

1. User must select the node that the resource actually exists.
2. What happens if the selection overlaps?
3. What happens if there's no match?

If there's a possibility of no match, then shouldn't we prevent Pods from being scheduled to the undesirable node from the first place?

One idea to achieve it is binding this kind of the configuration to the pool itself and remove nodeSelector from the `CiliumResourceNetworkConfig`. The node-local agent can still configure the right static route in this configuration model.

```
kind: CiliumNetworkDriverClusterConfig
<...>
  pools:
    - name: a-side
      filter:
        pfNames:
          - enp2s0f0np0
      config: <= New stanza
        sriov:
          ipv4:
            staticRoutes:
            - destination: 10.0.0.0/8
              gateway: 10.10.0.1
    - name: b-side
      filter:
        pfNames:
          - enp2s0f1np1
```

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,437 @@
...
+apiVersion: cilium.io/v2alpha1
+kind: CiliumResourceNetworkConfig
+metadata:
+  name: network-config-a
+spec:
+  - nodeSelector:
+      matchLabels:
+        kubernetes.io/hostname: kind-worker
+    ipv4:
+      ipv4Address: 10.10.0.10/24
```
YutaroHayakawa: Having this static IP address configuration on the resource referred from ResourceClaimTemplate feels wrong to me. ResourceClaimTemplate is reused among multiple claims meaning multiple resource may get the same IP address in this configuration.

You're planning to turn this to the IPAM reference in the future, so I'm not worrying too much about this specific case. However, we should avoid the same pattern in the future as well.


Nit: This is under ipv4 stanza, so it's obvious the address is IPv4 address. The field name can be address instead.

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,437 @@
...
+  spec:
+    devices:
+      config:
+        - requests:
+            - a-side
+          opaque:
+            driver: sriov.cilium.k8s.io
+            parameters:
+              vlanID: 123
+              ipv4Address: 192.0.2.1/30
```
YutaroHayakawa: Having static IP address configuration in the ResourceClaimTemplate feels weird to me. This means multiple claimed resource may get the same IP address.

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,437 @@
...
+
+```
+---
+apiVersion: cilium.io/v1
+kind: CiliumNetworkDriverClusterConfig
+metadata:
+  name: sriov-nodes-cluster-config
+nodeSelector:
+    matchLabels: 
+      sriov: true
```
YutaroHayakawa: This nodeSelector should be under spec. Maybe leftover from the previous design?

YutaroHayakawa (changes requested):
Came back to this after a while. Seems like there are bunch of updates from when I saw it last time. Made some feedbacks.

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: spoke with @pippolo84 , and it sounds like a good idea. let's get rid of this additional complexity

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: i see where you're coming from, and yes during the implementation it did not feel quite right.
however, the use case we want to address is when different nodes need to use different IDs depending on the underlying network topology.
for example, a pod might want access to BLUE network, but BLUE network has different vlan IDs between the physical node and the ToR switch. having the pod (or resourceclaim) referencing the BLUE name instead of containing the specifics of the underlying infrastructure (ex: vlan ID) hides all those details from the end user (the one ultimately claiming the resources) while allowing for flexibility in the network topology and simplifying the yaml logic.

with the approach you propose, we keep the benefits, but at the cost of a more verbose configuration spec - which will also need to allow for updates without requiring a restart. we're planning to handle config updates in the future, so the only drawback i see is the resulting yaml verbosity (say, you working with 20 or more pools over a number of nodes)

i'll chat with @pippolo84 about this one in particular, and would be happy to know if you can think of is an alternative way we could look into for encapsulating this complexity while allowing a selector based assignment!

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: let's get rid of that

edit: will keep that for now until the ipam bits are all sorted out

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: let's get rid of that

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: hmmm, just thought about the case where the pool contains many devices but they can be assigned with different configuration. think of a vlan trunk for instance - any VF on that uplink could join any of the vlan ids.
so the relationship isn't always a 1:1

bersoare:
@YutaroHayakawa , thank you for the feedback! just pushed a commit addressing most of them.
2 items still open in my view, that require further discussion:
https://github.com/cilium/design-cfps/pull/85#discussion_r3377581819
https://github.com/cilium/design-cfps/pull/85#discussion_r3377201977

cilium/CFP-43295-cilium-network-driver-dra.md
bersoare: now it is!

cilium/CFP-43295-cilium-network-driver-dra.md
```diff
@@ -0,0 +1,429 @@
...
+Alternatively, a configuration with an empty `nodeSelector` matches
+all nodes.
+
+Since only one configuration is used at a time for
+any given node, the cilium-operator is responsible for choosing
+the right configuration for each node, and once it selects
+a suitable configuration among the candidates, an object of type
+`CiliumNetworkDriverNodeConfig` is created for each node, containing
+the configuration specifics for the node.
+If the operator detecs that more than one configuration suits a node
```
tommyp1ckles: typo: detecs -> detects

tommyp1ckles:
> the driver will not change the existing configuration on the PF - the VFs will not be cleaned up. and based on my experiments so far, existing pods can still function as they were. only new workloads will be affected - the scheduler will no longer consider the node for workloads coming in simply because it doesn't advertise any resources.

Bit late to this discussion, I have a few questions regarding node restart Pod lifecycle:

* What would happen to Pods with leftover VFs following a Node restart in the config delete scenario?
* Following a Node restart we would need to wait for the Agent to setup the vfs again, but in that instance I don't think we take into account the scheduler anymore so the resource slice would not be taken into account. So what would happen in such case? Could the VFs be resolved after restarting the Pod?

Also, should we specify the behaviour regarding config changes in the CFP?

bersoare:
@tommyp1ckles , added a section for the config updates! lmk what you think

cilium/CFP-43295-cilium-network-driver-dra.md
pippolo84: Updated the CfP with `.spec.nodeConfigs` :+1:

cilium/CFP-43295-cilium-network-driver-dra.md
pippolo84: @YutaroHayakawa removed the additional CRD for now. It's not strictly needed for the core part of the driver, so we can reevaluate later in a follow up CfP or PR.

pippolo84:
@YutaroHayakawa addressed your feedback, do you mind taking another look? Thanks in advance :pray:

YutaroHayakawa (approved):
Now all of my concerns are addressed. Thanks! Approving.

pippolo84 (approved):

pippolo84:
Given the multiple approvals I think this is implementable.

About the parts that need further discussion, specifically IPAM and resource network configuration CRD, we'll defer that to specific follow up CfPs if needed.

PR #87, opened 2026-03-06
CFP-18405: ENI IPAM Multi-Pool Migration

HadrienPatte:
This CFP is a more detailed version of the design I mentioned in https://github.com/cilium/cilium/issues/19251#issuecomment-3981293722

### DIFF
--- a/cilium/CFP-18405-eni-ipam-multi-pool-migration.md
+++ b/cilium/CFP-18405-eni-ipam-multi-pool-migration.md
@@ -1,0 +1,196 @@
+# CFP-18405: ENI IPAM Multi-Pool Migration
+
+**SIG:** SIG-IPAM
+
+**Begin Design Discussion:** 2026-03-06
+
+**Cilium Release:** 1.20
+
+**Authors:** Hadrien Patte <hadrien.patte@datadoghq.com>
+
+**Status:** Draft
+
+## Summary
+
+Migrate ENI IPAM mode from the CRD allocator to the multi-pool allocator to reduce Kubernetes API churn, simplify prefix delegation handling, and lay the groundwork for future IPv6 support.
+
+## Motivation
+
+ENI IPAM mode currently uses the CRD allocator: the operator writes individual IPs to `CiliumNode.Spec.IPAM.Pool` and the agent allocates from there, writing per-IP status back to `Status.IPAM.Used`. This design has a couple limitations.
+
+**IPv6 is impossible.** The CRD allocator works by flattening prefixes into individual IP entries. For IPv4 prefix delegation, AWS assigns /28 prefixes (16 IPs), which is manageable to flatten. For IPv6, AWS assigns /80 prefixes (2^48 IPs). Flattening these into individual `Spec.IPAM.Pool` entries is not technically possible. Without migrating to a CIDR-aware allocator, ENI IPv6 delegated prefix support cannot be implemented.
+
+**Excessive Kubernetes API churn.** Every IP allocation and deallocation requires a CiliumNode status update. Since all agents watch all CiliumNodes, each update fans out to every node in the cluster. In a cluster with N nodes, a single allocation event generates N watch notifications. Under sustained churn (rolling deployments, autoscaling, job workloads), this creates significant and unnecessary load on the Kubernetes API server and etcd.
+
+## Goals
+
+* Migrate ENI IPAM mode (both secondary IP and prefix delegation) to the multi-pool allocator
+* Decouple ENI device configuration from the allocator code path
+* Reduce Kubernetes API churn for IP allocation and deallocation
+* Preserve backward compatibility for existing ENI IPAM users during the transition
+
+## Non-Goals
+
+* IPv6 support in ENI IPAM mode (deferred milestone)
+* IPv6 prefix delegation support in ENI IPAM mode (deferred milestone)
+* Changes to other cloud IPAM modes (Azure, AlibabaCloud)
+
+## Proposal
+
+### Overview
+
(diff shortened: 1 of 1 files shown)


cilium/CFP-18405-eni-ipam-multi-pool-migration.md
```diff
@@ -0,0 +1,196 @@
...
+All CIDRs regardless of subnet are written under a single pool name.
+
+##### Pros
+* Simpler operator logic; no pool name coordination needed.
+* Straightforward for the initial migration.
+
+##### Cons
+* Cannot express per-subnet constraints; future per-subnet pod placement would require a redesign.
+
+#### Option 2: Per-Subnet Pools
```
gandro: I feel like we should punt this to a future milestone and really go with option 1 for now.

In particular, I could also see use-cases where one wants to have multiple subnets in a pool (e.g. based on a subnet-tag-filter or similar), and thus having only one subnet per pool might be too limiting. But this is probably something we want to design separately.

cilium/CFP-18405-eni-ipam-multi-pool-migration.md
```diff
@@ -0,0 +1,196 @@
...
+
+##### Cons
+* Adds a short-lived flag that must be documented and then removed, with a deprecation notice in 1.21.
+* Risk of users enabling it prematurely (before all agents have fully rolled to 1.20), leaving some nodes without a populated `Spec.IPAM.Pool`.
+
+### Key Question 2: Pool Naming Strategy
+
+Should all ENI-allocated addresses share a single `"eni"` pool, or should there be more purpose-named pools?
+
+#### Option 1: Single `"eni"` Pool
```
gandro: nit: I do wonder if we should simply call this `"default"`. The "default" pool in Multi-Pool IPAM mode is the one that is used if the user hasn't requested anything special and I feel like we should probably mirror that in ENI mode.

cilium/CFP-18405-eni-ipam-multi-pool-migration.md
```diff
@@ -0,0 +1,196 @@
...
+
+##### Pros
+* Simple logic: no per-node format detection needed in the operator.
+* Fully safe for both upgrade and downgrade at any point.
+
+##### Cons
+* `Spec.IPAM.Pool` remains populated on all nodes even when all agents have upgraded, until 1.21.
+* Slightly more data written per CiliumNode object.
+
+#### Option 2: Temporary Opt-Out Flag to Disable Dual-Write
```
gandro, line 145: I'm mildly in favor of providing this. 

We could even go a step further and and enable this by default for new installations. While it is a potentially dangerous flag to enable, I do feel like with proper documentation this risk is mitigated.

gandro (approved):
I am so happy to see this being tackled! 🥹 

The plan here sounds very reasonable and implementable to me. I have left some comments on the key questions, but the major concerns of such a change are already addressed in the document from my perspective, so I am approving of this.

cilium/CFP-18405-eni-ipam-multi-pool-migration.md
HadrienPatte: Updated proposed pool name to `default`, and removed the `Pool Naming Strategy` key question and moved the ideas from its `Option 2: Per-Subnet Pools` to the `Per-Subnet and Multi-Subnet Pod Placement` future milestone section

cilium/CFP-18405-eni-ipam-multi-pool-migration.md
HadrienPatte: https://github.com/cilium/design-cfps/pull/87#discussion_r2906092913

cilium/CFP-18405-eni-ipam-multi-pool-migration.md
HadrienPatte: Do we have an established way of doing "if new installation" type of logic today? I feel like a lot of the great options on [this page](https://docs.cilium.io/en/stable/operations/performance/tuning/) currently default to being disabled (even on new installations) because they're not safe to be enabled on existing clusters.

HadrienPatte:
> The plan here sounds very reasonable and implementable to me. I have left some comments on the key questions, but the major concerns of such a change are already addressed in the document from my perspective, so I am approving of this.

Thanks, I have updated the CFP based on your comments and changed its status from `draft` to `implementable`. Will leave it open for more people to comment.

cilium/CFP-18405-eni-ipam-multi-pool-migration.md
gandro: Yes, we do, the `upgradeCompatibility` flag. Though it is possible that many users are not using it when upgrading, which would of course be pretty bad for them. If we want to be safe, maybe your original proposal to make this strictly opt-in is the way to go.

https://docs.cilium.io/en/stable/operations/upgrade/#step-2-use-helm-to-upgrade-your-cilium-deployment

joestringer:
Shall we merge this as-is? Then if the ongoing implementation necessitates some update, @HadrienPatte can come back and update it?

HadrienPatte:
Yes, let's merge. For people following along, here's the current status on the implementation of this CFP:

Merged:
* Preparation and refactoring:
  * https://github.com/cilium/cilium/pull/45025
  * https://github.com/cilium/cilium/pull/45027
  * https://github.com/cilium/cilium/pull/45082
  * https://github.com/cilium/cilium/pull/45089
* Operator dual support for both CRD and multipool agents:
  * https://github.com/cilium/cilium/pull/45110
  * https://github.com/cilium/cilium/pull/45124

In review:
* Switch ENI agents from CRD to multipool allocator: https://github.com/cilium/cilium/pull/45154

TODO:
* Update the IP release mechanism in the operator to read the release signals from the multipool agent
* Update documentation to reflect the updated ENI IPAM implementation: https://docs.cilium.io/en/stable/network/concepts/ipam/eni/

PR #88, opened 2026-03-13
CFP-44774: Cilium Control Plane at Scale

vipul-21:
This CFP proposes reusing the existing ClusterMesh etcd as a non-persistent, cache-backed data distribution layer so agents read Cilium CRDs from etcd instead of the API server, reducing API server load at scale.

Issue: https://github.com/cilium/cilium/issues/44774


cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+cross-replica consistency or replication between these etcd instances. Each
+replica independently watches the API server and populates its own etcd. Agents connect to one replica via sticky connections and stay connected to that replica
+for the duration of the connection. If a replica goes down, the agent
+reconnects to another replica and performs a full re-sync from scratch.
+
+**Bootstrap and connectivity:** Deploying the ClusterMesh API Server inside
+the same cluster whose agents depend on it creates a bootstrap problem:
+agents need ClusterMesh etcd for control plane state, but the ClusterMesh
+API Server itself may need functioning agents for pod networking or Service
+IP routing. To address this, we propose two options:
```
MrFreezeex, line 99: Unless there is a clear advantage on why having different modes I would suggest only supporting one. There are already many different options and combination within clustermesh/kvstore/kvstoremesh and adding more options and ways sounds difficult to maintain and keep in mind in the long run. So we should have a very clear reason to support multiple option here if we want to go forward with that IMO.

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+  CFP, but can be an evolution of this work.)
+* **Centralized control plane operations**: Move compute-intensive control
+  plane activities such as network policy calculation and ipcache computation
+  into the ClusterMesh API Server, reducing per-agent CPU overhead and
+  enabling consistent, cluster-wide policy evaluation. (Currently a non-goal
+  for this CFP, but can be a natural evolution of the architecture.)
+* **Multi-cluster integration**: Ensure the single-cluster KVStore mode
+  composes cleanly with existing multi-cluster ClusterMesh deployments,
+  allowing clusters that use this optimization to also participate in
+  cross-cluster service discovery and identity sharing.
```
MrFreezeex, line 315: I would tend to think that this should be fairly straightforward, any reasons to have this in future milestones?

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+  where every CRD mutation traverses both the Kubernetes API server and
+  ClusterMesh etcd. Agents and the operator would write Cilium state directly
+  to the ClusterMesh etcd, removing the extra API server round-trip and
+  reducing end-to-end propagation latency. (Currently a non-goal for this
+  CFP, but can be an evolution of this work.)
+* **Centralized control plane operations**: Move compute-intensive control
+  plane activities such as network policy calculation and ipcache computation
+  into the ClusterMesh API Server, reducing per-agent CPU overhead and
+  enabling consistent, cluster-wide policy evaluation. (Currently a non-goal
+  for this CFP, but can be a natural evolution of the architecture.)
```
MrFreezeex, line 311: Is that desirable? Clustermesh-apiserver is treated as a cache here so we would need in each clustermesh-apiserver to compute the same thing independently (and that those are consistent enough) or to also use the kube-apiserver to perform that which would be similar to the operator 🤔

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+  etcd restarts in large clusters.
+
+## Future Milestones
+
+* **Direct writes to ClusterMesh etcd**: Eliminate the duplication overhead
+  where every CRD mutation traverses both the Kubernetes API server and
+  ClusterMesh etcd. Agents and the operator would write Cilium state directly
+  to the ClusterMesh etcd, removing the extra API server round-trip and
+  reducing end-to-end propagation latency. (Currently a non-goal for this
+  CFP, but can be an evolution of this work.)
```
MrFreezeex, line 306: We were discussing in slack with Marco, Tamilmani and I that this was just looking like the kvstore mode and that due to the ephemeral nature of ClusterMesh etcd would not really work unless you make it persistent which would make it almost exactly like the kvstore mode. So I don't think it's a desirable milestone and I would suggest to remove or possibly change it to, as Marco was saying on slack, improving the existing kvstore mode.

MrFreezeex (changes requested):
Hi thanks for the proposal! I skimmed through this CFP and left a few comments. This is more a quick pass than a full review.

Overall I think this proposal is mostly focused on the why and sounds like it's missing some key details on how you would implement this. Like would you use the `pkg/clustermesh` code and try to treat the local cluster like almost any other remote cluster? Something else? ~~There's also the Service aspects which you didn't say explicitly that you would use the data from Kubernetes or clustermesh-apiserver in this mode.~~ EDIT: nevermind this is discussed

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+| **K8s API Server** | 8 vCPU / 32 GB |
+| **Nodes** | 1,000 worker nodes per cluster (4 vCPU / 16 GB) |
+| **Control Plane** | 1 control-plane node (8 vCPU / 32 GB) |
+| **Kubernetes** | v1.31 (kubeadm) |
+| **Cilium Chart** | v1.19.0-dev |
+| **Load Generator** | ClusterLoader2 |
+| **Namespaces** | 1,000 |
+| **Pods** | 40,000 |
+| **Deployments** | 4,000 |
+| **Pod Deployment Rate** | 100 pods/sec |
```
MrFreezeex, line 227: Does that deploy some Services as well? If not it sounds interesting to have as part of this benchmark to better match the reality and reflect what part of the load we are offloading to the clustermesh-apiserver

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+supports horizontal scaling, multiple replicas can distribute the fan-out
+load across agents.
+
+**Scope of offloading:** This proposal offloads agent *reads* of Cilium CRDs
+(CiliumIdentity, CiliumEndpoint, CiliumEndpointSlice and CiliumNode) from
+the API server.
+Agents and the operator continue to *write* CRDs to the Kubernetes API server
+as they do today. Additionally, Kubernetes-native resources such as Services,
+Endpoints, Pods, and Nodes are not covered by this proposal, agents continue
+to watch these directly from the API server.
```
MrFreezeex, line 81: Could you add some details on the reasoning behind this part?

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+
+Request latency performs better in ClusterMesh KVStore mode, with a significant reduction in latency(data is for 99th percentile latency).
+
+#### ClusterMesh API Server CPU and Memory Usage
+
+![ClusterMesh API Server CPU and memory usage over time](./images/CFP-44774-cilium-clustermesh-apiserver-cpu-memory.png)
+
+![Embedded etcd CPU and memory usage over time in ClusterMesh mode](./images/CFP-44774-cilium-clustermesh-etcd-cpu-memory.png)
+
+Clustermesh API server containers(i.e apiserver and etcd) CPU and memory usage remained stable during the test, with no significant spikes during workload churn.
```
MrFreezeex, line 266: It might be nice to have some quantile number or graph of memory/CPU usage. For instance, it is said here that the clustermesh-apiserver etcd was stable in term of CPU, does that mean it's always running 3 CPU (the max on the chart provided)? A nice way (if actually possible) to show it could be a graph of CPU of the two mode for both the k8s apiserver and clustermesh-apiserver on the same graph and likewise for memory.

vipul-21:
@MrFreezeex Thanks for the review. Yes, the implementation reuses `pkg/clustermesh`. The local cluster data is synced to the clustermesh API server’s etcd (similar to how remote cluster data is handled). Each agent then watches the same etcd to retrieve the required information.

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
+# CFP-44774: Cilium Control Plane at Scale
+
+**SIG: SIG Scalability, SIG-ClusterMesh**
```
giorio94, line 3: It makes sense IMO to involve the scalability and clustermesh SIGs for the initial design discussion, but I don't think it is sustainable to extend the scope of these SIGs to also oversee the new logic possibly resulting from this CFP. Both because outside of their main scope (it does not pertain multi-cluster, and SIG scalability should oversee scalability-related topics, but not end up owning any improvement potentially targeting improved scalability), and as both teams are seriously understaffed at the moment.

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+every node), each agent watches the same set of CRDs, so every update must be
+independently serialized and transmitted to every watcher. This drives up API
+server CPU consumption and can result in throttling or control plane
+instability, especially in environments where scaling API server resources is
+constrained or not feasible.
+
+An external KVStore can offload identity watches but still requires a
+separately provisioned and managed etcd cluster, and it only partially reduces
+API server pressure because endpoint slices and node objects continue to be
+served through CRD watches. A solution that offloads all agent CRD watches
```
giorio94, line 42: I don't follow the second part of the sentence (assuming you refer to CiliumEndpointSlice and CiliumNodes). When Cilium runs in KVStore mode, identity, endpoint and node entries are all watched from etcd, rather than the Kubernetes API Server. There are certain limitations and feature incompatibilities given that the overall trend in recent years has been to focus more on CRD mode, but these are orthogonal aspects that can be resolved separately.

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+while reusing infrastructure that already ships with Cilium would provide a
+broader reduction in API server load with minimal operational overhead.
+
+## Goals
+
+* Reduce Kubernetes API server CPU utilization by reducing the watch
+  connections and event throughput, improving cluster scalability with
+  limited API server resources.
+* Utilize ClusterMesh etcd as an alternative datastore, avoiding the need to
+  provision and maintain a separate external KVStore.
```
giorio94, line 52: To me, a key pre-requisite for adding the support for a new datastore is to have a unified abstraction (likely based on statedb) that allows to decouple the downstream consumers from the actual ingestion logic, which I don't see discussed in this CFP.

Currently, the code paths for ingesting data in CRD and KVStore mode (as well as from clustermesh) are diverse and fairly independent, mostly for historical reasons. However, that proved being a constant pain point, leading to feature incompatibilities, caveats only applying to one method or the other (e.g., around connection disruption on agent restart) and overall complexity in general. 

I don't think it is viable to introduce yet another datastore without carefully rethinking the supporting architecture, as it would lead to even more feature incompatibilities and complexity.

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+
+* Implementing dedicated centralized control plane activities (network policy
+  calculations, ipcache) in the ClusterMesh API Server.
+
+## Proposal
+
+### Overview
+
+To reduce Kubernetes API server load, we propose utilizing the existing
+ClusterMesh etcd as a non-persistent, cache-backed datastore for agent
```
giorio94, line 66: From a deployment perspective, I think it would be appropriate to keep the proposed datastore separate from the existing clustermesh-apiserver for better isolation, both in terms of permissions, and to prevent possible interference in case of e.g., certain replicas end up being overloaded due to either in-cluster or cross-cluster data synchronization. Potential failures do also have very different consequences and implications in the two cases. The underlying binary may be reused if appropriate though.

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+
+1. **Pod networking mode (recommended)**: The ClusterMesh API Server runs as
+   a regular pod (no host networking) and is exposed via a Kubernetes
+   Service. Agents attempt to connect to ClusterMesh etcd at startup; if the
+   connection fails or is not yet available, agents fall back to reading
+   CRDs directly from the Kubernetes API server. Once the ClusterMesh API
+   Server becomes reachable, agents switch over to reading from ClusterMesh
+   etcd. This mode requires no special scheduling or networking
+   configuration and is operationally identical to a standard Cilium
+   deployment.
```
giorio94, line 109: As mentioned in Slack, Cilium historically included a handover mechanism similar to the one proposed here, to support KVStore mode with etcd running in pod network. This turned out being a non-ending source of bugs and inconsistencies, because the two sources are not perfectly aligned at any point in time. Additionally, it also leads to unnecessary overhead on the Kubernetes API Server, if the informers are started only to be immediately terminated afterwards once successfully connected to etcd. That support got ripped out a few versions ago, and IMO we really need to stick to a single source of truth, either K8s or etcd, and not do hand-overs in case of failures or alike.

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+
+2. **Host networking mode**: The ClusterMesh API Server runs with
+   `hostNetwork: true`, bypassing the need for CNI-provided pod networking.
```
giorio94, line 112: IMO, this would be one more reason for separating the proposed deployment from the existing clustermesh-apiserver. One more limitation of using host network pods is that there can be only a single replica per node, as otherwise ports would conflict.

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+   Agents discover the ClusterMesh API Server by watching its Pod resource
+   from the Kubernetes API server to obtain the host IPs, then connect
+   directly to those IPs. This avoids depending on Service IP datapath
+   programming (which Cilium would need to provide) but
+   introduces host port conflicts and requires agents to maintain a
+   lightweight watch on the ClusterMesh API Server pods for IP discovery
+   and failover when pods reschedule to different nodes.
```
giorio94, line 119: We already bypass datapath-level service load balancing when connecting to KVStoreMesh, performing service DNS name resolution and backend selection via a custom dialer, to break possible chicken-and-egg dependencies. The same logic may be potentially adopted here, without having to watch for the pods themselves.

cilium/CFP-44774-cilium-control-plane-at-scale.md
giorio94, line 227: Which settings are you using for the clustermesh-apiserver? I assume not the default ones, as the etcd client is configured with 20 QPS, which could not sustain a pod deployment rate of 100/sec.

cilium/CFP-44774-cilium-control-plane-at-scale.md
```diff
@@ -0,0 +1,315 @@
...
+### Metrics Measured
+
+| Metric | Prometheus Metric |
+| --- | --- |
+| **API Server CPU** | `process_cpu_seconds_total` |
+| **Watch Event Throughput** | `apiserver_watch_events_total` |
+| **API Server Request/Response Latency** | `apiserver_request_duration_seconds_bucket` |
+| **ClusterMesh API Server CPU** | `container_cpu_usage_seconds_total` |
+| **ClusterMesh API Server Memory** | `container_memory_usage_bytes` |
+| **Pod Startup Latency** | `kubelet_pod_start_sli_duration_seconds_bucket` |
```
giorio94, line 212: How did you measure pod readiness? Time before the pod is marked as ready, or time before another pod is able to successfully connect to the new pod, subject to network policies (i.e., all information propagated to the other agents)? The former may be misleading, if the hosting agent does not require any information to be able to start the pod.

cilium/CFP-44774-cilium-control-plane-at-scale.md
giorio94: Agree with Arthur that this would be an improvement of the existing KVStore mode, and not compatible with the approach of using ephemeral etcd instances.

cilium/CFP-44774-cilium-control-plane-at-scale.md
giorio94: Agree with Arthur that this looks out of scope here, and likely problematic with the cache-based approach.

giorio94 (changes requested):
Thanks for the proposal!

A few initial high-level comments from my side as well.

cilium/CFP-44774-cilium-control-plane-at-scale.md
vipul-21: Agreed. This is intentionally listed as a non-goal for this CFP. We included it to acknowledge the architectural possibility, but as you both pointed out, it conflicts with the cache-based model where each replica operates independently. We'll remove this from the Future Milestones to avoid implying it's a planned direction.

cilium/CFP-44774-cilium-control-plane-at-scale.md
vipul-21: Agreed. The host networking approach adds some complexity without providing much benefit to the proposal. Will keep the recommended option and remove the other.

cilium/CFP-44774-cilium-control-plane-at-scale.md
vipul-21: You're right that it should be straightforward since we're reusing the same Cluster mesh infrastructure. We listed it as a future milestone because the POC was scoped to single-cluster only and we haven't validated the multi-cluster combination yet. Once the single-cluster mode lands, verifying multi-cluster composition should be a natural follow-up with minimal additional work.

cilium/CFP-44774-cilium-control-plane-at-scale.md
vipul-21: We did not deploy any Services yet. But yes that should be another set of benchmark we can run.

cilium/CFP-44774-cilium-control-plane-at-scale.md
vipul-21: We wanted to reduce the cluster wide load on the API server originating from cilium as the first step. This comes primarily from CiliumIdentity, CiliumEndpointSlice, and CiliumNode as every agent watches the full set of these. Similarly, K8s Services and EndpointSlices are watched cluster-wide, but offloading those is a larger change that should be evaluated separately.

cilium/CFP-44774-cilium-control-plane-at-scale.md
vipul-21: We took the max of the cpu usage during the load as mentioned. But I understand your ask, will try to get the quantile number by scraping the prometheus as ClusterLoader2 doesn't provide CPU/memory quantiles directly(afaik)

joestringer:
I've marked this PR as draft. When you have addressed the comments, bring the PR back to the attention of reviewers by scrolling to the bottom of the page and clicking the "Ready for Review" button. Thanks!

PR #90, opened 2026-04-01
CFP-45118: Add shared policy lpm trie proposal.

tsotne95:
Shared policy lpm trie

### DIFF
--- a/cilium/CFP-45118-shared-policy-lpm-trie.md
+++ b/cilium/CFP-45118-shared-policy-lpm-trie.md
@@ -1,0 +1,153 @@
+# CFP-45118: Shared Policy LPM
+
+**SIG: SIG-Policy, SIG-Datapath, SIG-Scalability** 
+
+**Begin Design Discussion:** 2026-04-01
+
+**Cilium Release:** 1.20+
+
+**Authors:** Tsotne Chakhvadze <tsotne@google.com>
+
+**Status:** Proposal
+
+# Sharing Policy Maps to Save Memory (Without BPF Arena)
+
+## Summary
+
+This proposal introduces a **Shared LPM Trie architecture** to deduplicate network policy rules across endpoints on a single node. 
+
+Currently, Cilium creates a per-endpoint policy map. If multiple pods run the same application profile (e.g. 1000 replicas of a microservice), Cilium creates replicate rules for each endpoint. This architecture proposes a single, node-wide Shared LPM Trie map where endpoints share a single set of rules, reducing control-plane memory usage significantly.
+
+This is a practical, immediate alternative to **BPF Arena**, as it works on older kernels and avoids complex pointer management.
+
+## Motivation
+
+As Kubernetes clusters scale to thousands of nodes and tens of thousands of pods, Cilium's per-endpoint policy architecture faces critical scaling bottlenecks:
+
+1.  **Per-Endpoint Map Exhaustion**: Each endpoint has a fixed-size policy map (typically 16,384 entries). Complex environments with fine-grained segmentations can exceed this limit, causing policy drops or failures to attach.
+2.  **Node Memory Pressure**: Replicating the same rule set across thousands of pods consumes gigabytes of memory per node. This reduces the memory available for user applications and increases infrastructure costs.
+3.  **Bpf Arena Blockers**: While BPF Arena can solve this, it requires very new kernels (Linux 6.9+) and introduces native pointer complexities that slow down adoption.
+
+Most memory savings (estimated ~99%) come from **Rule Set Sharing** (Level 1 Deduplication). We can achieve this blocking requirement immediately using standard, broadly supported BPF primitives (`BPF_MAP_TYPE_LPM_TRIE`) that work on older kernels. This allows us to scale to massive clusters without hitting map limits or wasting node RAM.
+
+## Goals
+
+*   **Memory Efficiency**: Reduce node memory footprint by sharing rule sets across identical endpoints.
+*   **Kernel Compatibility**: Support older kernels without requiring Arena.
+*   **Safety**: Change rules without dropping packets or accidentally mixing up different apps
+
+## Non-Goals
+
+*   Per-rule fine-grained pointer sharing (Level 2 Deduplication), which remains the domain of future BPF Arena enhancements.
+
+## Proposal
+
+### Overview
+
+The design replaces the private per-endpoint policy map with a two-stage lookup mechanism using two BPF maps:
+
+1.  **Overlay Map** (Hash): Maps an `Endpoint ID` to a `Rule Set ID`.
+2.  **Shared LPM Trie Map** (Global): Maps `Rule Set ID` + packet details to a verdict.
+
+```mermaid
+graph LR
+    subgraph Traffic ["Packet Flow"]
+        direction TB
+        EP["Endpoint ID: 42"]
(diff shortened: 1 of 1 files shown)


joestringer:
Initial comment, could you please follow https://github.com/cilium/design-cfps?tab=readme-ov-file#how-to-create-cfps? In particular could you file an issue at https://github.com/cilium/cilium to get a CFP number and embed that here, then also please follow the [standard CFP template](https://github.com/cilium/design-cfps/blob/main/cfps/CFP-003-template.md). The CFP template is intended to structure the discussion to ensure "why" is covered, as well as identifying impact of the proposal and any other key questions that you would like to draw attention to.

cilium/CFP-TODO-shared-policy-lpm-trie.md
```diff
@@ -0,0 +1,167 @@
...
+---
+
+## 1. Why do it this way?
+
+We previously talked about using "BPF Arena" to share memory. While Arena is great, it has some downsides:
+*   It requires very new Linux kernels (version 6.9 or newer).
+*   It is complex to write and hard to debug if something goes wrong.
+
+**The big secret**: Most of the memory savings (99%) come from letting pods share the *same set of rules* (e.g. "All web servers use Policy Set 1"). We can do this with standard BPF features without needing Arena.
+
```
jrajahalme: Currently each endpoint has a separate bpf policy map, tied to the endpoint's bpf program, even if they happen to have the same security identity. This used to be necessary due to endpoint specific config options. If the intent is to only share the policy map between multiple endpoints, then this should be accomplishable without *any* datapath changes by sharing the same bpf map between the endpoints.

Policy map sharing could be based on the endpoints already sharing the same security identity. Since it may still be possible for endpoint specific config to have an effect here, this should be an opt-in via a command line/helm option. This should also be rather straight-forward to implement.

It is also possible that endpoints with different security identities generate the same policy map, possibly from same or different rules. Maybe detecting that the selector policy is the same, even if generated for a different security identity could work as the driver for deduplication. Since each endpoint already points to a specific bpf policy map, also this option should be possible without any bpf datapath changes by simply reusing the same bpf map for all these endpoints.

One unknown if above cases is the effect of named ports, that can be endpoint specific.

Where bpf datapath work would be useful would be to deduplicate and share LPM tries *without* the remote identity in the LPM key. This would work by first looking up the *remote identity* from a separate uniques or shared bpf hash map, resulting into an *LPM identifier* that would then be used as part of the LPM trie key *instead* of the remote identity, like today. The first hash map could be unique or shared (along the lines of the 3 paragraphs above), while the LPM map holding all LPM tries would be shared among all the endpoints. Here the deduplication challenge is harder and must integrate with the algorithms in pkg/policy/mapstate.go. If deduplication is based on computed MapStateEntries, then named ports are not an issue either, as at the MapState level those have already been resolved, but this would mean that the Policy map computation would be performed for each endpoint separately, and then deduplicated after the fact. Agent code churn is much higher in this case.

tsotne95:
> Initial comment, could you please follow https://github.com/cilium/design-cfps?tab=readme-ov-file#how-to-create-cfps? In particular could you file an issue at https://github.com/cilium/cilium to get a CFP number and embed that here, then also please follow the [standard CFP template](https://github.com/cilium/design-cfps/blob/main/cfps/CFP-003-template.md). The CFP template is intended to structure the discussion to ensure "why" is covered, as well as identifying impact of the proposal and any other key questions that you would like to draw attention to.

thank you, made a bug in cilium and updated cfp based on template.

cilium/CFP-TODO-shared-policy-lpm-trie.md
tsotne95: Thanks for the great and detailed feedback, I truly appreciate it. Yes you are right and I see how `mapstate.go` works and why named ports could be a trap. Here is why I think the Shared LPM + Overlay is better, and how we can do it without a massive rewrite of the Go code (I could be wrong though 😅)

If we just reuse the old maps, it is easy for the kernel but hard for Go. I propose to use exactly 2 maps for the whole node. We create them once at startup and never delete them. Sharing old maps still means creating new maps for different apps(fronted, backend, ect.). In big clusters, we can still hit Linux file limits, I guess (but I agree limit is probably high enough). And it will be easier to update. If Pod A and Pod B share a map, and then you add a unique rule to Pod A, Go has to make a new map and swap it live. Instead, we can just change one number in the Overlay map (Pod A points to Set 7 instead of Set 5). No swapping or reloading needed.

And I think we don't need to touch the "math" inside `mapstate.go`. We  can just change what happens after it is done.:
 - Go finishes calculating the rules for Pod A.
 - Instead of pushing to a private map, we send the rules to a new manager.
 - This manager checks if another pod is already using these exact rules.
     - If Yes: It gets the ID for those rules.
     - If No: It saves the rules to the Shared LPM map and gets a new ID.
- We update the Overlay map so Pod A points to that ID.

And we avoid the named port trap because we do our sharing after Go converts the names to numbers. I'll take another look tomorrow, if I missed something. What do you think? 

Thank you again.

cilium/CFP-45118-shared-policy-lpm-trie.md
```diff
@@ -0,0 +1,153 @@
...
+
+    EP --> EP_Key
+    EP_Val --> LPM_Key
+    Pkt --> LPM_Key
+    LPM_Key --> LPM_Val
+```
+
+### 1. BPF Maps Structure
+
+#### The Overlay Map (Endpoint → Rule Set ID)
```
squeed, line 80: You might be able to eliminate this map. Instead, you could provide file descriptors to the endpoint BPF programs.

It's not a big deal to rebuild an endpoint if it transitions from shared to shared+overlay maps.

tsotne95:
I've been thinking more about the different ways to handle this, and I wanted to share a bit more context on why I believe the static Shared LPM approach is the safest path forward, especially for modern workloads.

1. Why fixed maps are safer than sharing old maps
If we try to share the legacy per-pod maps, we still have to create and delete BPF maps dynamically whenever a pod's rules change. Managing map and swapping them live in Go can be tricky and lead to race conditions.

By moving to two fixed maps (`Overlay` + `Shared LPM`) created once when the node starts:
- We never create or delete maps on the fly. We only add or remove rule entries.
- Moving a pod to a new rule set just means changing one number in the `Overlay` map.

2. The Agentic AI Use Case (Why rulesets are getting massive)
A great real-world example of why we need this right now is Agentic AI workloads (like coding agents, automated scrapers, or research bots, ect.).

- AI worker pods constantly need to talk to hundreds of different external web APIs, cloud services, or custom IP ranges to do their jobs.
- To keep them secure, admins write external egress policies using large IP lists (CIDRs) or DNS rules. Cilium expands these into thousands of raw kernel rules per pod.
- If a customer runs 50 or 100 AI worker replicas on a node, duplicating 10,000+ external access rules into every single pod's private map quickly wastes gigabytes of RAM and hits the hard 16K map limit, causing new pods to fail on startup.

joestringer:
Brief update, I see https://github.com/cilium/cilium/pull/47651 is now proposed. Does this CFP need revision?

tsotne95:
I somehow missed the notif. My pr is the direct implementation of this CFP as described.

joestringer:
@squeed do you think this CFP is implementable in its current state? Should we merge the CFP?

PR #91, opened 2026-04-07
CFP-45105 split cilium agent daemonset

mrkiani98:
# Summary

For better resource optimization in case of different usage pattern in different node pools based on nodes' sizes and workloads, we want to introduce the ability to split cilium agent daemonset into multiple daemonsets per different node pools.

The goal here is to make the chart flexible to let us deploy one operator alongside with multiple agent daemonsets per nodepool.

cilium repo CFP issue [link](https://github.com/cilium/cilium/issues/45105).

### DIFF
--- a/cilium/CFP-45105-split-cilium-agent-daemonset.md
+++ b/cilium/CFP-45105-split-cilium-agent-daemonset.md
@@ -1,0 +1,68 @@
+# CFP-003: Template
+
+**SIG: SIG-NAME** ([View all current SIGs](https://docs.cilium.io/en/stable/community/community/#all-sigs))
+
+**Begin Design Discussion:** 2026-04-07
+
+**Cilium Release:** 1.19.2
+
+**Authors:** Mohammadreza Kiani (mohammadrezakiani.shojaee@gmail.com)
+
+**Status:** Implementable
+
+## Summary
+
+Ability to split cilium agent daemonset into multiple daemonsets per different node pools.
+* NOTE: We can work on implementing this feature on the helm chart by opening respective PRs, but first, we want to know the maintainers' and also the community's opinion about the possible different options which we will explain here.
+
+## Motivation
+
+For better resource optimization in case of different usage pattern in different node pools based on nodes' sizes and workloads.
+
+## Goals
+
+The goal here is to make the chart flexible to let us deploy one operator alongside with multiple agent daemonsets per nodepool.
+
+## Proposal
+
+### Overview
+
+While using the community cilium helm chart to deploy the operator and agent, we faced some issues as we cannot split the agent daemonset into multiple daemonsets via separate releases in the same cluster because of some naming conflicts of some RBAC resources defined for the agent.
+
+The list of the problematic resources is as following:
+
+- ClusterRole/cilium
+- ClusterRoleBinding/cilium
+- Role/cilium-config-agent
+- RoleBinding/cilium-config-agent
+
+There are two options as following which both of them are possible to implement and there is no clear trade off to decide which one is better and that's why we will put both options here to first decide which one is a better go.
+
+### Option 1:
+
+Add the ability to enable/disable all related RBAC resources, so that we can have one release which will contain operator and all RBAC resources which will be shared via other multiple agent daemonset releases (in agent releases, the RBAC resources would be disabled).
+
+#### Pros
+
+- No duplicate RBAC resources.
+- Easier management of the resources.
+- Easier implementation and maintenance in the chart.
+
+#### Cons
+
+- No separation and isolation for the separated agent daemonsets RBAC resources.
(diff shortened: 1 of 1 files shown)


xmulligan:
cc: @cilium/sig-agent

cilium/CFP-45105-split-cilium-agent-daemonset.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Goals
+
+The goal here is to make the chart flexible to let us deploy one operator alongside with multiple agent daemonsets per nodepool.
+
+## Proposal
+
+### Overview
+
+While using the community cilium helm chart to deploy the operator and agent, we faced some issues as we cannot split the agent daemonset into multiple daemonsets via separate releases in the same cluster because of some naming conflicts of some RBAC resources defined for the agent.
```
squeed, line 30: I'm not entirely sure why this is desirable, but would separate namespaces be the solution?

mrkiani98: Separate namespaces still cannot be the solution as the name of the cluster role and cluster role binding is hardcoded and the same issue will happen again.

cilium/CFP-45105-split-cilium-agent-daemonset.md
```diff
@@ -0,0 +1,68 @@
...
+**Status:** Implementable
+
+## Summary
+
+Ability to split cilium agent daemonset into multiple daemonsets per different node pools.
+* NOTE: We can work on implementing this feature on the helm chart by opening respective PRs, but first, we want to know the maintainers' and also the community's opinion about the possible different options which we will explain here.
+
+## Motivation
+
+For better resource optimization in case of different usage pattern in different node pools based on nodes' sizes and workloads.
```
HadrienPatte, line 20: If the target objective here is essentially to be able to have different resource requests/limits based on node size, this is a problem we've faced as well and we have solved it with a different solution from the two presented in this CFP (we use a patched version of the cilium chart). Sharing our implementation in case it can be of interest.

Our solution is essentially a single cilium chart with multiple dynamic agent daemonsets.

In our values we define multiple different resources for daemonset "profiles":
```yaml
 agent:
    profiles:
      - name: default
        resources:
          limits:   { cpu: 90m, memory: 2Gi }
          requests: { cpu: 90m, memory: 2Gi }
      - name: small
        resources:
          limits:   { cpu: 90m, memory: 512Mi }
          requests: { cpu: 90m, memory: 512Mi }
      - name: medium
        resources:
          limits:   { cpu: 90m, memory: 1.5Gi }
          requests: { cpu: 90m, memory: 1.5Gi }
      - name: large
        resources:
          limits:   { cpu: 90m, memory: 2Gi }
          requests: { cpu: 90m, memory: 2Gi }
```

Then in the daemonset template, we loop over those profiles with `range`:
```yaml
{{ range $profile := .Values.agent.profiles }}
---
apiVersion: apps/v1
kind: DaemonSet
metadata:
  # "default" profile keeps the original name for backward compatibility
  name: cilium-agent{{ if ne $profile.name "default" }}-{{ $profile.name }}{{ end }}
spec:
  ...
```
And we update the daemonset node affinity to make sure no node ends up with two agents and every node does get an agent:
```yaml
 affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
          - matchExpressions:
              {{ if eq $profile.name "default" }}
              # Catch-all: matches nodes that have no profile label at all
              - key: "cilium.io/agent-profile"
                operator: DoesNotExist
              {{ else }}
              - key: "cilium.io/agent-profile"
                operator: In
                values:
                  - {{ $profile.name }}
              {{ end }}
```
And set the resources according to the profile:
```yaml
resources: {{ toJson $profile.resources }}
```

The main advantage of this design is that it's highly flexible:
* configure nodes to be labeled at provisioning with the appropriate label to opt-in to a sized profile (for example `cilium.io/agent-profile: small`)
* you can configure as many different profiles as you want and need
* if you don't set any custom profile, you just end up with the default one and everything renders as it does today, so no change for people not needing this advanced setup
* single chart solution: no duplicate chart, no duplicate RBAC resources

mrkiani98: thanks for the details. Actually for now, we also have a kind of similar approach with some custom in place implementation to have the multiple agent daemonsets with different resources set. But in general, the goal of this feature request is to make the chart dynamic enough to support this implementation without any custom patched setup.

joestringer:
Discussion from 2026-08-12 community meeting:

Seems like at this stage it's possible for users to implement split daemonsets with existing functionality, but before we make this a first-class feature it would be good to gather more use cases so that we can confirm several users can benefit from the solution.

PR #76, opened 2025-09-11
CFP-41634: Datapath Plugins

jrife:
## Summary

Allow operators of Cilium to integrate their own BPF programs into the Cilium datapath to enable custom BPF-based observability, encapsulation, and routing solutions.

## Motivation

As a cloud provider, we operate thousands of Cilium-enabled clusters in an environment where other teams and even savvy customers want to deploy their own BPF-based solutions alongside Cilium for various purposes. Their motivations are varied, but tend to fall into several categories:

* Custom transparent proxies
* Custom observability
* Custom encapsulation and routing

Cilium isn't the only component we manage, and while it forms the core of our datapath, we need to combine it with other systems. Up until now, we've needed to steer other teams and customers away from using BPF at all in their solutions, as there was no good way to deploy it and too many places where Cilium could get in the way.

BPF continues to increase in popularity and others want to use it to solve their problems. This proposal seeks to make this easier.

...

[continued in CFP]

PR #92, opened 2026-04-08
CFP-44188: Vtep Improvements with CRD

parlakisik:
Replace the static CLI flag-based VTEP configuration with a cluster-scoped CiliumVTEPConfig CRD that supports dynamic updates, per-node assignment via nodeSelector, and per-endpoint status reporting.

 Signed-off-by: Murat Parlakisik <parlakisik@gmail.com>


xmulligan:
cc: @cilium/sig-datapath

cilium/CFP-44188-ciliumvtepconfig-crd.md
```diff
@@ -0,0 +1,296 @@
...
+
+### The Case for Overlay-to-Gateway Simplicity
+
+If user doesnt want to manage BGP or L2 annoutment to send traffic to some network via  external gateway.
+The VTEP approach offers a fundamentally simpler model:
+
+Pods send traffic via the existing VXLAN overlay directly to an external
+vtep endpoint. No BGP sessions to configure and maintain. No L2 announcement
+policies. No route redistribution. The Cilium agent simply encapsulates
+traffic destined for external CIDRs and sends it to a known VTEP endpoint.
```
joestringer: BGP and L2 announcements features are about advertising addresses on connected networks in order to configure remote clusters how to transmit towards Cilium. This feature rather seems to be about how to configure Cilium in order to route traffic _from_ Cilium. Are they really equivalent?

cilium/CFP-44188-ciliumvtepconfig-crd.md
```diff
@@ -0,0 +1,296 @@
...
+
+5. It diffs desired state against **last-applied state** and performs
+   incremental BPF map updates:
+   - New CIDRs → `UpdateEntry()`
+   - Changed endpoints → `UpdateEntry()` (overwrite)
+   - Removed CIDRs → `DeleteByCIDR()`
+
+6. It updates Linux routing table entries for VTEP CIDRs.
+
+7. It writes per-endpoint status back to the CRD's `.status` subresource.
```
joestringer: As a general statement, we avoid putting logic into cilium-agent to update `.status` because as you scale up, this causes significant load and conflicts on kube-apiserver due to competing agents attempting to make similar updates, often at the same time.

cilium/CFP-44188-ciliumvtepconfig-crd.md
joestringer, line 1: I have two high level thoughts, related, which I wonder about for the overall architecture. I don't have a strong opinion on these approaches, but they seem within the possible design space, so they're worth considering:

1. Could Cilium integrate natively to the Linux stack to delegate routing of this traffic to the Linux routing table, then have another component sync the desired state into the kernel routing table?
2. Alternatively if this is difficult due to VNI selection, could perhaps [datapath plugins](https://github.com/cilium/design-cfps/pull/76) provide an alternative integration mode?

cilium/CFP-44188-ciliumvtepconfig-crd.md
```diff
@@ -0,0 +1,296 @@
...
+   matching configs, neither is applied and both configs receive an error
+   status.
+
+5. It diffs desired state against **last-applied state** and performs
+   incremental BPF map updates:
+   - New CIDRs → `UpdateEntry()`
+   - Changed endpoints → `UpdateEntry()` (overwrite)
+   - Removed CIDRs → `DeleteByCIDR()`
+
+6. It updates Linux routing table entries for VTEP CIDRs.
```
joestringer: This actually seems like the crux of the functionality proposed here (+ the encap/decap config). I wonder whether we really want to have a dedicated VTEP CRD for this or whether it's time to consider a more generic "routing" CRD (even if initially focused just on the VTEP use case). I had a draft for such an idea about three years ago, but it never got traction so I didn't end up posting it publicly. But if this is interesting, maybe I can dust it off.

cilium/CFP-44188-ciliumvtepconfig-crd.md
```diff
@@ -0,0 +1,296 @@
...
+| Property | Hash | LPM Trie |
+|---|---|---|
+| Lookup semantics | Exact match | Longest prefix match |
+| Key size | 4 bytes | 8 bytes (4 prefix + 4 IP) |
+| Preallocation | Supported | Not supported (`BPF_F_NO_PREALLOC` required) |
+| Max entries | 8 | 8 |
+
+The LPM Trie is strictly more capable. Existing configurations with uniform
+prefix lengths produce identical routing behavior.
+
```
joestringer, line 364: Key question: How will this integrate with the broader architecture? Ideally whatever we come up with, it has a path to eventually integrate with all other GA features where applicable.

Specifically, consider how does this interface with masquerading, egress gateway, encryption?

cilium/CFP-44188-ciliumvtepconfig-crd.md
```diff
@@ -0,0 +1,296 @@
...
+**Authors:** Murat Parlakisik <parlakisik@gmail.com>
+
+**Status:** Draft
+
+## Summary
+
+Replace the static CLI flag-based VTEP configuration with a cluster-scoped
+`CiliumVTEPConfig` CRD that supports dynamic updates, per-node assignment via
+`nodeSelector`, and per-endpoint status reporting. This enables production use
```
joestringer: is "endpoint" here a "VTEP endpoint"? I would suggest not using shorthand, because endpoint is already overloaded (twice - k8s and cilium have established meanings for this word which are not fully aligned)

cilium/CFP-44188-ciliumvtepconfig-crd.md
```diff
@@ -0,0 +1,296 @@
...
+1. **CiliumVTEPConfig CRD** — a cluster-scoped custom resource that declares
+   VTEP endpoints with optional node targeting
+2. **VTEPReconciler** — an agent-side controller that watches CRD events,
+   evaluates `nodeSelector`, and reconciles the BPF map
+3. **BPF LPM Trie map** — replaces the existing Hash map to support
+   variable-length prefix matching
+
+### CRD API
+
+```yaml
```
joestringer, line 134: I won't review the API now since there's plenty of other open questions to consider first, but the API will need to be reviewed.

cilium/CFP-44188-ciliumvtepconfig-crd.md
joestringer: Probably this is a case for understanding the tradeoff with @cilium/sig-scalability , especially size of targeted environments, and then considering how we might gain the desired operational visibility without introducing scalability concerns.

joestringer:
Converting to draft due to inactivity.

parlakisik:
I will work on that. I am in another conference

cilium/CFP-44188-ciliumvtepconfig-crd.md
parlakisik: Agree  The term is a "VTEP endpoint" (VXLAN tunnel endpoint) — I'll use that consistently and drop the bare "endpoint"

cilium/CFP-44188-ciliumvtepconfig-crd.md
parlakisik: They're not equivalent, but they're all connectivity mechanisms. VTEP is a bidirectional tunnel.
Iit fits environments where the advertise-based approaches don't apply: no BGP-speaking upstream to peer with and no L2 adjacency to the destination (L2 announcement is ARP-based and single-segment). In those cases VTEP gives pods a simple two-way overlay path to an external gateway over a routed underlay. 
I will reframe "The Case for Overlay-to-Gateway Simplicity" around connectivity alternatives with different mechanisms, rather than implying equivalence.

cilium/CFP-44188-ciliumvtepconfig-crd.md
parlakisik: I will move the api change to  v2alpha1 since this is a brand-new, unreviewed surface

cilium/CFP-44188-ciliumvtepconfig-crd.md
parlakisik: I would love to see routing CRD case . it may match with this changes .
My suggestion is , shipping  the VTEP work as CiliumVTEPConfig in v2alpha1 now and treat the generic routing direction as the next step .

cilium/CFP-44188-ciliumvtepconfig-crd.md
parlakisik: Thats a good catch . i will update this one as adding a per-node CiliumVTEPNodeConfig : one object per node,  Status holds the resolved endpoints plus per-endpoint sync state . Agent reconciles only after a local BPF map sync completes.

I a m trying to align with the pattern defined for CiliumBGPNodeConfig.

cilium/CFP-44188-ciliumvtepconfig-crd.md
parlakisik: in initial release , i focused on much more crd models . but i will improve this in the next releases

parlakisik:
Is there any suggestion with new changes ?

bowei:
Discussed on the call -- author is going to take fresh look to see if this could be implemented as a datapath plugin.

PR #93, opened 2026-04-24
CFP-45607: enable delegated-ipam in conjunction with Gateway API

matmerr:
tl;dr: when running with delegated ipam enabled, allow delegated ipam to also allocate ingress IP's for gateway API.

Cilium feature issue here: https://github.com/cilium/cilium/issues/45607

Implementation here: https://github.com/cilium/cilium/pull/45818

### DIFF
--- a/cilium/CFP-45607-delegated-ipam-ingress.md
+++ b/cilium/CFP-45607-delegated-ipam-ingress.md
@@ -63,13 +63,26 @@ For ingress IPs we use `cilium-ingress-<node-name>` we use as `CNI_CONTAINERID` 
 
 The other CNI parameters have defaults in the CNI call since we're only doing IPAM, not actual network plumbing:
 
-* `CNI_NETNS` = `/proc/1/ns/net` (host network namespace)
+* `CNI_NETNS` = path to an ephemeral, empty network namespace created by the agent for the duration of each ADD/DEL invocation (see [Ephemeral network namespace](#ephemeral-network-namespace) below)
 * `CNI_IFNAME` = `eth0` (must match between ADD and DEL, since many IPAM plugins key bookkeeping on container ID + ifname)
 * `K8S_POD_NAME` = `cilium-ingress-<node-name>`
 * `K8S_POD_NAMESPACE` = `kube-system`
 
-We also append [`CNI_NETNS_OVERRIDE=1`][cni-netns-override] to the environment, to avoid CNI spec validation around actual netns creation.
+### Ephemeral network namespace
 
+For each ADD/DEL the agent creates a fresh, empty network namespace using the same `pkg/netns` helper that cilium-health uses ([`netns.New()` in `pkg/health/health_connectivity_endpoint.go`][cilium-health-netns-new]), passes its path as `CNI_NETNS`, and closes the handle as soon as the plugin returns. Because no process ever runs inside it and the agent never pins it, the kernel destroys the namespace once the last reference is dropped.
+
+Although the spec is an IPAM-only delegation (no network plumbing is intended), the ephemeral netns is created as a precaution.
+
+### Plugin execution context
+
+The IPAM plugin binary is invoked as a child process of `cilium-agent`, which means it runs **inside the cilium-agent container**, not on the host. This has a few implications that operators of delegated IPAM plugins should be aware of:
+
+* Any filesystem paths the plugin reads or writes (state files, sockets, config) must be reachable from inside the cilium-agent container.
+* The plugin inherits the cilium-agent container's network namespace, user, and capabilities for everything except `CNI_NETNS` (which points at the ephemeral netns described above).
+* Plugins that shell out to other binaries must have those binaries available inside the container as well.
+
+This is a behavioral difference from the standard pod IP CNI flow, where the plugin is invoked by the kubelet/container runtime on the host and therefore sees host paths directly.
 
 ### Plugin binary path and volume mounts
 
@@ -199,6 +212,7 @@ The early [`UpdateCiliumNodeResource()`][daemon-l151] is gated to ClusterPool/Mu
 [cilium-ipam-block]: https://github.com/cilium/cilium/blob/70ae8d0ef536de807aab849291e5a68758cb8d4d/pkg/option/config.go#L3803
 [cni-netns-override]: https://github.com/containernetworking/cni/pull/890
 [cni-spec-lifecycle]: https://github.com/containernetworking/cni/blob/main/SPEC.md#lifecycle--ordering
+[cilium-health-netns-new]: https://github.com/cilium/cilium/blob/27eb281eb554ef7f00180a108ecdb404fbafa8fe/pkg/health/health_connectivity_endpoint.go#L286
 [daemon-l151]: https://github.com/cilium/cilium/blob/a1afcb4b88db6d3160b4c315ad8796748b30fbe8/daemon/cmd/daemon.go#L151-L156
 [daemon-l158]: https://github.com/cilium/cilium/blob/a1afcb4b88db6d3160b4c315ad8796748b30fbe8/daemon/cmd/daemon.go#L158
 [daemon-l217-2]: https://github.com/cilium/cilium/blob/a1afcb4b88db6d3160b4c315ad8796748b30fbe8/daemon/cmd/daemon.go#L217-L224
--- a/cilium/CFP-45607-delegated-ipam-ingress.md
+++ b/cilium/CFP-45607-delegated-ipam-ingress.md
@@ -1,0 +1,215 @@
+# Delegated IPAM for Gateway API Ingress IPs
+
+**SIG: sig-agent**
+
+**Begin Design Discussion:** 2026-04-21
+
+**Cilium Release:** 1.20
+
+**Authors:** Apurup Chevuru <achevuru@microsoft.com>, Mathew Merrick <matmerr@microsoft.com>
+
+**Status:** Implementable
+
+**Related CFP:** [Delegated IPAM Cilium IPs][cfp-delegated-ipam]
+
+
+## Summary
+
+Enable Gateway API / Ingress Controller (`--enable-envoy-config`) when Cilium is running with `--ipam=delegated-plugin`, by having cilium-agent invoke the delegated IPAM plugin directly to allocate and release ingress IPs.
+
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


xmulligan:
cc: @cilium/sig-agent

cilium/CFP-45607-delegated-ipam-ingress.md
```diff
@@ -0,0 +1,161 @@
+# Delegated IPAM for Gateway API Ingress IPs
+
+**SIG: SIG-Agent, SIG-CNI**
```
xmulligan: What is the SIG-CNI you are referencing here?

cilium/CFP-45607-delegated-ipam-ingress.md
```diff
@@ -0,0 +1,161 @@
...
+
+**Begin Design Discussion:** 2026-04-21
+
+**Cilium Release:** 1.18
```
xmulligan: 1.18 has already been released

cilium/CFP-45607-delegated-ipam-ingress.md
```diff
@@ -0,0 +1,161 @@
...
+
+**Authors:** Apurup Chevuru <achevuru@microsoft.com>, Mathew Merrick <matmerr@microsoft.com>
+
+**Status:** Draft
+
+**Related CFP:** [CFP-TODO: Delegated IPAM CiliumIPs][cfp-delegated-ipam]
```
xmulligan: should this still be a todo?

cilium/CFP-45607-delegated-ipam-ingress.md
matmerr: These were SIG's Will formerly referenced in [the "TODO CFP"](https://github.com/cilium/design-cfps/blob/main/cilium/CFP-TODO-delegated-ipam-cilium-ips.md) which this CFP revises/retargets, but the SIG architecture seems to have changed since then. 

But yeah, @xmulligan maybe you can point this in the right direction. The [CFP template](https://github.com/cilium/design-cfps/blob/main/cilium/CFP-TODO-delegated-ipam-cilium-ips.md) links out to this page, which links to [this doc](https://docs.cilium.io/en/stable/community/community/#all-sigs) on the cilium wiki, which has an broken link to [this section](https://docs.cilium.io/en/stable/community/community/#special-interest-groups), which points out to [SIG.md](https://github.com/cilium/community/blob/main/SIG.md). There I just see sig-community, `sig-community`, `sig-policy`, and `sig-scalability`. 

I'll update to just `sig-agent` like mentioned above

cilium/CFP-45607-delegated-ipam-ingress.md
matmerr: Bumped to 1.20

cilium/CFP-45607-delegated-ipam-ingress.md
matmerr: It's less of a TODO for this doc, and more of a friendly reference to a [related merged CFP](https://github.com/cilium/design-cfps/blob/main/cilium/CFP-TODO-delegated-ipam-cilium-ips.md) which was never assigned a CFP number and explicitly has TODO in the filename :)

I don't think Will went through the process (or maybe it didn't exist back then?) and it has the title of CFP 003, which is the same as [the template](https://github.com/cilium/design-cfps/blob/main/cfps/CFP-003-template.md). 

I'll just update this line to `Delegated IPAM CiliumIPs`

cilium/CFP-45607-delegated-ipam-ingress.md
```diff
@@ -0,0 +1,116 @@
...
+        Agent->>Plugin: Find plugin binary on configured path
+        Agent->>Plugin: CNI ADD (ContainerID=cilium-ingress-<node>, NetNS=/proc/1/ns/net)
+        Plugin-->>Agent: Allocated IPv4/IPv6 addresses
+        Agent->>Agent: Store ingress IPs on local node
+    end
+```
+
+### Allocation notes
+
+* If ingress IPs are already present in memory, skip allocation to avoid leaking IPs via duplicate CNI ADD calls.
```
tamilmani1989: even if it calls delegated ipam cni again, there won't be leak since you are using same container ID?

cilium/CFP-45607-delegated-ipam-ingress.md
```diff
@@ -0,0 +1,116 @@
...
+
+* Configurable flag for `--delegated-ipam-cni-bin-path` since this is outside of the CNI chaining lifecycle.
+
+
+### Deallocation
+
+Deallocation issues a CNI DEL with the same stable container ID.
```
tamilmani1989: so delete will be invoked always even if gateway API disabled and if `delegated-ipam-bin-path` set right?

what if some one allocated cilium ingress ip and accidentally removed `delegated-ipam-bin-path`  and restart cilium. will that leak ip?

cilium/CFP-45607-delegated-ipam-ingress.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Skipping identity restoration:
+
+We also need to consider skipping identity restoration from the [local IP cache.](https://github.com/cilium/cilium/blob/6f1c09105e01105d8dca315036a244505997e893/pkg/ipcache/restore/local_identity_restorer.go#L187).
```
tamilmani1989: is this different only for ingress IP allocated via delegated ipam? if so, why it's different for this case?

joestringer:
@cilium/sig-ipam could you also take a look? Seems like you might have some thoughts or insights into the structure.

cilium/CFP-45607-delegated-ipam-ingress.md
```diff
@@ -0,0 +1,215 @@
...
+### Stable container ID
+
+For ingress IPs we use `cilium-ingress-<node-name>` we use as `CNI_CONTAINERID` for a determinstic node-scoped reproducable key, which is appropriate for a single ingress IP.
+
+* Is stable across agent restarts on the same node, so IPAM plugins that key allocations by container ID return the same IP and avoid IP churn.
+* Is unique per node, avoiding collisions in delegated IPAM plugins backed by a cluster-wide store (where `CNI_CONTAINERID` may be a global key).
+
+The other CNI parameters have defaults in the CNI call since we're only doing IPAM, not actual network plumbing:
+
+* `CNI_NETNS` = `/proc/1/ns/net` (host network namespace)
```
gandro: I'm a bit worried about this. What if the delegated IPAM plugin starts messing with `eth0` in the host network namespace (for example, removes IPs from eth0)?

Have you considered spawning a fake network namespace for the ingress IP instead? Similarly to how Cilium does it with the cilium-health endpoint?

gandro (commented):
Thanks, from an IPAM perspective this looks reasonable to me. I have one concern though.

cilium/CFP-45607-delegated-ipam-ingress.md
gandro: On second though, it's only an IPAM plugin, not a full CNI delegation. So maybe this is less bad than I thought.

cilium/CFP-45607-delegated-ipam-ingress.md
matmerr: Correct, though in reality subject to the implementation of the IPAM provider for preserving the mapping of container ID to IP.

cilium/CFP-45607-delegated-ipam-ingress.md
matmerr: DEL only fires when `--enable-envoy-config=false && --ipam=delegated-plugin` is set, and if the binary is missing or the call errors out we log a warning.

If an ingress IP is allocated, and the access to delegated-ipam is removed, then that IP will inherently be leaked until access is restored combined with cilium restart, or IPAM provider state is reset.

cilium/CFP-45607-delegated-ipam-ingress.md
matmerr: Yes, the intention here is to manage all related aspects of state for this IP (to ensure state is in sync with IPAM provider), since the entire lifecycle of this IP is managed by cilium. This is opposed to pod IP's which are tied to the pod lifecycle managed by kubelet, and create/delete/restore calls from kubelet are maintained in the queues on disk even in the event of cilium restart.

cilium/CFP-45607-delegated-ipam-ingress.md
matmerr: Correct, only IPAM, no network plumbing intended. Additionally, the netns validation is omitted with `CNI_NETNS_OVERRIDE=1` being passed in the IPAM call, which was introduced into CNI skel parameters [here](https://github.com/containernetworking/cni/pull/890).

gandro: Makes sense. Still, as an additional safety measure against misbehaving plugins, would it make sense to create a dummy network namespace for the ingress IP?

matmerr: I think it's a reasonable precaution, we can probably reuse the same [netns package](https://github.com/cilium/cilium/blob/64a2eb910b92b74978ade9c0baf0dfc1c8f47516/pkg/health/health_connectivity_endpoint.go#L286) that health endpoint is using.

Do we know of any IPAM providers which do depend on a real netns? Curious if there are any other guardrails needed while we're at it.

gandro: > Do we know of any IPAM providers which do depend on a real netns? Curious if there are any other guardrails needed while we're at it.

I have to confess, I have not looked at any delegated IPAM plugins. It is very possible that all of them are well-behaved and I'm overly paranoid wanting to put them into a dedicated network namespace. But it seems more clean to me rather than than telling plugin to allocate for the host network namespace.

gandro (approved):
Thanks! I think it would also be fine to keep the network namespace around (the plugin might get confused if ADD and DEL use a different namespace?), but my concern was mainly around sandboxing, so approving from my side.

matmerr:
> Thanks! I think it would also be fine to keep the network namespace around (the plugin might get confused if ADD and DEL use a different namespace?), but my concern was mainly around sandboxing, so approving from my side.

That's fair, the delegated e2e's use `host-local` for the delegated IPAM, so during validation phase with gateway API I'll check with that one and see if there's any issues with different ephemeral namespaces.

gandro:
Status is implementable and committer approval has been given. Merging!

PR #94, opened 2026-05-05
CFP: Cilium policy selector expression

fristonio:
Draft PR: https://github.com/cilium/cilium/pull/46199

### DIFF
--- a/cilium/CFP-TODO-policy-selector-expressions.md
+++ b/cilium/CFP-TODO-policy-selector-expressions.md
@@ -1,0 +1,152 @@
+# CFP-XXXXX: Policy Selector Expressions
+
+**SIG: SIG-Policy**
+
+**Begin Design Discussion:** 2026-05-04
+
+**Cilium Release:** TBD
+
+**Authors:** Deepesh Pathak
+
+**Status:** Draft
+
+## Summary
+
+This CFP proposes extending [Cilium's policy selectors](https://github.com/cilium/cilium/blob/f7c0ac66d9d781ecb5bd72c59025620b01c912d8/pkg/policy/api/selector.go#L17), allowing users to write complex selection expressions than what is supported today with [`LabelSelector`](https://github.com/cilium/cilium/blob/f7c0ac66d9d781ecb5bd72c59025620b01c912d8/pkg/k8s/slim/k8s/apis/meta/v1/types.go#L267) - specifically operators such as prefix, suffix, and substring matching, along with full boolean composition.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


squeed:
Should we set a cost limit for CEL expressions? It could be caught in the Sanitize function().

We can always increase the cost limit, but never decrease it :-/.

joestringer:
Could you also draft some specific user stories about selections they would like to do? I think that would help to inform what the solution should look like in order to satisfy the intent of our users.

cilium/CFP-TODO-policy-selector-expressions.md
```diff
@@ -0,0 +1,115 @@
...
+
+##### Cons
+
+- The parser implementation needs to be maintained in cilium source tree.
+- Language spec is not standard and users must learn a new format.
+
+#### Option 2: CEL Expression
+
+A more natural option for the use-case is to use CEL(Common Expression Language) based expression for label selectors. Cilium already imports the `cel-go` library and has experimental support for CEL based expression language in hubble filters. This option will give us the opportunity to build a strong foundation for CEL usage in Cilium project, allowing easier adoption for similar usecases in future.
+
```
squeed, line 83: We should probably include some examples here, along with the expected CEL variables and functions we intend to expose.

fristonio: I was thinking of keeping the implementation details separate from the high level CFP. I am working on the implementation of [CEL env here](https://github.com/cilium/cilium/compare/main...fristonio:cilium:pr/fristonio/cel-env). Do you think it makes sense to keep this detailed documentation there(working on a package level README) or should I include everything here?

squeed: I would say that the CEL environment is probably worth briefly discussing, since it's essentially unchangeable once landed. We don't need to consider every possible function

squeed:
So, I'm thinking about the existing `source:key` convention for embedding sources in to keys. It works, but I don't particularly like it. In particular, the source is **not** part of the key. It is metadata attached to the **value**. (Side note: we get very lucky that `1.1.1.1/32` is not a valid k8s label, since we rely on that for CIDR selectors. In the future, I could even see us allowing multiple values for a given key, though this may not be worth the pain.)

Anyways.

From your comments, it seems that looking up labels via function calls is expensive. What the environment had a single variable, `labels`:

```go
map[string]struct{
  Value string `cel:"value"`
  Source string `cel:"source"`
}

	env, err := cel.NewEnv(
		ext.NativeTypes(
			ext.ParseStructTag("cel"),
			reflect.TypeFor[*Label](),
		),
		cel.Variable("labels",
			cel.MapType(cel.StringType, cel.ObjectType("main.Label"))),
	)
```

We could even add some useful macros, like `label(key)` expanding to `labels[key].value`

cilium/CFP-TODO-policy-selector-expressions.md
fristonio: I added some examples with the link to draft PR. For now the plan is to expose just a simple macro that allows label value lookup based on key in the same format as supported in `k8s.LabelSelector`(`<source>:<key>`).

jrajahalme (approved):

joestringer:
I've marked this PR as draft. When you have addressed the comments, bring the PR back to the attention of reviewers by scrolling to the bottom of the page and clicking the "Ready for Review" button. Thanks!

PR #96, opened 2026-05-13
cilium: add CFP-34702 per-endpoint VLAN

l1b0k:
Propose per-endpoint VLAN support so that a single tc/tcx attachment on the parent NIC can serve pods belonging to different VLANs, with policy and observability preserved.

Tracks cilium/cilium#34702 and cilium/cilium#44619.


xmulligan:
cc: @cilium/sig-datapath

squeed:
How do you foresee this interacting with load balancers? They can take on many forms:
- ServiceIPs
- NodePort
- LocalRedirectPolicy

joestringer:
Marked as draft awaiting further discussion on the above questions.

brb:
@l1b0k Are you aware of the [Datapath Plugins](https://github.com/cilium/design-cfps/pull/76) proposal? It's currently being implemented. But once it's done, I think it perfectly matches your need for the datapath needs. It allows you to add `PRE` and `POST` hooks.

l1b0k:
Thanks , I've gone through the Datapath Plugins CFP  in detail and it looks like a viable path for this.

PR #97, opened 2026-05-25
CFP-12781: Host Firewall Before NodePort DNAT

jakubhlavnicka:
Add CFP for enforcing host firewall ingress policy before NodePort DNAT/SNAT, which enables _CiliumClusterwideNetworkPolicy_ to match on original external source IPs and NodePort destination ports.

Tracks: https://github.com/cilium/cilium/issues/12781


joestringer:
My gut feeling is this is not something we can add a flag for and reasonably continue maintaining and supporting. More likely we need to consider new API syntax to match the relevant hook point.

cc @cilium/sig-policy we discussed this during the [most recent meeting](https://zoom.us/rec/share/BY1kxmt4JbkXjN4kbMFzxHoLPE5-SspFfFylPaoYJyMVz4JGg8RclG5xPGLIJQ1w.JqUO7VFB5EmZC-Ib).

julianwiedmann (changes requested):
I'm not seeing any consideration for `loadBalancerSourceRanges`. Can you describe why it's not sufficient for your purposes?

jakubhlavnicka:
@julianwiedmann - I've added a section to the CFP addressing why `loadBalancerSourceRanges` is not sufficient for this use case: https://github.com/cilium/design-cfps/pull/97/changes/f3894a4ee198dc88a7920ede251735b90dcc6bf0

@joestringer - Thank you for the feedback. Are you suggesting a new policy API field on `CiliumClusterwideNetworkPolicy`, something like `preSNATIngress` with the same syntax as the existing `ingress` field, that would explicitly target the pre-SNAT hook point? Or rather an entirely new policy kind for this purpose?

joestringer:
Reviewing stale discussions during community meeting 2026-08-12: This needs to be driven in coordination with @cilium/sig-policy , either through async review or discussed in an upcoming monthly SIG meeting.

jakubhlavnicka:
@joestringer thanks for picking this up in the community meeting, and sorry for the slow turnaround.

I've pushed a substantial revision that addresses the review feedback — specifically the concern about exposing the pre-NAT hook point as a policy API field rather than as a configuration flag that silently changes the meaning of the existing ingress field. The new document is _cilium/CFP-12781-pre-nat-host-policy.md_.

Summary of what changed:

 - New API field instead of a config flag. The design now proposes a _preNATIngress_ field on _CiliumClusterwideNetworkPolicy_ (type _[]IngressRule_, same syntax as ingress) rather than _hostFirewall.enforceBeforeNodePortDNAT_. Existing policy objects are unaffected — the new stage is only enforced when a policy explicitly populates the field, so there's no opt-in flag, no migration, and no breaking change.
 - Naming. The field was suggested in review as preSNATIngress; I've written it up as preNATIngress since the hook precedes both DNAT and SNAT, and the primary motivating rule matches a destination port, which is a DNAT concern. I've left this as an explicit open question with pros/cons for both, plus a third option (a per-rule enforcementStage discriminator).
 - Explicit comparison section between the config-flag and policy-field designs, covering portability, expressiveness, and API surface cost.
 - Validation and observability: L7/auth rejected at this hook point, a distinct _DROP_POLICY_PRE_NAT_ reason, Hubble stage discriminator, and a status condition when _preNATIngress_ is present but the stage is unavailable on a node.
 - Open questions written up for discussion rather than settled: scope of the stage (all native-device traffic vs. only LB-frontend traffic), behaviour for encapsulated/encrypted traffic, whether an egress counterpart is needed, and new field vs. new CRD (the RBAC argument for a separate kind is the strongest counter to the proposal as written).
 - Retained and expanded the analysis of why _loadBalancerSourceRanges_ doesn't cover these use cases.

Happy to take this to @cilium/sig-policy either way — async review here works for me, or I'm glad to walk through it at an upcoming monthly SIG meeting if that's more productive. Just let me know which you'd prefer and I'll follow up accordingly.

PR #99, opened 2026-06-13
CFP-46532: Synthetic Probing Subsystem for Connectivity Testing

arighnac:
## Summary

An observability subsystem that allows an user to generate out-of-band synthetic unidirectional flows to perform on-demand connectivity check/test across running workloads, with zero intrusion on the user workloads. The synthetic flows needs to support a wide range of CUJs, be observable via the Hubble sub-system, and generate more fine-grained debugging data from the Cilium dataplane.

## Motivation

- **On-Demand Connectivity Validation**: Allow developers and operators to quickly generate network flows across multiple endpoints via an API. This enables users to perform "on-demand" connectivity status checks between any two workloads with less friction. This also enables users to perform network validations before productionalizing workloads on k8s clusters.  
    
- **Troubleshooting in the Absence of Live Traffic**: A challenge in resolving network connectivity failures for users is the indeterminate volume of traffic at the time of reproduction. By generating synthetic flows, engineers can proactively induce and diagnose issues without waiting for user traffic, expediting root cause analysis.

- **Extract Important Debugging State out of the Dataplane:** Synthetic probes can act as a great interface to generate fine-grained debugging state out of Cilium, given that the generation of such packets would be controlled by the API.

(Discussion continued in CFP.....)


arighnac:
cc @jrife

msune:
MBOI:

* https://github.com/cilium/cilium/pull/44059

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
```diff
@@ -0,0 +1,779 @@
...
+
+* The reconciler **locates** the source endpoint:  
+  * If `spec.SourceEndpoint.IP` is specified, the reconciler resolves the IP to either a pod or a cluster node. It throws an error and updates the status to reflect if resolution fails.  
+  * If `spec.SourceEndpoint.K8sPodKey` is specified, the reconciler retrieves the pod object. If no pod object is found, it throws an error and updates the status.
+
+* The reconciler **updates** `status.SourceNode` to the node hosting the source endpoint.
+
+* `status.SourceNode` is a [selectable field](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/). cilium-operator also establishes filtered watches on the kube API server for any event that updates the field selector `status.SourceNode` to empty value. It also sets a client predicate to ensure that the older value and the new empty value are different.
+
+### flowgen Daemonset
```
jrife: We should just put this into the Cilium agent. Why a standalone `DaemonSet`?

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
```diff
@@ -0,0 +1,779 @@
...
+        // The user-specified destination port. Populated from the port specified in CiliumFlowGen.Spec.FlowProfile
+        __be16 user_dport;
+
+        // The unique identifier for the flow. Used to tag packets and look up flow information in BPF maps. Populated from CiliumFlowGen.Spec.TraceID.
+        __u64 trace_id;
+}
+
+```
+
+* The agent sets a watch on the kube API-server for any Delete event of the source pod. This step is to ensure that the `flowgen` sub-system is able to follow pod reschedules/restarts. Pod delete is the only way for a pod’s IP or host node to be updated.
```
jrife: We would end up using statedb to synchronize map state with API state, etc. if this logic was in the Cilium agent.

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
```diff
@@ -0,0 +1,779 @@
...
+
+The `flowgen-client` would be architected in such a way that it doesn’t expect return packets for any protocol (since it can only generate unidirectional synthetic probes). For example,
+
+* **TCP**: The client creates a `SOCK_STREAM` socket, and attempts to establish a TCP handshake. SYN retries are set to 1, to ensure no retries because packets would be dropped at the server end.
+
+* **UDP**: The client creates a `SOCK_DGRAM` socket. Since UDP is connectionless, the client simply invokes `sendto()`. The kernel prepends the configured IPv4 options to the IP header of every outgoing datagram.
+
+## Section 5: Synthetic Probing: Dataplane
+
+### Section 5.1: Dataplane Init Steps
```
jrife, line 538: Why not just deploy the `flowgen-client` as a non-host-network pod and use the normal loader machinery to do the BPF programming. We might do something like plugins to load `cil_from_flowgen`+`cil_to_flowgen` in place of the normal `cil_from_container`...

If not a pod then maybe something similar to the Cilium health check endpoint where we have the Cilium agent managing all of this then the flowgen client "endpoint" is programmed as any other would be. That way we could eliminate the client needing to be a standalone pod.

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
```diff
@@ -0,0 +1,779 @@
...
+  * Validate and Store the value as `flowgen_trace_id`.
+
+* Retrieve the user-specified 4-tuple (`user_saddr`, `user_daddr`, `user_sport`, `user_dport) from the `flowgen_map` BPF map, using `flowgen_trace_id` as the key.
+
+* Retrieve source endpoint info from `ENDPOINTS_MAP` BPF Map, using `__lookup_ip4_endpoint(ip)`. Lets call the source endpoint as `sep`.  
+* Update source and destination MAC of the packet to `sep->mac` and `sep->node_mac` respectively. This step updates the MAC headers to the user source endpoint and makes it look like the packet came from the user source pod network namespace.   
+* Modify the packet source and destination IP’s and ports to the user 4-tuple, using `ctx_store_bytes`.
+
+* BPF-Redirect the packet to the ingress hook of the user source pod lxc interface.  
+  * `ctx_redirect(ctx, sep->ifindex, BPF_F_INGRESS)`
```
jrife: This ingress redirection won't work with netkit, since it would skip `cil_from_container` (I realize this is why you limited the scope initially to just `veth`), but considering [this patch](https://lore.kernel.org/bpf/20260618182035.43811-1-jordan@jrife.io/) was merged to support `BPF_F_EGRESS` with `bpf_redirect_peer`, it's probably worth mentioning how we intend to make this work with netkit now. We'd basically just swap out one or the other depending on the datapath mode and would need a startup probe that disallows the use of flowgen with netkit if the system's kernel doesn't have `BPF_F_EGRESS` support.

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
```diff
@@ -0,0 +1,779 @@
...
+
+* Packet hits the `cil_from_container` eBPF prgoram attached to the lxc interface of the user source pod.
```
jrife: "prgoram" -> "program"

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
arighnac: Yes, it makes sense to punt `flowgen-agent` into `cilium-agent` itself. We can keep this as a separate Hive, and just ensure that the `flowgen-agent` watchers can run as isolated background goroutines so that it doesn't impact cilium-agent.

The daemonset serves multiple purposes though, both an agent as well as a `flowgen-client` which generates the actual synthetic packets. It seems there are two options: have the client embedded within `cilium-agent` itself, or have the `flowgen-client` as a standalone process launched by `cilium-agent` and communication between the two over an UDS. I am erring towards having a separate process to perform synthetic traffic generation. Let me know what you think!!

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
arighnac: Agreed. Responded in your eariler comment.

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
arighnac: > Why not just deploy the flowgen-client as a non-host-network pod and use the normal loader machinery to do the BPF programming. We might do something like plugins to load cil_from_flowgen+cil_to_flowgen in place of the normal cil_from_container...

I am erring towards the side of this. The flowgen-client can be deployed within a `netns`, and cilium can just plug in a `flowgen` veth pair, on which we can load the special BPF programs using plugins. I hope the deployment of `flowgen-client` will still be managed by cilium, since we do not want users to deploy this separately???

> If not a pod then maybe something similar to the Cilium health check endpoint where we have the Cilium agent managing all of this then the flowgen client "endpoint" is programmed as any other would be. That way we could eliminate the client needing to be a standalone pod.

I am erring towards Cilium managing deployment of `flowgen-client` process in an isolated namespace along with loading the `flowgen` BPF programs on the special `flowgen` veth pair. Wdyt?

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
arighnac: @jrife Thanks a lot for the patch!! I wil update the CFP with this.

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
arighnac: Updated.

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
arighnac: I have updated this in the CFP to remove the `flowgend` daemonset deployment, and instead have `flowgen-agent` run as a Hive within `cilium-agent` itself.

arighnac:
> MBOI:
> 
> * [bpf: flowtracer cilium#44059](https://github.com/cilium/cilium/pull/44059)

@msune I saw your PRs and it definitely seems that the ideas are aligned (you are proposing the DSCP headers, I am proposing IPv4 options, which we already use for packet tracing in Cilium), although the CFP builds an API and a robust subsytem around generating the synthetic packets? Could you please review this CFP and see if you want any specific areas to be addressed?

arighnac:
@joestringer @jrife Will it be possible for someone to take a look at this proposal?

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
```diff
@@ -0,0 +1,863 @@
...
+* By spawning a native OS process, `cilium-agent` can assign restrictive `cgroups` and lower CPU `nice` priorities to the client, guaranteeing that heavy synthetic tests can never preempt the kernel from executing real networking tasks.  
+* Executing the binary natively within a tightly bounded network namespace cleanly avoids the complexities and multi-threading risks associated with attempting to dynamically enter/exit Linux namespaces (`setns()`) via Go goroutines.
+
+**Cons:**
+
+* **IPC Complexity**: Requires defining and maintaining an internal communication interface (RPC/HTTP) strictly bound to localhost for the agent to negotiate tests with the client.  
+* **Packaging and Lifecycle Overhead**: Necessitates shipping a secondary independent executable within the `cilium-agent` container image and explicitly scripting its lifecycle (start/stop/kill actions) and health checks inside the agent's Go logic.
+
+##### Option 2: Embedded Hive Cell (In-Band)
```
jrife, line 511: I think this should be preferred over spinning up a different process. The cons mostly seem like possible bugs but we could easily avoid those with smart implementation choices. It's more complicated IMO to package and deploy a new binary.

cilium/CFP-46532-synthetic-probing-for-connectivity-test.md
```diff
@@ -0,0 +1,863 @@
...
+
+The `flowgen-client` would be architected in such a way that it doesn’t expect return packets for any protocol (since it can only generate unidirectional synthetic probes). For an example,
+
+* **TCP**: The client creates a `SOCK_STREAM` socket, and attempts to establish a TCP handshake. SYN retries are set to 1, to ensure no retries because packets would be dropped at the server end.
+
+* **UDP**: The client creates a `SOCK_DGRAM` socket. Since UDP is connectionless, the client simply invokes `sendto()`. The kernel prepends the configured IPv4 options to the IP header of every outgoing datagram.
+
+## Section 5: Synthetic Probing: Dataplane
```
jrife, line 536: The only thing I'll say is that this use case fits very neatly into the datapath plugins mold. I wonder how others would feel about having "in-tree" plugins for things like this. To an end user these are normal Cilium feature but the implementation uses the plugin tooling to keep things modular. Maybe @joestringer  could weigh in.

jrife (changes requested):
Leaving some more feedback. Let's get another datapath reviewer to take a look as well, since I'm biased and have already seen this internally.

christarazi:
@cilium/sig-datapath Anyone from the group interested in taking a look?

smagnani96 (commented):
First pass, some thoughts and questions I'll ask here because some of them are related. Apologies if some of them have already been answered or I missed it from the document.

1. I don't want to ping them, but I believe this requires a look also from the Agent and Security teams;
2. Feature OFF by default, turned ON at needs, right?
3. What are we interested into? Just ALL trace/drop events related to this packet? Or do we plan on pushing additional data as in the Marc PoC flowtracer?
4. If so, what changes are needed by the DP? In which programs? And testing in CI? Potentially hitting BPF complexity.
5. Regardless, I think bpf_host needs to be patched anyway to catch the pkt before endpoint redirection (and instead send to our new fictitious program), right?
6. Map key should be __u64, nor __u32 I think
7. Why not reusing the same `trace_id` previously added for user traffic?
8. In case of misrouting, what happens to a flowgen pkt the DP sent it somewhere unintended?
9. How do we prevent malicious actors (or bad users) from abusing of this new feature? Is it possible to generate synthetic traffic on behalf of whoever in the cluster? Who can create CiliumFlowGen (cluster RBAC...)?
10. Regarding crafting the synthetic packet, why do we need socket? Couldn't we do something as we do in BPF program testing (i.e. crafting and injecting the ctx)?
11. Given the IP option, I imagine IPv6 is out of this feature completely, right?
12. `redirect_to_flowgend_dst/cil_to_flowgen` act on bare IP-option presence with no check against an active flowgen_map session, risk of destination-side spoofing?

PR #100, opened 2026-06-21
CFP-26521: Expose URL path in Hubble HTTP metrics

amaanx86:
tl;dr: add an opt-in `path` label to the Hubble HTTP Prometheus metrics, with operator-supplied templates that collapse variable URL segments (IDs, slugs) into stable placeholders.

The original issue was closed in 2023 on cardinality grounds and reopened in 2024 once the discussion moved toward a templated, opt-in design. This CFP writes that design down so implementation work can start. Key points:

- New optional `pathContext` block in the HTTP metric config; works with the legacy `hubble-metrics` config string or the dynamic config map from CFP-30788 (the agent accepts one of the two at a time).
- Templates are matched in declaration order; wildcard tokens are the Go 1.22 `net/http.ServeMux` ones (`/users/{id}`, `/static/{rest...}`).
- The label value is the template string itself, so the set of possible values is fixed at config load: never more than N templates + 1 miss value, deterministic across agent restarts. No runtime cardinality cap; config validation caps the template count instead.
- Miss policy: `onMiss.action` of `emit` (with an operator-picked label value) or `drop_sample`.
- Two small self-observability metrics for the matcher itself.
- Backwards compatible: if `pathContext` is unset, the metric labels are identical to today.

Pulling templates from L7 `CiliumNetworkPolicy` rules (suggested by @apt-itude on the issue) is discussed as an alternative configuration source and deferred to a follow-up CFP.

Cilium issue: https://github.com/cilium/cilium/issues/26521

### DIFF
--- a/hubble/CFP-26521-http-path-metrics.md
+++ b/hubble/CFP-26521-http-path-metrics.md
@@ -1,0 +1,259 @@
+# CFP-26521: Expose URL path in Hubble HTTP metrics
+
+**SIG: SIG-HUBBLE-API/Observability**
+
+**Begin Design Discussion:** 2026-06-21
+
+**Cilium Release:** TBD
+
+**Authors:** Amaan Ul Haq Siddiqui <amaanulhaq.s@outlook.com>
+
+**Status:** Implementable
+
+## Summary
+
+Add an opt-in `path` label to the Hubble HTTP Prometheus metrics. The label value comes from matching the request URL path against operator-supplied templates. The value emitted is the template string itself, so the set of possible label values is fixed by the config, not by traffic.
+
+## Motivation
+
+Hubble already parses the HTTP URL from every L7 flow and shows it in `hubble observe`, but the Prometheus metrics do not carry it. Per-route metrics like "which routes are slow" and "which routes return 5xx" require exporting flow logs and aggregating them somewhere else.
+
+cilium/cilium#26521 was filed in June 2023, closed on cardinality grounds, and reopened in February 2024. Raw paths are effectively unbounded, so a raw `path` label would produce unbounded Prometheus cardinality. Templating bounds it: `/users/abc/orders` and `/users/xyz/orders` collapse into `/users/{id}/orders`.
+
+## Goals
+
+* Add a `path` label to the Hubble HTTP metrics (`hubble_http_requests_total`, `hubble_http_responses_total` in V1, `hubble_http_requests_total` in V2, and `hubble_http_request_duration_seconds` in both).
+* Opt-in only. When the feature is not enabled, metric shapes are unchanged.
+* A distinct-value count for `path` that is fixed by the config alone, deterministic across agent restarts.
+* Works with either the legacy `hubble-metrics` config string or the dynamic config map from CFP-30788. The agent accepts one of the two at a time.
+
+## Non-Goals
+
+* Exposing query strings, fragments, or request bodies.
+* Auto-learning templates from observed traffic. Operators provide the templates.
+* Adding a `path` label to DNS, gRPC, or non-HTTP metrics. Same problem, separate CFP.
+* Replacing the flow-export pipeline.
+* Coupling the feature to L7 `CiliumNetworkPolicy`. This is discussed below as a deferred alternative.
+
+## Proposal
+
+### Overview
+
+Add an optional `pathContext` block to the HTTP metric config. When set, the HTTP handler derives a `path` label by matching the request URL path against the templates. On no match, the configured miss policy fires. The label value is always a template string or the fixed miss value, never anything derived from the request, so the full set of label values is known at config load.
+
+This happens after the L7 parser fills the flow's HTTP `Url` field. No BPF or proxy changes.
+
+### Handler change
+
+`pkg/hubble/metrics/http/handler.go` registers the metrics in `Init`. With `pathContext` set, `path` is appended to the label sets of:
+
+* `hubble_http_requests_total`
+* `hubble_http_responses_total` (V1 only)
+* `hubble_http_request_duration_seconds`
+
+With `pathContext` unset, no label is added. The registered metrics are identical to today.
+
+The match input is the flow's HTTP URL, read via `flow.GetL7().GetHttp().GetUrl()`. With the Envoy proxy, this is a full URL of the form `scheme://host/path`, built from Envoy access-log fields in `pkg/envoy/accesslog.go:ParseURL`. The handler parses it with `net/url` and uses the `Path` component for matching. Query strings and fragments are not used. A URL that fails to parse counts as a miss.
+
+V1 increments `hubble_http_requests_total` on the REQUEST flow, and `hubble_http_responses_total` plus the duration histogram on the RESPONSE flow. The matcher runs once on each. The parser fills `Url` on both, so the templated label value is the same on both sides.
+
+V2 only processes the RESPONSE flow. The matcher runs once.
+
+### Implementation: template matching
+
+Small segment-based matcher. No regex engine in v1.
+
+At config load, each template is compiled once into an ordered list of segment matchers:
+
+* literal segment (exact string match)
+* `{name}` placeholder (exactly one segment, never empty, no slashes)
+* `{name...}` greedy tail (rest of the path including slashes, may be empty, must be the last segment)
+
+The two wildcard forms are the ones Go's `net/http.ServeMux` accepts since Go 1.22, so the syntax is already familiar. A placeholder must span a whole segment (`/b_{bucket}` is invalid), as in ServeMux. Placeholder names are restricted to `[A-Za-z_][A-Za-z0-9_]*`, an ASCII-only subset of ServeMux's Go-identifier rule. ServeMux itself is not reusable as the engine: its matcher is not exported, and most-specific-wins is the wrong semantic for an ordered template list. The matcher here is simpler, and the differences are part of the spec:
+
+* Templates match in declaration order, first match wins. There is no most-specific-wins ranking, so a broad template listed first shadows a narrower one listed later.
+* A template always matches the whole path. A trailing slash is a literal character, not a subtree match, so ServeMux's `{$}` token is not needed and not supported. Subtree capture is written out as `/static/{rest...}`, which matches `/static/` (empty tail) but not `/static`.
+* Segments compare byte for byte. No percent-decoding, no dot-segment normalization. The input is the path as the L7 parser recorded it.
+
+At request time: split the URL path on `/`, walk the compiled templates in declaration order, first match wins. Matching is a linear walk over at most 100 compiled templates per metric. The label value emitted is the template string with placeholders intact, not the request path. `/api/v1/users/abc` and `/api/v1/users/xyz` both produce `path="/api/v1/users/{id}"`.
+
+A malformed template, a duplicate template, an `onMiss.value` equal to one of the template strings, or a template list longer than the cap (see Bounding cardinality) is a config validation error. For the dynamic config map this follows the existing watcher behavior (`pkg/hubble/metrics/metric_config_watcher.go`): the reload is rejected, the last good config stays active, and the error is logged. Today the watcher only logs; the load-errors metric under Self-observability adds a counter for these rejections. For the legacy string, validation fails Hubble metrics initialization, the same as an invalid context option today.
+
+The grammar reserves `:` inside placeholders for a future per-segment regex variant (`{name:regex}`). Adding it later is a parser extension. Existing templates keep working.
+
+### Config
+
+Two ways to configure it, matching how Hubble metric options already work.
+
+Legacy flat string form for `hubble-metrics`. The parser splits on the first `:` to separate the metric name from its options, then on `;` between options, then on `|` between values inside one option (`pkg/hubble/metrics/api/api.go:ParseStaticMetricsConfig`):
+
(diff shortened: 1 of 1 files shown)


xmulligan:
cc: @cilium/sig-hubble-api

hubble/CFP-26521-http-path-metrics.md
```diff
@@ -0,0 +1,176 @@
...
+    maxCardinality: 200
+```
+
+Templates are matched in order, first match wins. `{name}` matches one path segment (no slashes) and gets replaced by the literal template text in the label value. So both `/api/v1/users/abc` and `/api/v1/users/xyz` end up with `path="/api/v1/users/{id}"`. `{name=*}` is a greedy catch-all if you really need it.
+
+`onMiss` says what to do when nothing matches:
+
+* `drop_label` (default): emit the sample with `path=""`. You still count the request, the label just doesn't add a new series.
+* `drop_sample`: don't emit the sample at all. Use this if you only care about templated routes.
+* `bucket_other`: emit with `path="other"`. One visible bucket for everything you didn't template.
```
gandro: Isn't this the same as `drop_label`, but using `path="other"` instead of `path=""`? If so, why have both?

hubble/CFP-26521-http-path-metrics.md
```diff
@@ -0,0 +1,176 @@
...
+
+`maxCardinality` is the ceiling. Once you hit it, new path values fall through to the `onMiss` path. The check is O(1) per request, not a hash set lookup. Default is the template count plus a small overhead for `other`.
```
gandro: To make sure I understand this: The default basically means "no limit" on the path label, right? If I have N templates, it will be impossible to have more then N+1 path labels anyway.

If so, call this out explicitly please.

hubble/CFP-26521-http-path-metrics.md
```diff
@@ -0,0 +1,176 @@
...
+
+If `pathContext` isn't set, the metric labels are identical to today's. Existing dashboards, recording rules, and alerts don't break.
+
+If `pathContext` is set, you get an extra label and you'll need to update your queries. That's the deal you opted into.
+
+`pathContext` is additive in the config schema. Older Hubble agents reading a newer config map ignore the field, same as other additions in CFP-30788.
+
+## Impacts / Key Questions
+
+### Impact: Cardinality
```
gandro, line 206: style nit: This section here in particular is written in a very sassy tone, even more so than the rest of the CFP. If this was an LLM writing it, please tone it down, in particular in this section here.

hubble/CFP-26521-http-path-metrics.md
```diff
@@ -0,0 +1,176 @@
...
+
+This is the whole reason the feature was rejected the first time, so worth being honest about it.
+
+If you write a sensible template list, your label cardinality is bounded by `len(templates) + 1` (the `+1` for `other` if you use it). That's what `maxCardinality` enforces as a hard cap.
+
+If you write `{anything=*}` as your template, you'll get one series per unique URL and Prometheus will hate you. That's an operator mistake, same shape as any other unbounded label, and we'll call it out in the docs. The hard cap stops it from being catastrophic but won't stop it from being annoying.
```
gandro: This paragraph in particular I think should be reformulated. "Prometheus will hate you" in my opinion feels too informal for a design doc. I think readability would be improved if this was formulated using a more matter-of-fact language.

hubble/CFP-26521-http-path-metrics.md
```diff
@@ -0,0 +1,176 @@
...
+
+Ship Option 1 first. It's smaller, doesn't cross subsystems, and works for everyone. Option 2 can come later as a `templateSource: ciliumNetworkPolicy` flag once we've actually used Option 1 for a release and understand what the CNP-to-template mapping should look like.
+
+### Key Question: What's the default when nothing matches
+
+Three sensible answers above (`drop_label`, `drop_sample`, `bucket_other`). Default proposed: `drop_label`. You still count the request, you just don't get a new series for it. Picked this over `bucket_other` because `other` is itself a series that competes with your real ones, and over `drop_sample` because silently dropping samples is a surprise.
+
+### Impact: Helm string config is ugly
+
+The legacy Helm string format with semicolons and commas is already hard to read for the existing options. Adding `pathContext` inline makes it worse. The expectation is that anyone turning on path labels is also on the dynamic config map from CFP-30788, where the YAML form is much nicer. We support the string form for parity but the docs will point people at the YAML one.
```
gandro: Can you show an example how the Helm configuration string will look like?

hubble/CFP-26521-http-path-metrics.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Future Milestones
```
gandro: Since these are stated as non-goals, I think I would leave them as such and not make them future milestones. Listing them under "future milestones" implies that we should do these things, but I'm personally of the opinion that in a particular the first two should not be done. The CNP approach is already discussed in key questions, so nothing is lost by removing it here.

hubble/CFP-26521-http-path-metrics.md
```diff
@@ -0,0 +1,176 @@
...
+metrics:
+- name: http
+  contextOptions:
+    - name: sourceContext
+      values: [pod]
+    - name: destinationContext
+      values: [pod]
+  pathContext:
+    templates:
+      - /api/v1/users/{id}
```
gandro, line 110: What implementation for template matching are you planning to use?

gandro (changes requested):
Thanks for the CFP! I have some feedback, but overall I think this is workable.

Could you disclose the AI influence level for this CFP text please? I do get "LLM vibes" from many formulations, but I might be mistaken. If you've used an LLM, please make sure to follow the AI policy (in particular also make sure to not offload design decisions and communication to LLMs, as mentioned here https://github.com/cilium/community/blob/main/AI-POLICY.md#unacceptable-use)

chancez (commented):
I like it. 

What I would also suggest is writing out how we might support more advanced patterns in the future, to ensure we don't accidentally create a syntax that we cannot extend in the future. 

Currently your proposal starts with single path segment matches and a basic wildcard greedy match which I believe is a very good place to start and I don't think we should try to do more initially.

However, I also know people are going to ask for regex support eventually so it's best to plan how that would fit into your existing config syntax so that we can add it in a backwards compatible way later.

hubble/CFP-26521-http-path-metrics.md
chancez: Yeah, might make sense to have a single option that lets you configure the label for samples that didn't match.

amaanx86:
Thanks for the review @gandro / @kaworu .. fair point on the tone

To be upfront English isn't my first language and I leaned on an LLM pretty heavily for the prose in this draft. The design and technical decisions are mine but the writing clearly shows it. Claiming AIL:2 going forward LLM stays limited to light grammar fixes only.

Will push a revision that covers everything addressed above.

hubble/CFP-26521-http-path-metrics.md
amaanx86: Plan is a small custom matcher. No regex engine for v1.. At config load each template gets pre compiled into an ordered list of segment matchers. At request time the path is split on slash and walked against the compiled forms in order and first match wins. No backtracking. Parse failures get surfaced via the load errors metric. Will add this as a short implementation note in the Proposal.

hubble/CFP-26521-http-path-metrics.md
amaanx86: Will rewrite the line to say so plainly.

hubble/CFP-26521-http-path-metrics.md
amaanx86: Going to fold them into one knob with an action (emit or drop_sample) and a label value the operator picks

hubble/CFP-26521-http-path-metrics.md
amaanx86: Doing a full rewrite pass on the doc to drop the tone. Tracking under the AIL 2 commitment from my top level reply.

hubble/CFP-26521-http-path-metrics.md
amaanx86: Will rewrite.

hubble/CFP-26521-http-path-metrics.md
amaanx86: Agreed... & removing the Future Milestones section

hubble/CFP-26521-http-path-metrics.md
gandro: Thanks for adding some Helm pieces, but the user-facing configuration string is still missing. Can you show examples how configuration looks like with `helm install --set` and `values.yaml` both.

hubble/CFP-26521-http-path-metrics.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Impacts / Key Questions
+
+### Impact: Cardinality
+
+The feature was rejected once before on cardinality grounds. The bound:
+
+With good templates, the `path` label adds at most `N + 1` series per metric with `onMiss.action: emit`, or `N` with `drop_sample`. `maxCardinality` enforces this.
+
+A bad template like `{anything=*}` produces one series per unique URL until `maxCardinality` clamps it. Same class of mistake as any other unbounded label. Docs will call it out. The cap stops runaway growth but does not stop the scrape and storage cost from going up.
```
gandro: I don't understand this. I would expect `{anything=*}` to put everything in the same bucket?

hubble/CFP-26521-http-path-metrics.md
```diff
@@ -0,0 +1,198 @@
...
+      values: [pod]
+  pathContext:
+    templates:
+      - /api/v1/users/{id}
+      - /api/v1/users/{id}/orders
+      - /api/v1/orders/{id}/items
+    onMiss:
+      action: emit          # emit | drop_sample
+      value: ""             # label value when action == emit
+    maxCardinality: 200   # explicit cap; omit to default to N+1
```
gandro: There is one big issue with this: If the argent restarts, the metrics will be reset and the agent doesn't know which labels were emitted before. So this becomes ineffective, since it will emit labels for the first `maxCardinality` hits after restart, which can (and likely will) be different if one has a lot of templates.

I think it might be safer to instead put a hard cap on `templates` (i.e. `N` in the calculation below) for now.

gandro (commented):
Thanks. Some more questions around cardinality and there are still some unresolved comments from before.

hubble/CFP-26521-http-path-metrics.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Implementation: template matching
+
+Small segment-based matcher. No regex engine in v1.
+
+At config load, each template is compiled once into an ordered list of segment matchers:
+
+* literal segment (exact string match)
+* `{name}` placeholder (one segment, no slashes)
+* `{name=*}` greedy tail (rest of the path including slashes, must be the last element)
```
gandro: Alternative syntax: `{NAME...}` to match the Go stdlib behavior. But I have no strong opinion here.

hubble/CFP-26521-http-path-metrics.md
amaanx86: Folded into a single `onMiss` knob in the revision: `action: emit|drop_sample` plus one operator-picked `value`..`drop_label` and `bucket_other` are gone.

hubble/CFP-26521-http-path-metrics.md
amaanx86: Added a Helm examples section: values.yaml and --set for the string form, plus the dynamic configmap form via `hubble.metrics.dynamic.config.content` but one thing I hit while testing: `--set` exits 0 and silently truncates the value at the first unescaped `}` or `,` (helm v4.2.2), so the braces in templates need escaping the same way the docs already escape labelsContext commas. Doc recommends `--set-json` for CLI installs, no escaping needed there.

hubble/CFP-26521-http-path-metrics.md
amaanx86: Written up in the doc now under "Implementation: template matching"

hubble/CFP-26521-http-path-metrics.md
amaanx86: You're right. That line was a leftover from an earlier draft where the label carried the substituted path

hubble/CFP-26521-http-path-metrics.md
amaanx86: Took it

hubble/CFP-26521-http-path-metrics.md
amaanx86: the emitted value is always a template string or the fixed miss value, so the whole value set is fixed at config load and a runtime cap has nothing left to catch. Dropped `maxCardinality` entirely and put a hard cap on the template count instead (100 per metric, a constant for now, enforced at config validation)

gandro (approved):
Thanks, I think this version now looks good to me. Please squash all your commits into one.

I'll also let Chance (and others) chime in before we merge.

amaanx86:
@gandro sure lets wait for others .. and I have squashed the commit

chancez (approved):
LGTM

gandro:
Merging!

Feedback can always be given post-merge too. A CFP can and should be updated even once merged.

PR #101, opened 2026-06-26
CFP-46760: eBPF Runtime Benchmarking

amchelapurath2023:
## Summary

Introduce BPF program execution statistics tracking and telemetry based
on the Linux BPF_ENABLE_STATS syscall. This adds on-demand benchmarking
capabilities via a new 'cilium bpf benchmark' CLI subcommand and routes
cumulative execution counts and runtimes to Prometheus metrics.

## Motivation
Cilium's BPF datapath currently lacks runtime visibility, making it difficult to analyze the performance impact of configuration changes or detect latency regressions in large-scale environments without SSHing into individual nodes to run `bpftool` or similar queries manually. 

[continued in CFP]
Signed-off-by: Anshul Chelapurath <achelapurath@google.com>


amchelapurath2023:
Requesting review @jrife

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
```diff
@@ -0,0 +1,161 @@
...
+    
+    Agent -->|"Exposes metrics on :9962/metrics"| Prometheus["Prometheus Scraper"]
+```
+
+### 1. CLI Hierarchy & Command Flow
+
+This proposal introduces a new CLI command subtree under `cilium-cli`: `cilium bpf benchmark`
+
+#### A. Enabling & Disabling Measurement
+- `cilium bpf benchmark enable --node=<node>`: Patches the targeted agent's configuration options at runtime (`EnableBPFStats=enable`) via the daemon's runtime configuration API. The local agent overrides its configuration on-demand and triggers the BPF_ENABLE_STATS syscall on the host.
```
jrife: An agent flag should suffice to enable/disable metrics collection. [Example](https://github.com/cilium/cilium/blob/68e8b29906de64d4343544057b16171fcc6ae416/pkg/maglev/maglev.go#L63)

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
```diff
@@ -0,0 +1,161 @@
...
+- `cilium bpf benchmark disable --node=<node>`: Patches the targeted agent's configuration options (`EnableBPFStats=disable`) to close the syscall file descriptor and stop BPF statistics collection on the host.
+
+#### B. Tabular Reporting & Aggregation
+The `report` command retrieves metrics on-demand via a pod exec connection to a targeted `cilium-agent` node. The agent queries the kernel and returns the data.
```
jrife: To clarify, a human should `kubectl exec` into the Cilium agent and run `cilium bpf benchmark ...`. The CLI itself wouldn't be running exec under the hood.

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
```diff
@@ -0,0 +1,161 @@
...
+
+```bash
+# cilium bpf benchmark report --node=kind-worker
+NODE         POD                   BPF PROGRAM          TYPE       TOTAL RUNS   TOTAL RUNTIME   AVG LATENCY
+kind-worker  (host:cilium_host)    cil_from_host        SchedCLS   24           0.00s           26368 ns
+kind-worker  endpoint-1625         cil_from_container   SchedCLS   11           0.00s           25790 ns
+kind-worker  (host:eth0)           cil_from_netdev      SchedCLS   439          0.01s           16934 ns
+```
+
+**Supported Options:**
```
jrife, line 110: No `--pod` filter or something like that?

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
```diff
@@ -0,0 +1,161 @@
...
+- `--sort=<field>`: Sort by `avg` (default), `total`, or `runs`.
+- `--prog-type=<type>`: Filter results by BPF program type (e.g. `SchedCLS`).
+- `--save`: Save report outputs to a JSON file (for diff comparison)
```
jrife: It's probably better to just have a `--json` option that just prints to stdout. You can always pipe the output to a file.

jrife (changes requested):

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
amchelapurath2023: I thought for now we were focusing only on node level, but I can add this in easily

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
amchelapurath2023: so then I'm assuming I should remove the --node flag for enable/disable.

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
jrife: Let's add that.

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
jrife: We shouldn't have an `enable`/`disable` command. We should just use Cilium flags to enable/disable the feature.

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
```diff
@@ -0,0 +1,173 @@
+# CFP-46760: Cilium eBPF Runtime Benchmarking
+
+**SIG:** SIG-Datapath, SIG-Observability  
+**Sharing:** Public  
+**Begin Design Discussion:** 2026-06-26  
+**Cilium Release:** 1.19
```
jrife: 1.21?

jrife (approved):

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
mauriciovasquezbernal: `BPF_ENABLE_STATS` has a performance penalty on all ebpf programs running on the host. Do we want to have this enabled for the full run of the cilium agent or allow this to be enabled / disabled at runtime?

amchelapurath2023: The setting of the CLI tool would primarily be in a dev env so majority of the tool's overhead can be disregarded. If we see the tool is being used in a prod env then we can take the gRPC approach and also enable/disable `BPF_ENABLE_STATS` at runtime.

xmulligan:
cc: @cilium/sig-datapath

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
```diff
@@ -0,0 +1,173 @@
...
+## Goals
+
+- **Daemon Configuration**: Enable operators to toggle BPF statistic tracking node-wide via a standard `cilium-agent` command-line flag.
+- **Tabular & Structured Reporting**: Expose BPF runtime statistics in a tabular format within the CLI for easy interpretation.
+- **Prometheus Telemetry Integration**: Periodically scrape and expose cumulative run counts and execution times on the `cilium-agent` BPF metrics endpoint.
+- **Local Performance Comparison (Diff)**: Compare BPF runtime statistics from local test runs against baseline configurations to catch performance changes during development.
+
+## Non-Goals
+
+- **Granular Tail-Call Latency Isolation**: The stats timer hooks in the Linux kernel only run pre-invocation of the entry program and post-completion of the final exit return. Isolating individual tail-called program latencies within the aggregated metric is out of scope.
```
dylandreimerink, line 35: Its not clear what the "aggregated metric" includes. Does this mean that it does include the tail calls for each entrypoint just that they are not individually listed? Or does it mean that we do not record metrics for tail calls at all?

We should definitely record metrics for the tail calls and attribute them to the entrypoint, otherwise you will not get useful metrics.

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
```diff
@@ -0,0 +1,173 @@
...
+    end
+
+    %% Telemetry Export
+    Agent --->|"Exposes metrics on :9962/metrics"| Prometheus["Prometheus Scraper"]
+```
+
+### 1. CLI Hierarchy & Command Flow
+
+This proposal introduces a new CLI command subtree under `cilium`: `cilium bpf benchmark`
+
```
dylandreimerink: May I suggest adding a hive shell command for this instead. https://docs.cilium.io/en/stable/contributing/development/hive/#testing-with-hive-script

This would make the command available under `cilium shell`. You can still add a `cilium bpf benchmark` alias which would then call the script command as its implementation see https://github.com/cilium/cilium/blob/main/cilium-dbg/cmd/metrics_list.go.

This would also allow us to use the scripttests in the future to assert performance numbers. In something like https://github.com/cilium/cilium/issues/27450.

It also eliminates your `Design Tradeoffs: Pod Exec vs. gRPC API` design tradeoff since shell commands run directly in the cilium agent process and the stdin/out is simply transferred over unix socket to cilium-dbg. So no need for re-inventing anything there.

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
amchelapurath2023: Thanks for the comments @dylandreimerink. 

1. The tail calls are included in the runtime of the total program, but specific runtime of the tail call itself is not visible due to the kernel's limitations (ex: prog a which takes 3 seconds makes a tail call to prog b which takes 2 seconds. The benchmark will show that prog a took 5 seconds in total without listing prog b).

2. Recording tail call metrics are out-of-scope as of right now, however we still think that the tool will be useful for our cases. I can update the future milestones to include tail call metrics, possibly with the use of [datapath plugins](https://github.com/cilium/design-cfps/pull/76).

dylandreimerink: > The tail calls are included in the runtime of the total program, but specific runtime of the tail call itself is not visible due to the kernel's limitations (ex: prog a which takes 3 seconds makes a tail call to prog b which takes 2 seconds. The benchmark will show that prog a took 5 seconds in total without listing prog b).

Ah, that is an interesting detail about how the measurement works. That is good, since its what we want (I think). Thanks!

dylandreimerink (approved):

pchaigno (approved):
This looks good to me. Thanks for the write up!

I'd be interested to know how much overhead enabling this adds. If it is insignificant (which I guess it isn't), we may consider enabling it by default. Knowing the overhead will also be key to document it properly for users.

cilium/CFP-46760-eBPF-Runtime-Benchmarking.md
```diff
@@ -0,0 +1,202 @@
+# CFP-46760: Cilium eBPF Runtime Benchmarking
+
+**SIG:** SIG-Datapath, SIG-Observability  
+**Sharing:** Public  
+**Begin Design Discussion:** 2026-06-26  
+**Cilium Release:** 1.21  
+**Authors:** Anshul Chelapurath <achelapurath@google.com>  
+**Status:** Draft
```
pchaigno: @xmulligan This CFP looks ready to go. Is all we need to change this line to `Implementable` and merge?

xmulligan: Yes I would just update that line and since it has approval from at least one committer it is ready to merge 🙂

amchelapurath2023: Made some nit changes based on feedback from https://github.com/cilium/cilium/pull/47186. Thanks for the reviews everyone!

amchelapurath2023:
> This looks good to me. Thanks for the write up!
> 
> I'd be interested to know how much overhead enabling this adds. If it is insignificant (which I guess it isn't), we may consider enabling it by default. Knowing the overhead will also be key to document it properly for users.

@pchaigno 
Ran a small experiment to get some numbers

Setup: Two single-node k3s clusters on isolated Linux 6.8 arm64 VMs (Apple Silicon). One cluster runs Cilium --enable-bpf-stats=true, the other without. iperf3 run pod-to-pod.

Kernel overhead: Enabling BPF stats adds two ktime_get_ns() calls per BPF program invocation which is about 20 ns each on bare metal, probably slightly higher in a VM. Across three 10-second TCP runs at ~26 Gbits/sec. Throughput was statistically identical between the stats and baseline clusters.

Wall time of Prometheus scrape (time curl -s http://$CILIUM_POD_IP:9962/metrics > /dev/null): 13 ms per scrape

Wall time of CLI (time kubectl exec -n kube-system $CILIUM_POD -- cilium-dbg bpf stats report > /dev/null): 170ms

The wall time for CLI is much higher most likely kubectl's exec overhead.

pchaigno:
@amchelapurath2023 Interesting. Maybe we could enable it by default after letting it stabilize for one release.

PR #102, opened 2026-07-10
CFP-47089: Gateway API Session Persistence

thorn3r:
This proposal extends Cilium's Gateway API implementation to support session persistence via the Route Rule API as defined in [GEP-1619][1].

[1]: https://gateway-api.sigs.k8s.io/geps/gep-1619/#achieving-session-persistence

### DIFF
--- a/cilium/CFP-47089-gateway-api-session-persistence.md
+++ b/cilium/CFP-47089-gateway-api-session-persistence.md
@@ -1,0 +1,137 @@
+# CFP-47089: Gateway API Session Persistence
+
+## Meta
+
+SIG: SIG-ServiceMesh
+Sharing: Public
+Begin Design Discussion: July 10, 2026
+End Design Discussion: TBD
+Cilium Release: 1.20
+Authors: tim <thorn3r@cisco.com>
+
+## Summary
+
+This proposal extends Cilium's Gateway API implementation to support session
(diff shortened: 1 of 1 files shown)


xmulligan:
cc: @cilium/sig-servicemesh

youngnick:
As @gcs278 points out on cilium/cilium#47089, I totally missed that the HTTPRoute fields are also being deprecated and will be removed. I think we need to hold on this one until there's some more movement in upstream on the Backend proposal (see kubernetes-sigs/gateway-api#4876 for the details).

thorn3r:
I'll mark this as a draft and revise it when the Backend API is ready

thorn3r:
Had a chat offline with @youngnick. Since the timeline on the Backend API implementation is not clear, we won't hold this feature based on its availability. This CFP is open for review again

youngnick (approved):
Thanks @thorn3r, this LGTM for this phase of development.

joestringer:
If this is considered "Implementable" at this stage, I think we can add that to the metadata at the top of the document and merge?

thorn3r:
It's actually already been implemented and merged: https://github.com/cilium/cilium/pull/48029.

Seems like I should've merged the CFP after Nick's approval. I'll merge this now and keep in mind for future submissions

PR #103, opened 2026-08-21
CFP-48102: Cilium Gateway API TLS Options

arybolovlev:
This proposal adds Cilium Gateway API TLS Options for `Gateway` and `ListenerSet` listeners that terminate TLS connections.

Cilium issue: https://github.com/cilium/cilium/issues/48102.


arybolovlev:
cc: @youngnick

cilium/CFP-48102-gateway-api-tls-options.md
```diff
@@ -0,0 +1,175 @@
...
+ 
+Failures that depend on the certificate, Envoy build, or runtime environment are left to Envoy and surfaced through the existing listener programming error path.
+ 
+Cilium will reject configurations where:
+ 
+- A supported option is configured on a listener where Cilium does not terminate downstream TLS.
+- A TLS version is not supported by the implementation.
+- The minimum TLS version is greater than the maximum TLS version.
+- A list-valued option is empty or contains empty entries.
+- An unsupported Cilium TLS option key is present.
```
youngnick, line 115: We're probably going to have to validate the cipher suites instead, depending on the Envoy behavior if it gets an invalid cipher suite.

If Envoy will NACK a change that includes an invalid cipher suite, then we absolutely must validate all the comma-separated values in `cilium.io/tls-cipher-suites`, but we probably should anyway.

(ideally we will important some constant or something from Envoy or go-control-plane that keeps a list for us, rather than having to maintain one).

cilium/CFP-48102-gateway-api-tls-options.md
```diff
@@ -0,0 +1,175 @@
...
+ 
+Invalid Cilium TLS options will cause Cilium to reject only the affected listener. Other valid listeners on the same `Gateway` or `ListenerSet` will continue to be programmed.
+ 
+### Key Question: How should listener TLS option failures affect parent status?
+ 
+This CFP defines listener-level behavior: invalid listener-scoped Cilium TLS options reject only the affected listener. Other valid listeners on the same `Gateway` or `ListenerSet` continue to be programmed.
+ 
+The implementation still needs to decide how these partial listener failures are reflected in parent `Gateway` and `ListenerSet` status conditions.
+ 
+One option is to preserve existing aggregate parent status behavior and rely on listener conditions for the precise failure. Another option is to surface partial listener failures in parent-level conditions so operators can detect the issue without inspecting every listener.
```
youngnick: This is covered by the general Listener validity rules - if there are other valid Listeners, then the Gateway overall can be `Accepted` and `Programmed`. If the invalid listener is the only one, then it must not be `Programmed`.

youngnick (approved):
A couple of minor non-blocking nits, but this CFP LGTM aside from those.

cilium/CFP-48102-gateway-api-tls-options.md
arybolovlev: You are right. I have removed this question from the list. Thanks!

cilium/CFP-48102-gateway-api-tls-options.md
arybolovlev: Good point. Since these values are supplied as free-form text, the API schema cannot fully validate them. We will need to add explicit validation, so I have added another bullet to call this out.

Thank you!

PR #104, opened 2026-08-21
CFP-48090: Multi-Pool Resource IPAM for the Cilium Network Driver

pippolo84:
CfP for the Cilium Network Driver dedicated IPAM mode (Multi-Pool Resource IPAM)

https://github.com/cilium/cilium/issues/48090


cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,529 @@
...
+without the operator.
+
+The driver records allocated addresses in two places in
+`ResourceClaim.status.devices`:
+
+- The driver's opaque `data` contains the selected Resource IPAM pool and the
+  complete serialized device configuration needed for restore and release.
+- Standard `networkData.ips` exposes the addresses to Kubernetes and users.
+
+Resource IPAM returns host prefixes (`/32` for IPv4 and `/128` for IPv6) by
```
pippolo84: Returning host prefixes is something that derives from the existing Multi Pool Pod IPAM.

We could change this behavior adding a field in `CiliumResourceIPPool` to specify a netmask that should be used when configuring the interface.
Compared to leaving this setting to a specialized "network config" CRD, it has the advantage of keeping the address and netmask specs in the same place, but it forces all the addresses allocated from the pool to use the same netmask.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,529 @@
...
+or deleting a pool stops the operator from delegating new CIDRs from that
+capacity. CIDRs already delegated to nodes remain reserved until the agents can
+release them safely, and their remaining local capacity may still satisfy
+allocations during that transition.
+
+The pool name scopes uniqueness. Cilium guarantees that two live allocations
+from the same `CiliumResourceIPPool` do not receive the same address. Different
+pools are independent allocation domains and may intentionally use overlapping
+address space when they represent isolated networks. Cluster operators are
+responsible for avoiding overlap between pools, including Pod IP pools, that
```
bersoare: i would rewrite the last sentence to something along the lines of "it is the responsibility of the operator to ensure there is alignment between the pool addressing and the underlying network infrastructure to avoid undesired side effects"

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,529 @@
...
+request/allocation handshake used by Multi-Pool Pod IPAM:
+
+```yaml
+apiVersion: cilium.io/v2
+kind: CiliumNode
+metadata:
+  name: worker-1
+spec:
+  ipam:
+    resourcepools:
```
bersoare: shouldnt this be `resourcePools`? (camel case)

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,529 @@
...
+The former selects an address allocation domain; the latter identifies
+the device pool from which Kubernetes allocated the hardware resource.
+
+There is no implicit default and no fallback to another Resource IPAM pool. If
+`ip-pool` is omitted, Resource IPAM performs no allocation for that request. If
+the named pool does not exist, does not provide a required address family, or
+is exhausted, claim preparation fails with a retryable error rather than
+configuring the device for a different network.
+
+In the initial implementation, the requested address families follow the
```
bersoare: since these new ipam pools may represent a different network topology than what is effectively used for the control/management plane (eth0 device), i think we should not have the address families in CNI and network driver tied together. 
for example, an operator might be running cilium ipv4 only in the cluster over which they manage the deployments and collect telemetry, while at the same time providing dual stack services over a dedicated fabric with the network driver.
tying the two planes together like this can force users into adding more complexity to their cluster than they need

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,529 @@
...
+The driver records allocated addresses in two places in
+`ResourceClaim.status.devices`:
+
+- The driver's opaque `data` contains the selected Resource IPAM pool and the
+  complete serialized device configuration needed for restore and release.
+- Standard `networkData.ips` exposes the addresses to Kubernetes and users.
+
+Resource IPAM returns host prefixes (`/32` for IPv4 and `/128` for IPv6) by
+default. The `spec.ipv4.maskSize` and `spec.ipv6.maskSize` values are not used
+as interface prefixes. Subnet prefix, routes, gateways, and other connectivity
```
bersoare: implicitly, by not dealing with any of the routing details here, the IPAM proposal in itself does not ensure connectivity will be possible between pods attached to the same pool. is this intentional? if so, we should probably write it down to make it more obvious

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,529 @@
...
+
+### Demand and preallocation
+
+For each pool and family, the agent computes demand from:
+
+- addresses currently in use by successfully prepared resources;
+- allocations currently pending because local capacity is exhausted; and
+- an optional per-pool preallocation buffer.
+
+The preallocation buffer allows frequently used pools to avoid a control-plane
```
bersoare: how is this buffer determined, and how does it change the "un-buffered" allocation process?
on the other hand, this may look like a premature optimization since the rationale for it is too vague
we could add some detail here (or maybe drop this part altogether)?

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,529 @@
...
+
+Multi-Pool Resource IPAM is available only when the Cilium Network Driver is
+enabled. It does not require `ipam.mode=multi-pool`; the cluster may use any
+supported Pod IPAM mode.
+
+Two optional configuration surfaces mirror Multi-Pool Pod IPAM:
+
+- The operator can auto-create pools at startup with
+  `--auto-create-cilium-resource-ip-pools`.
+- The agent can configure warm capacity per pool with
```
bersoare, line 465: to me it sounds like this toggle is intended to reduce the allocation churn by anticipating demand on the agents themselves.
maybe worth adding here the implications of using the toggle, what valid values are, when to use/not use it?

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,529 @@
...
+  `--resource-ipam-multi-pool-pre-allocation=<pool>=<addresses>`.
+
+Pools can also be created and managed directly as Kubernetes resources. The
+operator and agent receive the RBAC permissions required to observe the new
+resource and reconcile the corresponding `CiliumNode` state. The CRD is
+installed with the Cilium chart.
+
```
bersoare, line 474: i wonder if we should also consider having a way to expose the allocations to the pod itself (through environment vars or something like that)

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,529 @@
...
+### Observability
```
bersoare, line 475: statedb tables too?

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: Clarified how the buffer is handled by the implementation and how it should be set by the cluster operator. Also added a practical example to give a sense of how much underutilization of the global pool may result from a preallocation value.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: Might be interesting, but I wouldn't commit to that in the CfP. I think it is better to wait for practical uses and then evaluate the best way to expose the addresses.
Besides, there isn't anything in this proposal that might limit our future options in that regard.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: It's a good idea, but I'm unsure if we should commit to that in the CfP.

Exposing a "read-only" per-driver table of allocated addresses is easy enough and we can open a follow up PR to add that: it's an addition without any kind of trade off needed.

It also goes in the direction of what we should do for Pod IPAM (see https://github.com/cilium/cilium/issues/40198). But since we are already reasoning about that at a more general IPAM level, I'd like to first explore if a "centralized" solution might be more convenient (e.g: have something already in place for pod IPAM and reuse the harness/structure for DRA resources, similarly to what I'm proposing here for the multi pool manager)

Update: see comment https://github.com/cilium/design-cfps/pull/104#discussion_r3907581918

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: Added a reference to the preallocation example to clarify intended usage

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: Might be worth adding options in the CiliumResourceIPPool object to also set up a minimal network config that makes the resources usable. It also solves my concern in https://github.com/cilium/design-cfps/pull/104/changes#r3831538891.

Let me update the CfP with an extended version of the CRD.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,577 @@
...
+The following Kubernetes objects expose the allocation path without recording
+every IP allocation in `CiliumNode`:
+
+- `CiliumResourceIPPool` shows the configured cluster-wide capacity and node
+  allocation granularity.
+- `CiliumNode.spec.ipam.resourcePools` shows per-node demand and delegated
+  CIDRs.
+- `ResourceClaim.status.devices[].networkData.ips` shows addresses allocated to
+  a prepared device.
+
```
bersoare, line 490: i'd also add any IPAM allocations to the statedb table (`networkdriver-dra-devices`)

bersoare (approved):
one last comment, but LGTM overall! thank you

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: Oh, right, I was focused on the feature branch implementation and forgot we already have a stateDB table to manage allocations from https://github.com/cilium/cilium/pull/47558. I'll update the CfP accordingly :+1:

pippolo84:
> one last comment, but LGTM overall! thank you

Uploaded an updated version where I mentioned the stateDB table for observability and proposed an optional section in CiliumResourceIPPool for basic network configuration (interface prefix and static routes).

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: Added an optional field `networkConfig` in CiliumResourceIPPool to allow for basic configuration of DRA resources alongside the address allocation.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,652 @@
...
+also requires removing its network configuration in the same update or an
+earlier update.
+
+The Network Driver watches `CiliumResourceIPPool` objects and uses its current
+view of the selected pool during resource preparation. After allocating an
+address, it finds the network configuration for the pool CIDR containing that
+address and applies the configured prefix length and routes to the claimed
+resource. If the corresponding pool CIDR has no `networkConfig` entry, the
+address uses a host prefix (`/32` for IPv4 or `/128` for IPv6) and no routes are
+installed from the pool.
```
networkop, line 212: I wonder what's the rationale behind this decision as opposed to using the prefix of the spec.ipv4.cidrs.
if these interfaces are plugged into the physical network, host prefix address does not give them any way to reach any other IP on the subnet,

pippolo84: This comes directly from the way the addresses are assigned for pods in Cilium, where we also add a link-scoped route to the gateway.
Here we don't have that, so yes, maybe using the mask from the prefixes make more sense as a default. In any other case, it is possible to specify a custom mask in networkConfig.prefixLength.

pippolo84: Updated the proposal to derive the default prefix from the pool CIDR.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,652 @@
...
+Keeping basic network configuration in `CiliumResourceIPPool` associates an
+allocated address, its interface prefix length, and its static routes in one
+administrative object. It also lets the Network Driver obtain the configuration
+from the pool it already watches, without coordinating a second resource lookup
+or allowing unrelated network configuration to select the address semantics.
+
+The trade-off is that every address allocated from a configured CIDR receives
+the same basic network configuration on every node. This API does not represent
+topology-specific gateways or routes, VLANs, sysctls, or device-specific
+settings. Such configuration requires a future Network Driver API rather than
```
gandro, line 562: Personally, I would prefer it if we moved the `networkConfig` to such a PR. Otherwise the danger is there that we start to overload the `CiliumResourceIPPool` CR.

Also, I'm of course motivated by the fact that I would like to use the `CiliumResourceIPPool` outside of the Network Driver purpose, but that might be an explicit Non-Goal.

dylandreimerink (approved):
Great CFP. It all seems clear and well thought out to me

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: I agree. Updated the CfP with a separate `CiliumResourceNetworkConfig` to offer basic network configuration without polluting the CiliumResourceIPPool CRD.

bersoare: no concerns from my end

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,720 @@
...
+to network settings do not affect pool capacity or allocation state.
+
+```yaml
+apiVersion: cilium.io/v2alpha1
+kind: CiliumResourceIPPool
+metadata:
+  name: blue-network
+spec:
+  ipv4:
+    cidrs:
```
gandro, line 152: Is there any reason we're not using the v2 `CiliumPodIPPool` schema (with `pool:`) for the CIDRs? https://github.com/cilium/cilium/pull/44383

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,720 @@
...
+and no static routes are installed. For example, if the selected pool contains
+`10.20.0.0/24`, entries for either `10.20.0.0/25` or `10.20.0.0/16` do not
+match; allocating `10.20.0.100` assigns `10.20.0.100/24` to the DRA interface
+without additional routes. A matching entry that omits `prefixLength` also
+retains this default, and an omitted `routes` list installs no static routes.
+
+Network configuration is applied only at preparation time. Updating or deleting
+a `CiliumResourceNetworkConfig` has no effect on resources prepared with its
+previous contents. The resolved configuration is stored with the prepared
+device so that restart recovery and unprepare do not depend on the current
```
gandro, line 265: What about node restart? It's not a problem because we persist things in the resource claim?

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,720 @@
...
+`prefixLength` specified by the exactly matching `CiliumResourceNetworkConfig`
+entry overrides this default. The pool's `maskSize` is not considered when
+choosing the interface prefix.
+
+The Network Driver watches both `CiliumResourceIPPool` and
+`CiliumResourceNetworkConfig` objects. During resource preparation, it finds
+the entry whose `cidr` exactly matches the pool CIDR containing the allocated
+address and applies its routes. If the matching entry specifies a
+`prefixLength`, that value overrides the default derived from the pool CIDR.
+The match is based on CIDR equality, not address containment. A network
```
gandro, line 250: What's the intent with correlating the IPAM pool definition with the network configuration? It seems the only reason to correlate the network config with the IPAM pool seems to be to be able to fallback to a default `prefixLength` if there is no configuration. But is that really worth it?

I think there is a lot to be gained if the two are more independent. If I have a lot of small CIDRs in my IPAM pool (for whatever reason, e.g. because I have an already fragmented IP space), with the current proposal I have to repeat the routes quite often. For example, I could totally see one wanting define a IPAM pool containing a bunch of prefixes like `10.10.0.0/24`, `10.20.0.0/24`, `10.30.0.0/24`, but want to share the same network configuration for all these routes. Rather than duplicating the routes and prefix section for each pool CIDR, I think it would be much nicer to to just have a section for `10.0.0.0/8` that contains the routes for those three IPAM CIDRs.

To me, it would feel much more natural if the network config is selected simply based on the assigned IP on the device. The selection of network config (given the device IP) is then done by 
1. checking if there is a cidr in `CiliumResourceNetworkConfig` that contains that device IP, or if none match
2. checking if there is a cidr in the `CiliumResourceIPPool`

The nice thing about that is that in the future you might even support `CiliumResourceNetworkConfig` for statically allocated IPs (if we ever want to support that) - i.e. having a resource claim that only has a `network-config`, without an `ip-pool`.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: That's correct, IP addresses are stored in `ResourceClaim` Status (specifically [here](https://github.com/kubernetes/kubernetes/blob/c7c3571e83e48b1edd1a0e50ac229654aa7c50fd/pkg/apis/resource/types.go#L2240)). They are the only parameters that we need to "reallocate" after a restart.
Other settings, like routes, are relevant only when the NRI hook `RunPodSandbox` is called.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: Isn't `pool:` used in `CiliumPodIPPool` only used to set additional per-CIDR configuration? With the updated proposal (moving all additional network configuration in a separate CRD) that shouldn't be needed in CiliumResourceIPPool.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: I like the idea very much. Only one concern: how should we manage possible conflicts? Specifically:

1) In case of overlapping CIDRs from different `CiliumResourceNetworkConfig` -> Here I think we could select the narrower CIDR. This will also allow to define common network configuration for broader CIDRs and then add specific settings for narrower CIDRs contained in the broad ones.
- In case of equal CIDRs from different `CiliumResourceNetworkConfig` -> I think a simple solution is to mark both `CiliumResourceNetworkConfig` as conflicting and simply disregard both of them until the conflict is fixed. We could expose to the user that a `CiliumResourceNetworkConfig` is conflicting reporting an error the resource Status.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
gandro: > Other settings, like routes, are relevant only when the NRI hook RunPodSandbox is called.

Pardon my ignorance - that is called again on node restart, right? So if routes change, they will be updated after node restart? Or are they also persisted in the resource claim?

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
gandro: 1. Yes, I think longest-prefix-match is probably the most intuitive way to go.
2. Under what circumstances would two `CiliumResourceNetworkConfig` apply to the same device? Is this if two pods are attached to the same host device, but select a different `network-config` in their claim? Yes, I think rejecting conflicting CIDRs is probably the cleanest. Alternatively, a deterministic merge algorithm would probably also work (i.e. sort network configs by name and later configs overwrite earlier configs), but might be a bit more surprising.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
gandro: Well, it allows for reserving ranges, which is purely an IP allocation concept and doesn't influence routing. So you can say "allocate from `10.10.0.0/16`, except for the range `10.10.0.1-10.10.0.10`. Might be useful for resources too.

pippolo84: Sorry, I was focusing on the network configuration here.
About the `reservedRanges` under `pool` it's explained in the section "Coexistence with statically managed addresses", but I'll add the field here too and also expand that section with a CiliumResourceIPPool example and a detailed explanation about how to reserve ranges. :+1:

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: > that is called again on node restart, right? So if routes change, they will be updated after node restart? Or are they also persisted in the resource claim?

First of all, sorry, I read "agent restart". About a node restart:

- the Network Driver does not register itself as NRI plugin and DRA plugin until it has restored all previous allocations
- the allocations state is persisted in k8s (ResourceClaim Status), including the routes (specifically, they are stored as opaque data in `ResourceClaim.status.devices[].data`, while IP addresses are also available [here](https://github.com/kubernetes/kubernetes/blob/c7c3571e83e48b1edd1a0e50ac229654aa7c50fd/pkg/apis/resource/types.go#L2240) for observability reasons). Therefore, at startup the driver restores them in memory.
- K8s replays the DRA preparation hook (PrepareResourceClaims) for each allocated device, but the Driver implementation is idempotent, so that has no effect
- When the pods are recreated, the NRI hook `RunPodSandbox` is called and each allocated device is prepared using the stored configuration from the previous run

Therefore, in case of a change in the CiliumResourceNetworkConfig, the configuration of allocated devices doesn't change, not even in case of a node restart.

I'll rephrase a bit the CfP, since `applied only at preparation time` does not tell the full picture in case of a node restart.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: After reading again the comment I realized that the suggestion was not to remove `network-config` selector from the ResourceClaim/ResourceClaimTemplate. Therefore, no conflict can happen, since each allocated device selects a single CiliumResourceNetworkConfig. I'll update the CfP to follow this approach. :+1:

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
```diff
@@ -0,0 +1,778 @@
...
+specific entry using longest-prefix matching. The selected entry's
+`prefixLength`, when present, is applied to the address, and its routes are
+installed for the claimed resource. For example, the `10.20.0.0/16` entry
+above applies to addresses allocated from both `10.20.0.0/24` and
+`10.20.1.0/24` in the example pool.
+
+If no network configuration is selected, or the selected configuration has no
+entry containing the allocated address, the Network Driver uses the prefix
+length of the pool CIDR containing that address and installs no static routes.
+A matching entry that omits `prefixLength` also uses this default, while still
```
gandro: This is very surprising to me. I would find it much more intuitive that a network config CIDR entry also implicitly defines the prefix length, if no explicit `prefixLength` is specified. Personally I feel like that is how most people will think about subnets, and I anticipate that an explicit `prefixLength` is will be rather a corner-case.

So if I have an CIDR entry in `CiliumResourceNetworkConfig` for `10.0.0.0/16` with only routes and no prefix length, I would expect it to set `/16` as the network mask, independently of whatever the IPAM CIDR was.

gandro (commented):
Neat, thanks! I really like the direction. One thing that stood out to me.

cilium/CFP-48090-cilium-network-driver-multi-pool-resource-ipam.md
pippolo84: Yeah, makes sense: if a `CiliumResourceNetworkConfig` is added to "describe" a CIDR, then it's fair to assume that all the allocated addresses part of that should have that prefix length.

gandro (approved):

PR #106, opened 2026-09-18
CFP-44138: Cilium Clusterwide Local Redirect Policy

ajmmm:
CFP-44138: Cilium Clusterwide Local Redirect Policy

https://github.com/cilium/cilium/issues/44138


cilium/CFP-44138-cilium-clusterwide-local-redirect-policy.md
```diff
@@ -0,0 +1,504 @@
...
+CCLRP should take precedence over CLRP when both resources identify the same frontend, so that an explicitly migrated CCLRP is not shadowed by the legacy policy while both resources coexist.
+
+The precise status representation is an implementation detail for the initial release, but validation and reconciliation failures must be visible through agent diagnostics, logs, and events or equivalent Kubernetes-facing feedback. A full status API is a possible follow-up milestone.
+
+### Observability
+
+The existing local redirect debugging commands should be extended to identify:
+
+* whether an entry came from CLRP or CCLRP;  
+* the policy that owns the entry;
```
ajmmm, line 273: Self review:
- move below `the resolved frontend address and port`
- rephrase to clearly express it should be clear which policy has overridden which frontend

cilium/CFP-44138-cilium-clusterwide-local-redirect-policy.md
```diff
@@ -0,0 +1,504 @@
...
+
+CCLRP is intentionally cluster-scoped. Creating or modifying one requires cluster-level authorisation, and the resulting redirect can affect traffic from workloads in any namespace. This is a larger authorisation boundary than the current resource presents, but it accurately reflects the effect of the feature.
+
+The documentation and RBAC examples must make this explicit. Namespace-specific ownership or delegation is outside the initial proposal.
+
+### Key Question: how should backend namespace scope be expressed?
+
+The examples above make use of the `io.kubernetes.pod.namespace` label in `localEndpointSelector` to align with the selector conventions of `CiliumClusterwideNetworkPolicy`. However, this creates a safety foot gun for local redirects: omitting that label broadens the selector to pods in every namespace, which can cause unintended pods to become redirect backends.
+
+This is a more serious failure mode than an overly broad policy selector because the selected pods change the load-balancer destination for matching traffic. The API should make namespace scope explicit and should not silently default to all namespaces as a consequence of a missing label.
```
ajmmm, line 391: Self review: the question shouldn't state functional requirement. It should be moved below into the option as a 'could'.

cilium/CFP-44138-cilium-clusterwide-local-redirect-policy.md
```diff
@@ -0,0 +1,504 @@
...
+*Note: `skipRedirectFromBackend` field will still remain opt-in for CLRP.*
+
+The safety of this change is not guaranteed. The original implementation of CLRP was augmented with loop prevention via socket lookups (`bpf_sk_lookup_*` helpers) in October 2020. This appears to have suffered from incorrect behaviour under certain scenarios and was replaced with the current skip-map implementation in April 2024.
+
+At the time of the original augmentation, queries were raised about any known use cases that may require a request from an LRP backend to be sent back to that same backend. The original contributor confirmed there were no known use cases.
+
+At the time of writing, when reviewing this feature within the context of the aforementioned use cases above, it is believed there are still no known use cases that would require a request from an LRP backend to be sent back to that same backend.
+
+Original socket lookup implementation: [https://github.com/cilium/cilium/pull/13287](https://github.com/cilium/cilium/pull/13287)  
+Revised "skip-map" implementation: [https://github.com/cilium/cilium/pull/26144](https://github.com/cilium/cilium/pull/26144)
```
ysksuzuki, line 250: Thanks for adding the historical context. My understanding is as follows:

- [#13287](https://github.com/cilium/cilium/pull/13287) introduced loop prevention as the automatic/default behavior, where supported by the kernel.
- During its review, [a reviewer asked](https://github.com/cilium/cilium/pull/13287#pullrequestreview-508629840) whether there could be a valid use case in which traffic from an LRP backend should be redirected back to the same backend. The response was: [“It seems unlikely. Currently we don't have such a use case.”](https://github.com/cilium/cilium/pull/13287#issuecomment-708611476)
- The problem with that implementation was its socket-lookup-based detection. A listening socket matching the port could incorrectly cause translation to be skipped, as reported in [#15195](https://github.com/cilium/cilium/issues/15195) and [#20164](https://github.com/cilium/cilium/issues/20164).
- [#26144](https://github.com/cilium/cilium/pull/26144) replaced that implementation with the netns-cookie-based skip maps and introduced `skipRedirectFromBackend`. Its commits acknowledge that loop prevention [had previously been enforced by default](https://github.com/cilium/cilium/commit/e8974f31fa193e196ece191b626124f8672cf748), but do not explain why the default behavior was changed or identify a use case requiring `skipRedirectFromBackend=false`.

My inference is that making the new mechanism opt-in may have been motivated by its implementation constraints, rather than by a requirement to support both behaviors. It introduced a kernel requirement and additional BPF map management. The implementation review also noted that [only a small number of map entries was expected (https://github.com/cilium/cilium/pull/26144#discussion_r1567671727), and the maps were [initialized only when a policy requested the feature](https://github.com/cilium/cilium/pull/26144#discussion_r1571376821). This is only an inference, since the rationale was not documented explicitly.

Given this history, I do not see evidence of a use case that requires `skipRedirectFromBackend=false`. Therefore, making loop prevention the default and only behavior for CCLRP appears reasonable. I think the CFP should state this conclusion and its rationale directly, instead of saying that the safety of the change is not guaranteed.

However, the resource consumption of the skip maps should still be evaluated. The current IPv4 and IPv6 skip maps each have [a maximum of 100 entries](https://github.com/cilium/cilium/blob/debffba0a964b463d90855ac97a911fd672af6ef/pkg/loadbalancer/maps/types.go#L1266-L1275). Entries are keyed by the backend network namespace and frontend address/port. Could the CFP evaluate the expected skip-map usage for the intended CCLRP use cases and whether the current maximum of 100 entries needs to be revisited?

cilium/CFP-44138-cilium-clusterwide-local-redirect-policy.md
```diff
@@ -0,0 +1,504 @@
...
+
+CCLRP should express the relationship directly as `port` to `targetPort`, following the general Kubernetes Service model. Port names should not be required to connect the frontend and backend sides of a redirect.
+
+### Loop prevention exposes datapath details
+
+The `skipRedirectFromBackend` field exposes an implementation detail of the socket load balancer and requires users to make the same runtime decision independently for every policy. However, there is no known useful purpose for redirecting a backend's traffic back to itself.
+
+Loop prevention should be an agent-level behaviour, enabled when the required kernel capability is available and reported when it is not.
+
+## Goals
```
ysksuzuki, line 101: I think zero-downtime migration from CLRP to CCLRP should be an explicit goal of this CFP. This is necessary to support migration in large clusters that already depend on CLRP. For example, when CLRP is used for NodeLocal DNS, even temporarily withdrawing the local redirect may cause DNS traffic from all workloads on the affected nodes to fall back to the central kube-dns Service. The resulting sudden increase in traffic could overload kube-dns and cause a cluster-wide DNS outage.

cilium/CFP-44138-cilium-clusterwide-local-redirect-policy.md
```diff
@@ -0,0 +1,504 @@
...
+
+The initial CRD will be implemented with the `cilium.io/v2alpha1` API version. During the alpha phase, CLRP will remain supported but feature-frozen, except for critical fixes and changes required to enable migration to CCLRP.
+
+The CCLRP CRD will only be promoted to the `cilium.io/v2` API version once it is stable and migration-ready, at which point the existing CLRP CRD will be marked as deprecated.
+
+It is expected that both promotion and deprecation will be implemented in the same Cilium minor release. CCLRP and CLRP can continue to co-exist for several additional minor releases, providing sufficient time for administrators to migrate to the new resource.
+
+It is expected that CLRP will be removed in a future Cilium minor release; this scheduling is outside the scope of this document.
+
+### Migration approach
```
ysksuzuki, line 365: Could you describe both the concrete migration procedure and the implementation approach supporting it? In particular, I would like to understand how the existing StateDB `Service`, `Frontend`, and `Backend` model represents CLRP and CCLRP during coexistence, and how ownership of a frontend is transferred without temporarily withdrawing the redirect.

cilium/CFP-44138-cilium-clusterwide-local-redirect-policy.md
```diff
@@ -0,0 +1,504 @@
...
+
+1. The Kubernetes-facing layer validates and reflects CCLRP intent.  
+2. A reconciliation layer resolves Services, ports, and node-local backend endpoints.  
+3. The existing load-balancer integration applies concrete local redirect entries.
+
+The internal tables, Go structures, and controller decomposition may evolve during implementation. They should not become additional user-facing API commitments in this CFP.
+
+A prototype implementation will be used to validate the API and reconciliation behaviour with tests covering address matching, Service matching, namespace-restricted selectors, explicit port translation, invalid configuration, backend churn, nodes without local backends preserving the original frontend behaviour, and kernel capability handling.
+
+The initial CRD will be implemented with the `cilium.io/v2alpha1` API version. During the alpha phase, CLRP will remain supported but feature-frozen, except for critical fixes and changes required to enable migration to CCLRP.
```
ysksuzuki, line 357: Could you propose a concrete migration schedule, including the target releases for introducing CCLRP as `v2alpha1`, promoting it to `v2` and deprecating CLRP, and eventually removing CLRP?

ysksuzuki (changes requested):
Thank you for the CFP! I like the new CCLRP API. I've left a few comments.

cilium/CFP-44138-cilium-clusterwide-local-redirect-policy.md
ajmmm: I will improve the detail in the history then; most CFPs don't seem to go this far, though, so hence I kept it quite light. I had assumed the detail in the linked issue/PRs was sufficient.

I also agree the skip maps sizing should probably be assessed, but I wonder if that is just an implementation task rather than a design decision? The design of the mechanism exists today, and I don't believe it to be in question, beyond removing the API-level exposure to the administrator.

Now you bring it up, two things though:

1. I wonder if the sizing is documented properly. It's a constraint that should probably be visible.

2. I wonder if that should be runtime sizeable. What do you think?

cilium/CFP-44138-cilium-clusterwide-local-redirect-policy.md
ajmmm: Yes, sorry, I missed explicitly calling it out as a goal. Good spot!

cilium/CFP-44138-cilium-clusterwide-local-redirect-policy.md
ajmmm: I can expand on that, yes, but to do so would obviously re-introduce the data model into the CFP again. I actually took that out in an effort to get to a good first CFP within our time constraints.

As-is, there's a `1:N` relationship between the CLRP and the number of possible frontends it has to reconcile against. At first glance this makes it somewhat difficult to store the reconciliation status between CLRP and said frontends. It's my view the status field in the frontends table should only reflect status in respect to the BPF reconciler.

In my prototype I did explore introducing a new mapping table, which sat between the reflected LRP state from Kubernetes and the load balancer control plane state. This introduces a nice place to store LRP/frontend-specific status, but caused some duplication, so I parked it.

The reason I mention this is that it also facilitated alignment between the existing CLRP and CCLRP mappings. It provided a simple mechanism by which to enlighten the existing CLRP controller for overlap with CCLRP, allowing the CLRP controller to "stand down" and yield to CCLRP.

I will have a think on another approach here that can be documented in the CFP.

cilium/CFP-44138-cilium-clusterwide-local-redirect-policy.md
ajmmm: Is a concrete migration schedule really something that is to be set out in a CFP? I don't see other CFPs doing this.

I can introduce an example phasing across releases. For example:
* if CCLRP is ready for testing by release N, and
* CCLRP transitions to stable along-side CLRP being marked deprecated by release N+1, then
* CLRP will likely be removed by release N+3.

cilium/CFP-44138-cilium-clusterwide-local-redirect-policy.md
ysksuzuki: I think reassessing the skip map sizing belongs in this CFP because the proposal expands the use of the mechanism from CLRPs that explicitly opt in to every active CCLRP. The current limit of 100 might be introduced when loop prevention was opt-in and only a small number of entries was expected.

I think we should first estimate how much additional skip-map capacity will be required when all CCLRPs use the equivalent of `skipRedirectFromBackend=true` on capable kernels, and compare that with the current limit of 100 entries.  Based on that estimate, we can evaluate whether to increase the default size or make it configurable. Since changing the size of an existing pinned BPF map may require recreating or migrating it, the upgrade path should also be considered to avoid a period in which redirects are active but the required skip map entries are missing.

